PR #10, opened 2019-08-17
Add sparse support

gilad-rubin:
There's only one test that fails on my environment: "test_fit_psx", regardless of the addition of sparse support. I believe it has to do with rounding floating point numbers.


gilad-rubin:
I will try to add the relevant tests you suggested as soon as possible.

Regarding the conversion of dense matrices into sparse - I will try to take a look later on.
Thanks.

rsmith49:
@cgnorthcutt Just wanted to let you know - looks like these changes got overwritten by the [remove ipynb [ ] artifacts](https://github.com/cgnorthcutt/cleanlab/commit/cabcdc162d783b9d9770ed12b77440d58b745aec) commit. Adding back in the `check_X_y` method from `sklearn` allowed sparse matrices again

cgnorthcutt:
Thanks @rsmith49  for letting me know. This should be fixed. Let me know if it is not.

PR #50, opened 2020-10-10
Update import path

anishathalye:
fastText has updated their import path to `fasttext`. This patch makes it so that cleanlab works with the latest release of fastText on PyPI. See <https://fasttext.cc/blog/2019/06/25/blog-post.html>.


cgnorthcutt:
Thank you! To integrate fasttext back into cleanlab, I need to add tests back and add it back to requirements. Notes to future self.

PR #59, opened 2021-03-22
Doc

popokatapepel:
since I had quite a struggle with mapping your paper to the source code im seeing I created a html-doc for the package from your docstring using sphinx-autoapi. maybe thats useful for other users

simply do:

    $ cd docs
    $ make html

the html referencs is in `docs/build` afterwards


cgnorthcutt:
Hi @popokatapepel ,  Great suggestion. I've merged. But let's take it a step further and host the docs on readthedocs.org.

Would you like to take a stab at modifying the current set-up to integrate automatically with readthedocs?

popokatapepel:
sure. although i never uploaded any docs i think i will figure it out

popokatapepel:
ok i just tried it and that is pretty straight forward. However, the documentation already exists at https://cleanlab.readthedocs.io/en/latest/ so you probably already solved it your self.

cgnorthcutt:
Hi @popokatapepel , indeed, I added a hook to readthedocs, but I'm wondering if you would be willing to help witt the formatting? For example, https://pip.readthedocs.io/ looks nice. Even some simple changes such as adding spacing between the function declarations/docs and some light color (to lightly match the logo).

PR #98, opened 2021-12-27
Sync license text

anishathalye:
See old PR #96.

This patch updates metadata in `setup.py` and comments in other files to clarify that the current license is AGPLv3, as is specified in the current `LICENSE` file.

This patch also removes the "this agreement applies to this version and all previous versions" text from the README and code comments, because it is redundant; the previous code was already released under a _more liberal_ license, so giving the option of AGPLv3 is not useful (and perhaps confusing).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -571,10 +571,8 @@ License
 
 Copyright (c) 2017-2050 Curtis G. Northcutt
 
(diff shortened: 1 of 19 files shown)


cgnorthcutt (commented):
LGTM. Thanks @anishathalye

PR #99, opened 2021-12-27
Switch copyright assignment to Cleanlab

anishathalye:


cgnorthcutt (changes requested):
Please change to 2017-2022 given that the new year has arrived.

Otherwise, LGTM. Thanks @anishathalye

PR #105, opened 2022-01-30
Add basic contributing guide

anishathalye:
Mainly added common-sense notes on submitting issues/PRs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,55 @@
-Contributions to cleanlab are welcome. When contributing, please keep the following in mind:
+# Contributing to cleanlab
 
-1. Avoid adding blank lines or spaces to the ends of lines.
-2. Follow PEP-8 coding style.
-3. Always add tests to check the code you added. Tests go in `cleanlab/tests/`.
-4. Submit a pull request (it will automatically run your code and make sure nothing breaks).
+All kinds of contributions to cleanlab are greatly appreciated. If you're not
+looking to write code, submitting a [feature request](#feature-requests) or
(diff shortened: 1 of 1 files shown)


anishathalye:
Related to this, it might be good to add a development guide (e.g. describing how to run the tests).

CONTRIBUTING.md
```diff
@@ -1,8 +1,55 @@
...
+
+Want to write code to help improve cleanlab? Awesome!
+
+If there are [open issues][issues], you're more than welcome to work on those.
+If you have your own ideas, that's great too! In that case, before working on
+substantial changes to the code base, it is *highly recommended* that you first
+[open an issue][issue] describing what you intend to work on.
+
+Any changes to the code base should try to follow the style and coding
+conventions used in the rest of the project.
```
jwmueller, line 45: Should we ask that:  All code should adhere to the [PEP8 style](https://www.python.org/dev/peps/pep-0008/)?

jwmueller (approved):
lgtm!

CONTRIBUTING.md
anishathalye: I think standardizing code formatting is a great idea, but just asking people to follow a style is not enough. We should set up:

- An auto-formatter that's as easy to run as `go fmt` / `cargo fmt`, which will guarantee compliance with the style rules
- Checking in CI that code is formatted correctly

Python doesn't have a language-mandated standard for this, and I don't know off the top of my head which is the most popular community project for doing this (maybe [Black](https://black.readthedocs.io)?). I'll look into this and set it up as a separate PR.

PR #108, opened 2022-02-01
Follow requirement to allow finalizers to run

anishathalye:
Tests on some platforms and Python versions (Python 3.7, 3.8, and 3.9 on Ubuntu and macOS) were hanging with coverage==6.3.

See https://github.com/nedbat/coveragepy/issues/1310 and https://github.com/nedbat/coveragepy/issues/1312 for some discussion about this issue. Some people have pinned coverage==6.2 as a temporary workaround. As discussed in #1312, coverage fixes this issue in 00da68ef1a0b4d43c003babae0cb8f91beaf06d2 by only setting signal handlers from the main thread; I confirmed that using the latest commit of coverage (aad5ece47bf12bceff4296516f23171a06b34bb5) fixes the issue of tests hanging without any modification to our code.

However, according to the pytest-cov documentation (https://pytest-cov.readthedocs.io/en/latest/subprocess-support.html#if-you-use-multiprocessing-pool), when using `multiprocessing.Pool`, we are supposed to ensure that `join()` is called on the pool, rather than `terminate()`. Following this requirement makes the tests work with coverage==6.3 as well.

### DIFF
--- a/cleanlab/pruning.py
+++ b/cleanlab/pruning.py
@@ -29,6 +29,7 @@ from __future__ import (
 from sklearn.preprocessing import MultiLabelBinarizer
 import multiprocessing
 from multiprocessing.sharedctypes import RawArray
+from contextlib import contextmanager
 import sys
 import os
 import time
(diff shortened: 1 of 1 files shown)


jwmueller:
lgtm, but will defer final review to @cgnorthcutt 

It's a bit weird that we need to change our source code to accommodate the coverage tests though, hopefully that doesn't happen again going forward...

anishathalye:
If we don't care about coverage on Windows, we don't need this. `coverage==6.3` was trying to fix this limitation of not requiring a `join()` and saving coverage data on `terminate()` as well (in https://github.com/nedbat/coveragepy/commit/dd575eec0fa173bcb6c680d2f1c822a28280d011), but it didn't get it quite right, hence CI breaking. https://github.com/nedbat/coveragepy/commit/00da68ef1a0b4d43c003babae0cb8f91beaf06d2 fixes the issue, so using this version of coverage (not yet released on PyPI) won't require code changes on our end, but I think it won't save coverage in forked threads on Windows? (But it wasn't doing this before anyways?)

Maybe that's a fine trade-off, and we can ignore this PR (and pin an old version of coverage until a new version is released). #109

PR #109, opened 2022-02-02
Ban coverage==6.3

anishathalye:
See https://github.com/nedbat/coveragepy/issues/1312. We can remove this once a newer version of coverage is released that includes https://github.com/nedbat/coveragepy/commit/00da68ef1a0b4d43c003babae0cb8f91beaf06d2.

Competing PR: #108


jwmueller (approved):
I like this simpler solution :)

PR #115, opened 2022-02-16
Dedupe deps

anishathalye:
This was duplicating content in setup.py (and the two had gotten out of sync, listing different versions for scipy).

According to the Python Packaging User Guide (https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/), the `install_requires` specifies what a project minimally needs to run correctly, while the requirements.txt file contains an exhaustive list of pinned versions for the purpose of repeatable installations of a complete environment. Cleanlab is a Python package that users will `pip install`, so we don't need a requirements.txt.

Some related projects like scikit-learn don't have a requirements.txt, and others that do have such a file use it for a different purpose, e.g. PyTorch and Keras use the file to list dev dependencies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,10 +25,9 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python }}
-      - name: Install dependencies
+      - name: Install test dependencies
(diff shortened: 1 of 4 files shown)


jwmueller (approved):
LGTM

PR #128, opened 2022-03-10
Major API change. Introducing Cleanlab 2.0

cgnorthcutt:
# CHANGELOG Cleanlab 1.0.1 --> 2.0:

At a high level, this update redesigns `cleanlab` to be scalable and extensible for growth (e.g., adding new ways to rank data and labels, adding new methods for computing data and labels quality, adding new tasks like regression and object detection, etc.) This update also simplifies most of the naming conventions, redesigning `cleanlab` to be more developer friendly and less academic.

## Module name changes:
- `pruning.py` --> `filter.py`
- `latent_estimation.py` --> `count.py`
- parent module/folder `models/` --> `example_models/`

## New module created:
- `rank.py`
  - moved all ranking and ordering functions from `pruning.py`/`filter.py` to here

## Method name changes:
- `pruning.get_noise_indices()` --> `filter.find_label_issues()`
- `count.num_label_errors()` --> `count.num_label_issues()`

## Methods added:
- `rank.py` adds
  - two ranking functions to rank data based on label quality for entire dataset (not just examples with label issues)
  - `get_self_confidence_for_each_label()`
  - `get_normalized_margin_for_each_label()`
- `filter.py` adds
  - two more methods added to `filter.find_label_issues()` (select method using the `filter_by` parameter)
    - `confident_learning`, which has been shown to work very well and may become the default in the future, and
    - `predicted_neq_given`, which is useful for benchmarking a simple baseline approach, but underperformant relative to the other filter_by methods)
- `classification.py` adds
  - `LearningWithNoisyLabels.get_label_issues()`
    - for a canonical one-line of code use:`LearningWithNoisyLabels().fit(X, y).get_label_issues()` 
    - no need to compute predicted probabilities in advance

## Naming conventions changed in method names, comments, parameters, etc.
- `s` -> `labels` 
- `psx` -> `pred_probs`
- `label_errors` --> `label_issues`
- `noise_mask` --> `label_issues_mask`
- `label_errors_bool` --> `label_issues_mask`
- `prune_method` --> `filter_by`
- `prob_given_label` --> `self_confidence`
- `pruning` --> `filtering`

## Parameter re-ordering:
- re-ordered (`labels`, `pred_probs`) parameters to be consistent (in that order) in all methods.
- re-ordered parameters (e.g. `frac_noise`) in filter.find_label_issues()

## Parameter changes:
- in `order_label_issues()`
  - param: `sorted_index_method` --> `rank_by`
- in `find_label_issues()`
  - param: `sorted_index_method` --> `return_indices_ranked_by`
  - param: `prune_method` --> `filter_by`

## Global variables changed:
- `filter.py`
  - Only require 1 example to be left in each class
  - `MIN_NUM_PER_CLASS = 5` --> `MIN_NUM_PER_CLASS = 1`
  - enables cleanlab to work for toy-sized datasets


cleanlab/rank.py
```diff
@@ -0,0 +1,159 @@
...
+        [np.mean(psx[i][labels[i]]) for i in label_issues_idx]
+    )
+    if rank_by == "prob_given_label":
+        return label_issues_idx[np.argsort(self_confidence)]
+    elif rank_by == "normalized_margin":
+        psx_er, labels_er = psx[label_issues_mask], labels[label_issues_mask]
+        max_prob_not_label = np.array(
+            [max(np.delete(psx_er[i], l, -1)) for i, l in enumerate(labels_er)]
+        )
+        margin = self_confidence - max_prob_not_label
```
JohnsonKuan, line 76: Can reuse rank scoring function below:

`margin = get_normalized_margin_for_each_label(labels_er, psx_er)`

cleanlab/latent_algebra.py
```diff
@@ -109,18 +109,18 @@ def compute_inv_noise_matrix(py, noise_matrix, ps=None):
...
+        # 'ps' is p(labels=k) = noise_matrix * p(y=k)
+        # because in *vector computation*: P(labels=k|y=k) * p(y=k) = P(labels=k)
         if ps is None:
             ps = noise_matrix.dot(py)
 
-        # Estimate the (K, K) inverse noise matrix P(y = k_y | s = k_s)
+        # Estimate the (K, K) inverse noise matrix P(y = k_y | labels = k_s)
         inverse_noise_matrix = np.empty(shape=(K,K))
-        # k_s is the class value k of noisy label s
+        # k_s is the class value k of noisy label labels
```
calebchiam, line 119: Confusing

cleanlab/latent_algebra.py
```diff
@@ -264,13 +264,13 @@ def compute_py(ps, noise_matrix, inverse_noise_matrix, py_method='cnt',
         err += ' should be in [cnt, eqn, marginal, marginal_ps]'
         raise ValueError(err)
 
-    # Clip py (0,1), .s.t. no class should have prob 0, hence 1e-5
+    # Clip py (0,1), .labels.t. no class should have prob 0, hence 1e-5
```
calebchiam, line 267: Typo

cleanlab/noise_generation.py
```diff
@@ -76,8 +76,8 @@ def noise_matrix_is_valid(noise_matrix, py, verbose=False):
 
 
 def generate_noisy_labels(y, noise_matrix, verbose=False):
-    """Generates noisy labels s (shape (N, 1)) from perfect labels y,
-    'exactly' yielding the provided noise_matrix between s and y.
+    """Generates noisy labels labels (shape (N, 1)) from perfect labels y,
```
calebchiam, line 79: Can we style these differently? e.g. Generates noisy labels `labels` (shape (N, 1))

cleanlab/noise_generation.py
```diff
@@ -124,29 +124,29 @@ def generate_noisy_labels(y, noise_matrix, verbose=False):
...
-    # Generate s
+    # Generate labels
     s = np.array(y)
     for k in range(K):  # Iterate over true class y == k
-        # Get the noisy s labels that have non-zero counts
+        # Get the noisy labels labels that have non-zero counts
         s_labels = np.where(count_joint[:, k] != 0)[0]
-        # Find out how many of each noisy s label we need to flip to
-        s_counts = count_joint[s_labels, k]
+        # Find out how many of each noisy labels label we need to flip to
```
calebchiam, line 137: "labels label"

cleanlab/util.py
```diff
@@ -294,19 +294,19 @@ def onehot2int(onehot_matrix):
 
 
 def estimate_pu_f1(s, prob_s_eq_1):
-    """Computes Claesen's estimate of f1 in the pulearning setting.
+    """Computes Claesen'labels estimate of f1 in the pulearning setting.
```
calebchiam, line 297: labels -> s

cleanlab/util.py
```diff
@@ -294,19 +294,19 @@ def onehot2int(onehot_matrix):
...
       Binary label (whether each element is labeled or not) in pu learning.
 
     prob_s_eq_1 : iterable (list or np.array)
-      The probability, for each example, whether it is s==1 P(s==1|x)
+      The probability, for each example, whether it is labels==1 P(labels==1|x)
 
     Output (float)
     ------
-    Claesen's estimate for f1 in the pulearning setting."""
+    Claesen'labels estimate for f1 in the pulearning setting."""
```
calebchiam, line 309: labels -> s

cleanlab/noise_generation.py
```diff
@@ -35,21 +35,21 @@ def noise_matrix_is_valid(noise_matrix, py, verbose=False):
     # Number of classes
     K = len(py)
 
-    # Let's assume some number of training examples for code readability, 
+    # Let'labels assume some number of training examples for code readability,
```
calebchiam, line 38: labels -> s

cleanlab/latent_algebra.py
```diff
@@ -33,21 +33,21 @@
...
-    """Compute ps := P(s=k), py := P(y=k), and the inverse noise matrix.
+    """Compute ps := P(labels=k), py := P(y=k), and the inverse noise matrix.
 
     Parameters
     ----------
 
     s : np.array
-        A discrete vector of labels, s, which may contain mislabeling. "s"
+        A discrete vector of labels, labels, which may contain mislabeling. "labels"
         denotes the noisy label instead of \\tilde(y), for ASCII reasons.
```
calebchiam, line 43: Style the second labels differently. Also, the second statement is now obsolete since we're no longer using 's'.

cleanlab/filter.py
```diff
@@ -144,24 +143,24 @@ def _prune_by_class(k, args=None):
...
     else:  # Multiprocessing - data is shared across sub-processes
-        s, s_counts, prune_count_matrix, psx, multi_label = _get_shared_data()
+        labels, label_counts, prune_count_matrix, psx, multi_label = _get_shared_data()
 
-    if s_counts[k] > MIN_NUM_PER_CLASS:  # No prune if not MIN_NUM_PER_CLASS
-        num_errors = s_counts[k] - prune_count_matrix[k][k]
-        # Get rank of smallest prob of class k for examples with noisy label k
-        s_filter = np.array(
-            [k in lst for lst in s]) if multi_label else s == k
+    if label_counts[k] > MIN_NUM_PER_CLASS:  # No prune if not MIN_NUM_PER_CLASS
```
calebchiam, line 150: If we're using a > instead of >=, then the actual minimum number per class is 2, right?

Perhaps change MIN_NUM_PER_CLASS to 2, and use >= here for clarity.

calebchiam: Relevant for L181-183 as well.

cleanlab/filter.py
```diff
@@ -290,38 +281,43 @@ def get_noise_indices(
 
     num_to_remove_per_class : list of int of length K (# of classes)
       e.g. if K = 3, num_to_remove_per_class = [5, 0, 1] would return
-      the indices of the 5 most likely mislabeled examples in class s = 0,
-      and the most likely mislabeled example in class s = 1.
+      the indices of the 5 most likely mislabeled examples in class labels = 0,
+      and the most likely mislabeled example in class labels = 1.
```
calebchiam, line 285: labels -> label

README.md
```diff
@@ -65,24 +65,24 @@ Past release notes and **future features planned** is available [here](https://g
 `cleanlab` is:
 
 1.  **backed-by-theory** - Provable perfect label error finding in some realistic conditions.
-2.  **fast** - Non-iterative, parallelized algorithms (e.g. < 1 second to find label errors in ImageNet with pre-computed probabilities)
+2.  **fast** - Non-iterative, parallelized algorithms (e.g. < 1 second to find label issues in ImageNet with pre-computed probabilities)
 3.  **general** - Works with any dataset, model, and framework, e.g., Tensorflow, PyTorch, sklearn, xgboost, etc.
 
-## Find label errors with PyTorch, Tensorflow, sklearn, xgboost, etc. in 1 line of code
+## Find confident_learning with PyTorch, Tensorflow, sklearn, xgboost, etc. in 1 line of code
```
calebchiam, line 71: confident_learning -> label issues

README.md
```diff
@@ -65,24 +65,24 @@ Past release notes and **future features planned** is available [here](https://g
...
 
 ```python
 # Compute psx (n x m matrix of predicted probabilities) on your own, with any classifier.
 # Here is an example that shows in detail how to compute psx on CIFAR-10:
 #    https://github.com/cleanlab/examples/tree/master/cifar10
 # Be sure you compute probs in a holdout/out-of-sample manner (e.g. via cross-validation)
-# Now getting label errors is trivial with cleanlab... its one line of code.
-# Label errors are ordered by likelihood of being an error. First index is most likely error.
-from cleanlab.pruning import get_noise_indices
+# Now getting confident_learning is trivial with cleanlab... its one line of code.
```
calebchiam, line 78: confident_learning -> label issues

cleanlab/classification.py
```diff
@@ -244,28 +250,28 @@ def fit(
...
+        labels : :obj:`np.array`
+          A binary vector of labels, labels, which may contain mislabeling.
 
         psx : :obj:`np.array` (shape (N, K))
-          P(s=k|x) is a matrix with K (noisy) probabilities for each of the N
+          P(labels=k|x) is a matrix with K (noisy) probabilities for each of the N
           examples x.
           This is the probability distribution over all K classes, for each
-          example, regarding whether the example has label s==k P(s=k|x). psx
+          example, regarding whether the example has label labels==k P(labels=k|x). psx
```
calebchiam, line 268: labels -> label, but also needs to be rewritten slightly to avoid "label label"

cleanlab/classification.py
```diff
@@ -244,28 +250,28 @@ def fit(
...
           should have been computed using 3 (or higher) fold cross-validation.
           If you are not sure, leave psx = None (default) and
           it will be computed for you using cross-validation.
 
         thresholds : :obj:`iterable` (list or np.array) of shape (K, 1)  or (K,)
-          P(s^=k|s=k). List of probabilities used to determine the cutoff
+          P(labels^=k|labels=k). List of probabilities used to determine the cutoff
```
calebchiam, line 274: labels -> label

cleanlab/classification.py
```diff
@@ -431,7 +431,7 @@ def predict_proba(self, *args, **kwargs):
         return self.clf.predict_proba(*args, **kwargs)
 
     def score(self, X, y, sample_weight=None):
-        """Returns the clf's score on a test set X with labels y.
+        """Returns the clf'labels score on a test set X with labels y.
```
calebchiam, line 434: labels -> s

calebchiam:
There are a number of "s -> labels" replacement errors throughout the repo:

- Some of these are apostrophe errors, e.g. "Cleanlab's" -> "Cleanlab'labels"
- Some are readability issues, e.g. "the labels s" -> "the labels labels"
- Some are semantic issues, where originally `s` was used to denote a single label class `s`, e.g. `P(s=k|y=k)` -> `P(labels=k|y=k)`

I've tagged these where I've seen them but ran into enough instances that I'm holding off on that to avoid spamming the page with review comments.

README.md
jwmueller: this substitution error seems to recur, so worth checking every occurrence of "confident_learning" in the new code.

jwmueller: and every occurrence of "labels"

cleanlab/filter.py
```diff
@@ -290,38 +281,43 @@ def get_noise_indices(
...
       if ``num_to_remove_per_class == k``, then either k-1, k, or k+1
       examples may be removed for any class. This is because noise rates
       are floats, and rounding may cause a one-off. If you need exactly
       'k' examples removed from every class, you should use ``'prune_by_class'``
 
-    prune_method : str (default: 'prune_by_noise_rate')
-      Possible Values: 'prune_by_class', 'prune_by_noise_rate', or 'both'.
-      Method used for pruning.
+    filter_by : str (default: 'prune_by_noise_rate')  TODO: change default to cl_off_diag?
+      Possible Values: {'prune_by_class', 'prune_by_noise_rate', 'both',
```
jwmueller, line 297: why include `prune_by_*` prefix when the arg name already says `filter_by`?

cleanlab/filter.py
```diff
@@ -290,38 +281,43 @@ def get_noise_indices(
...
       *high probability* of being mislabeled for every non-diagonal
-      in the prune_counts_matrix (see pruning.py).
+      in the prune_counts_matrix (see filter.py).
       2. 'prune_by_class': works by removing the examples with *smallest
       probability* of belonging to their given class label for every class.
       3. 'both': Finds the examples satisfying (1) AND (2) and
       removes their set conjunction.
+      4. 'confident_learning': Find examples that are confidently labeled as a class
+      that'labels different from their given label while computing the confident joint.
+      5. 'argmax_not_equal_given_label': Find examples where the argmax predictions does not match
```
jwmueller, line 309: 'misclassified' instead of 'argmax_not_equal_given_label'?

jwmueller: "the argmax predictions" 
->
"the predicted class (i.e. argmax of the predicted probabilities)"

cleanlab/filter.py
```diff
@@ -290,38 +281,43 @@ def get_noise_indices(
...
       1. 'prune_by_noise_rate': works by removing examples with
       *high probability* of being mislabeled for every non-diagonal
-      in the prune_counts_matrix (see pruning.py).
+      in the prune_counts_matrix (see filter.py).
       2. 'prune_by_class': works by removing the examples with *smallest
       probability* of belonging to their given class label for every class.
       3. 'both': Finds the examples satisfying (1) AND (2) and
       removes their set conjunction.
+      4. 'confident_learning': Find examples that are confidently labeled as a class
+      that'labels different from their given label while computing the confident joint.
```
jwmueller, line 308: language is unclear

cleanlab/filter.py
```diff
@@ -290,38 +281,43 @@ def get_noise_indices(
...
+                        'confident_learning', 'argmax_not_equal_given_label'}
+      Method used for filtering/pruning out the label issues.
       1. 'prune_by_noise_rate': works by removing examples with
       *high probability* of being mislabeled for every non-diagonal
-      in the prune_counts_matrix (see pruning.py).
+      in the prune_counts_matrix (see filter.py).
       2. 'prune_by_class': works by removing the examples with *smallest
       probability* of belonging to their given class label for every class.
       3. 'both': Finds the examples satisfying (1) AND (2) and
       removes their set conjunction.
```
jwmueller, line 306: clearer as: "Removes only the examples that

jwmueller, line 306: clearer as:
"Removes only the examples that would be filtered by both (1) AND (2)"

cleanlab/filter.py
```diff
@@ -290,38 +281,43 @@ def get_noise_indices(
...
 
-    prune_method : str (default: 'prune_by_noise_rate')
-      Possible Values: 'prune_by_class', 'prune_by_noise_rate', or 'both'.
-      Method used for pruning.
+    filter_by : str (default: 'prune_by_noise_rate')  TODO: change default to cl_off_diag?
+      Possible Values: {'prune_by_class', 'prune_by_noise_rate', 'both',
+                        'confident_learning', 'argmax_not_equal_given_label'}
+      Method used for filtering/pruning out the label issues.
       1. 'prune_by_noise_rate': works by removing examples with
       *high probability* of being mislabeled for every non-diagonal
```
jwmueller, line 301: could clarify what "for every non-diagonal" means

cleanlab/filter.py
```diff
@@ -290,38 +281,43 @@ def get_noise_indices(
...
       are floats, and rounding may cause a one-off. If you need exactly
       'k' examples removed from every class, you should use ``'prune_by_class'``
 
-    prune_method : str (default: 'prune_by_noise_rate')
-      Possible Values: 'prune_by_class', 'prune_by_noise_rate', or 'both'.
-      Method used for pruning.
+    filter_by : str (default: 'prune_by_noise_rate')  TODO: change default to cl_off_diag?
+      Possible Values: {'prune_by_class', 'prune_by_noise_rate', 'both',
+                        'confident_learning', 'argmax_not_equal_given_label'}
+      Method used for filtering/pruning out the label issues.
```
jwmueller, line 299: descriptions of each option should be standardized as to what is being "removed" or "found". Currently some descriptions say "find" (eg. "confident_learning"), others say "remove" (eg. options 1 and 2).

cleanlab/filter.py
```diff
@@ -554,50 +582,90 @@ def reduce_prune_counts(prune_count_matrix, frac_noise=1.0):
...
+    `count.compute_confident_joint`.
+
+    This method does not support multi-label labels.
+
+    Parameters
+    ----------
 
     labels : np.array
-      A binary vector of labels, which may contain label errors.
+        A discrete vector of noisy labels, i.e. some labels may be erroneous.
```
jwmueller, line 636: clarify: for dataset with `K` classes, labels must be in `{0,1,...,K-1}`?

cleanlab/filter.py
```diff
@@ -554,50 +582,90 @@ def reduce_prune_counts(prune_count_matrix, frac_noise=1.0):
...
 
-    sorted_index_method : str ['normalized_margin', 'prob_given_label']
-      Method to order label error indices (instead of a bool mask), either:
-        'normalized_margin' := normalized margin (p(s = k) - max(p(s != k)))
-        'prob_given_label' := [psx[i][labels[i]] for i in label_errors_idx]
+    psx : np.array (shape (N, K))
+        P(label=k|x) is a matrix with K (noisy) probabilities for each of the
+        N examples x. This is the probability distribution over all K classes,
```
jwmueller, line 640: using the term 'noisy' here seems confusing to me, I'd reserve it for wrong labels, and use something like 'estimated' here.

cleanlab/filter.py
```diff
@@ -554,50 +582,90 @@ def reduce_prune_counts(prune_count_matrix, frac_noise=1.0):
...
+        for each example, regarding whether the example has label labels==k P(labels=k|x).
```
jwmueller, line 641: I'd replace this text with:
"Each row of this matrix corresponds to an example `x` and contains the model-estimated probabilities that `x` belongs to each possible class. The columns must be ordered such that these probabilities correspond to class 0,1,2,... Properly formatted `psx` may be obtained from the `predict_proba()` function in sklearn."

cleanlab/filter.py
```diff
@@ -236,50 +235,42 @@ def multiclass_crossval_predict(pyx, labels):
...
+      A binary vector of labels, labels, which may contain mislabeling. "labels" denotes
       the noisy label instead of \\tilde(y), for ASCII encoding reasons.
 
     psx : np.array (shape (N, K))
-      P(s=k|x) is a matrix with K (noisy) probabilities for each of the N
+      P(labels=k|x) is a matrix with K (noisy) probabilities for each of the N
       examples x.
       This is the probability distribution over all K classes, for each
-      example, regarding whether the example has label s==k P(s=k|x).
+      example, regarding whether the example has label labels==k P(labels=k|x).
```
jwmueller, line 266: See my comments about this arg below for `find_argmax_not_equal_given_label` function.

cleanlab/filter.py
```diff
@@ -554,50 +582,90 @@ def reduce_prune_counts(prune_count_matrix, frac_noise=1.0):
...
 
     labels : np.array
-      A binary vector of labels, which may contain label errors.
+        A discrete vector of noisy labels, i.e. some labels may be erroneous.
 
-    sorted_index_method : str ['normalized_margin', 'prob_given_label']
-      Method to order label error indices (instead of a bool mask), either:
-        'normalized_margin' := normalized margin (p(s = k) - max(p(s != k)))
-        'prob_given_label' := [psx[i][labels[i]] for i in label_errors_idx]
+    psx : np.array (shape (N, K))
```
jwmueller, line 638: Are we committed to the `psx` arg name? 
Or do you prefer something more intuitive like: `preds`, `y_pred`, `pred_probas`?

In particular if trying to extend to other ML tasks like regression, it may be nice to have a thematic set of names for the model predictions across different tasks.  We could use `psx` for all model predictions regardless of task, it's not that intuitive of a name though (seems more mathematical).

cleanlab/filter.py
```diff
@@ -236,50 +235,42 @@ def multiclass_crossval_predict(pyx, labels):
...
+    You can return ONLY the indices of the label issues in your dataset, by setting
+    return_indices_ranked_by = {`prob_given_label`, `normalized_margin`}.
+
     number of indices returned is specified by frac_of_noise. When
     frac_of_noise = 1.0, all "confident" estimated noise indices are returned.
     * If you encounter the error 'psx is not defined', try setting n_jobs = 1.
 
+    WARNING! frac_noise and num_to_remove_per_class parameters are only supported when filter_by
+    is either 'prune_by_noise_rate', 'prune_by_class', or 'both'. They are not supported for methods
+    'confident_learning' or 'argmax_not_equal_given_label'. TODO.
```
jwmueller, line 253: `frac_noise` and `num_to_remove_per_class` seem like secondary args that should appear after more important args like: `filter_by`, `return_indices_ranked_by`.

cleanlab/filter.py
```diff
@@ -236,50 +235,42 @@ def multiclass_crossval_predict(pyx, labels):
...
+def find_label_issues(labels, psx, confident_joint=None, frac_noise=1.0,
+                      num_to_remove_per_class=None, filter_by='prune_by_noise_rate',
+                      return_indices_ranked_by=None, multi_label=False, n_jobs=None, verbose=0):
+    """By default, this method returns a boolean mask for the entire dataset where True represents
+    a label issue and False represents an example that is confidently/accurately labeled.
+
+    You can return ONLY the indices of the label issues in your dataset, by setting
+    return_indices_ranked_by = {`prob_given_label`, `normalized_margin`}.
+
     number of indices returned is specified by frac_of_noise. When
```
jwmueller, line 247: `frac_of_noise` -> `frac_noise`

cleanlab/filter.py
```diff
@@ -236,50 +235,42 @@ def multiclass_crossval_predict(pyx, labels):
...
     frac_of_noise = 1.0, all "confident" estimated noise indices are returned.
```
jwmueller, line 248: is there rationale for not allowing `frac_of_noise` to be > 1 (i.e. to get higher recall in label-error detection)?

jwmueller: FYI the current code seems to still run with `frac_noise` > 1 even though it has no effect, should probably raise error.

cleanlab/filter.py
```diff
@@ -236,50 +235,42 @@ def multiclass_crossval_predict(pyx, labels):
...
-      Assumes columns of inverse_noise_matrix sum to 1.
-
     confident_joint : np.array (shape (K, K), type int) (default: None)
-      A K,K integer matrix of count(s=k, y=k). Estimates a a confident
-      subset of the joint distribution of the noisy and true labels P_{s,y}.
+      A K,K integer matrix of count(labels=k, y=k). Estimates a a confident
+      subset of the joint distribution of the noisy and true labels P_{labels,y}.
       Each entry in the matrix contains the number of examples confidently
-      counted into every pair (s=j, y=k) classes.
+      counted into every pair (labels=j, y=k) classes.
```
jwmueller, line 273: would be good to specify which function may have been previously called to produce this `confident_joint`

cleanlab/filter.py
```diff
@@ -290,38 +281,43 @@ def get_noise_indices(
...
       (instead of a bool mask) where error indices are ordered by the either:
-      ``'normalized_margin' := normalized margin (p(s = k) - max(p(s != k)))``
-      ``'prob_given_label' := [psx[i][labels[i]] for i in label_errors_idx]``
+      ``'normalized_margin' := normalized margin (p(label = k) - max(p(label != k)))``
+      ``'prob_given_label' := [psx[i][labels[i]] for i in label_issues_idx]``
 
     multi_label : bool
-      If true, s should be an iterable (e.g. list) of iterables, containing a
+      If true, labels should be an iterable (e.g. list) of iterables, containing a
       list of labels for each example, instead of just a single label.
```
jwmueller, line 321: clarify more what `multi_label` setting is. Also would be nice to provide toy example of what `labels` may look like when `multi_label == True`

cleanlab/filter.py
```diff
@@ -330,137 +326,169 @@ def get_noise_indices(
       Set this to 1 to REMOVE parallel processing (if its causing issues).
 
     verbose : int
-      If 0, no print statements. If 1, prints when multiprocessing happens."""
+      If 0, no print statements. If 1, prints when multiprocessing happens.
```
jwmueller, line 329: would be nice to print the number of label issues found if `verbose = 1`

cleanlab/filter.py
```diff
@@ -236,50 +235,42 @@ def multiclass_crossval_predict(pyx, labels):
...
-        frac_noise=1.0,
-        num_to_remove_per_class=None,
-        prune_method='prune_by_noise_rate',
-        sorted_index_method=None,
-        multi_label=False,
-        n_jobs=None,
-        verbose=0,
-):
-    """Returns the indices of most likely (confident) label errors in s. The
+def find_label_issues(labels, psx, confident_joint=None, frac_noise=1.0,
```
jwmueller, line 238: I ran some (small-scale) experiments with deterministic underlying X->Y relationship, large sample-size, accurate model, low noise rate, and uniform (non-class-conditional) label noise (ie. the easiest possible setting).  

I found default value of `frac_noise = 1` tends to give much better recall than  precision in detecting label errors. However reducing `frac_noise` to say 0.5 or  0.9 did not boost the precision much, indicating there's hopefully big gains  to made by improving the label-quality-scoring functions. If `frac_noise>1` values were allowed, I wonder if recall could be further boosted without harming precision...

@JohnsonKuan

cleanlab/models/fasttext.py
```diff
@@ -246,7 +246,7 @@ def predict_proba(self, X=None, train_data=True, return_labels=False):
         for labels, text in data_loader(fn=fn, indices=X,
                                         batch_size=self.batch_size):
             pred = self.clf.predict(text=text, k=len(self.clf.get_labels()))
-            # Get p(s = k | x) matrix of shape (N x K) of pred probs for each x
+            # Get p(labels = k | x) matrix of shape (N x K) of pred probs for each x
             psx = [[p for _, p in sorted(list(zip(*l)), key=lambda x: x[0])] for
                    l in list(zip(*pred))]
             psx_list.append(np.array(psx))
```
jwmueller, line 252: Do we want to keep fasttext.py and the other `models/` files in the main repo?  Or should they be moved to `cleanlab/examples` or potentially even a new sklearn-wrappers repo?

At the very least, I'd suggest renaming this `models/` folder to something like `example_models/` to make it more clear these models are arbitrary choices and not being utilized by the rest of cleanlab's core functionality.

jwmueller:
`from __future__ import ...` lines can be removed throughout

cleanlab/count.py
```diff
@@ -48,12 +47,12 @@
 )
 
 
-def num_label_errors(
+def num_label_issues(
```
jwmueller, line 50: Is this function really necessary? 
Can't a user just do `error_inds = find_label_issues(..., return_indices_ranked_by=...)` and then: `len(error_inds)`?

What is user supposed to conclude when `len(error_inds)` does not match output of this function?  This function supports much fewer options than `find_label_issues`, also unclear why that is.

jwmueller: [Quality of life improvement] I also advocate `find_label_issues()` should automatically print out the number of label issues during its execution.

cleanlab/count.py
```diff
@@ -102,17 +101,17 @@ def calibrate_confident_joint(confident_joint, s, multi_label=False):
     ----------
 
     confident_joint : np.array (shape (K, K))
-        A K,K integer matrix of count(s=k, y=k). Estimates a confident subset of
-        the joint disribution of the noisy and true labels P_{s,y}.
+        A K,K integer matrix of count(labels=k, y=k). Estimates a confident subset of
+        the joint disribution of the noisy and true labels P_{labels,y}.
```
jwmueller, line 105: Are we sticking with notation `labels` for observed (noisy) labels and `y` for true labels?

cleanlab/latent_algebra.py
```diff
@@ -18,7 +18,7 @@
 # ## Latent Algebra
```
jwmueller, line 18: Consider moving lower level helper functions like those in this `latent_algebra.py` and `util.py` files to a `utils/` subdirectory. 
In particular, we want to be careful about our documentation becoming cluttered by functions intended only for the most advanced users.  If all helper functions are in a `utils/` subdirectory, then the documentation website can hide their documentation inside a pull-down `utils` tab that presumably only advanced users will ever explore.

@weijinglok

cleanlab/rank.py
cgnorthcutt: Great feedback. Done.

README.md
cgnorthcutt: Ran a search and only found these two cases for `confident_learning`. Also, ran a similar search / fix for the `labels labels` issue.

cleanlab/classification.py
cgnorthcutt: done for all

cleanlab/classification.py
cgnorthcutt: done for all

cleanlab/classification.py
cgnorthcutt: fixed everywhere

cleanlab/filter.py
cgnorthcutt: convention. methods are called PBNR, PBC in the paper (PB = prune by). agree its redundant, but didn't feel it was necessary to further separate the paper naming of the method (where many of our developers come from) and potentially confuse folks as to what the method means over 8 chars.  (e.g. `prune_by_class` would become `class` which is a terrible name for a method)

cleanlab/filter.py
cgnorthcutt: Changed to `4. 'confident_learning': Returns the examples in the off-diagonals of the confident joint.
      These are the examples that are confidently predicted to be a different label than their
      given label.`

cleanlab/filter.py
cgnorthcutt: LGTM. changed.

cleanlab/filter.py
cgnorthcutt: second change made.

for first change, I've changed `argmax_not_equal_given_label` to `predicted_neq_given`. I want to avoid the term `misclassified` because misclassified implies `pred != true label`, where in this case, we mean instead `pred != given label`, which is slightly different.

cleanlab/filter.py
cgnorthcutt: done.

cleanlab/filter.py
cgnorthcutt: clarified everywhere

cleanlab/filter.py
cgnorthcutt: good point. fixed everywhere

cleanlab/filter.py
cgnorthcutt: made some adjustments, but added a few things. most important thing to include here is that these are computed via cross validation which isn't obvious with sklearn for any model, but took some thoughts from here and added everywhere

cgnorthcutt:
```python
psx : np.array (shape (N, K))
        P(label=k|x) is a matrix with K model-predicted probabilities.
        Each row of this matrix corresponds to an example x and contains the model-predicted
        probabilities that x belongs to each possible class.
        The columns must be ordered such that these probabilities correspond to class 0,1,2,...
        `psx` should have been computed using 3 (or higher) fold cross-validation.
```

cleanlab/filter.py
cgnorthcutt: I could switch to `pred_probas` everywhere

cgnorthcutt: pred == list of predictions (not a matrix of probabilities)

cgnorthcutt: I don't see any standard internal representation for this, even in `sklearn`. Try searching the entire repo on GitHub: https://github.com/scikit-learn/scikit-learn/search?q=pred_proba

I'm fine to leave as is for now for this reason, but we could also switch it to plx for prob(labels=x) or just `pred_probs`

cleanlab/filter.py
cgnorthcutt: Great point!

cleanlab/filter.py
cgnorthcutt: Nice. fixed.

cleanlab/filter.py
cgnorthcutt: It's an easter egg use case. Basically, I want the ability to do your first comment without confusing users so i say it needs to be between 0 and 1, but for advanced users, you should be able to increase it... but eventualy... things will break. Instead of handling all the edge cases, I just tell the users what the input should be and if they want to break it they can.

Perhaps I could add a warning saying a succinct version of the above (i.e., `Warning `frac_noise` is intended to be between 0 and 1. You can use large values, but things may break. Use at your own risk``)

Thoughts?

cleanlab/filter.py
cgnorthcutt: good idea. done

cleanlab/filter.py
cgnorthcutt: clarified everywhere.

cleanlab/filter.py
cgnorthcutt: added

```
    if verbose:
        print('Number of label issues found: {}'.format(sum(label_issues_mask)))
```

cleanlab/filter.py
cgnorthcutt: This should be noted in project tracking and will likely not be addressed in this PR. @JohnsonKuan @jwmueller

cleanlab/models/fasttext.py
cgnorthcutt: I like the idea of moving to example_models for now. Its a good point. Let's add this to project tracking for 2.0.1+

cleanlab/latent_algebra.py
cgnorthcutt: Note that several of the mathematical equalities provided in latent_algebra are pretty impressive results from our PhD work (some solved by friend Tailin Wu) and not all of them published. I agree this is not a core module and we could move it to utils (happy to make this change), but right now it does serve the purpose of showing of some impressive results for those familiar with the field and it will be harder for folks to stumble across those equalities if they move to a utils folder. Given that what do you think about moving `latent_algrebra` to a `utils/` folder now... versus waiting until we add more tasks and things start to get more cluttered?

cleanlab/count.py
cgnorthcutt: I changed `s` to `labels` for the most part everywhere. wAre you proposing keeping the notation as `s` in the context of mathematical formula?

cleanlab/count.py
cgnorthcutt: `find_label_issues()` now prints out number of label issues when `verbose=1`.

`sum(find_label_issues(...))` usually matches `num_label_issues`, but not always. For example, it ignores the `MIN_NUM_PER_CLASS = 1` constraints required when filtering, but most importantly!! its computed **more accurately** than `find_label_issues` -- the reasons is because its computed just based on the trace, ignoring all off-diagonals (which is how find_label_issues works). By simply summing over the diagonals (which are the most accurately computed elements in the joint), you can get a more accurate estimate of the label errors.

This will become the canonical way to find errors simply by combining a ranking/scoring function from `rank.py` with `num_label_issues()`. I'll update the docstring in `num_label_issues()` to reflect this information.

cgnorthcutt: docstring now reads

```
Estimates the number of label issues in `labels`.
    
    This method is **more accurate** than `sum(find_label_issues())` because its computed using only
    the Trace(joint), ignoring all off-diagonals (used by `find_label_issues` and harder to
    estimate). Here we sum over only diagonal elements in the joint (which have more data
    are more constrained, and therefore easier to compute).
    
    tl;dr - Use this method to get the most accurate estimate of number of label issues when you
    don't need the indices of the label issues. This is the canonical way to find errors
    simply by combining a ranking/scoring function from `rank.py` with `num_label_issues()`.

File: cleanlab/count.py [hunk 39]
```diff
@@ -48,12 +47,12 @@
...
 
 
-def num_label_errors(
+def num_label_issues(
         labels,
         psx,
         confident_joint=None,
 ):
-    """Estimates the number of label errors in `labels`.
+    """Estimates the number of label issues in `labels`.
```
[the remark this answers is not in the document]
cgnorthcutt: address above

cleanlab/filter.py
JohnsonKuan: I created a new [Jira ticket ](https://cleanlab.atlassian.net/jira/software/projects/CLOS/boards/3?selectedIssue=CLOS-19) for this so we can revisit.

cleanlab/latent_algebra.py
cgnorthcutt: (i think) i've gotten this fixed everwhere. I left ASCII note in one place in the code, but removed everywhere else.

cleanlab/latent_algebra.py
cgnorthcutt: fixed thanks

cleanlab/filter.py
```diff
@@ -15,53 +15,52 @@
...
-    w = '''If you want to see estimated completion times
-    while running methods in cleanlab.pruning, install tqdm
-    via "pip install tqdm".'''
+    w = '''To see estimated completion times for methods in cleanlab.filter, "pip install tqdm".'''
     warnings.warn(w)
 
-# Leave at least this many examples in each class after
-# pruning, regardless if noise estimates are larger.
-MIN_NUM_PER_CLASS = 5
+# Leave at least 1 example in each class after pruning, regardless if noise estimates are larger.
```
JohnsonKuan, line 56: Should rename "pruning" -> "filtering" everywhere in the comments.

cleanlab/filter.py
JohnsonKuan: I prefer `pred_probs`. Most intuitive name for me.

cleanlab/rank.py
```diff
@@ -0,0 +1,159 @@
...
+    -------
+      label_issues_idx : np.array (int)
+        Return the index integers of the label issues, ordered by
+        the normalized margin."""
+
+    assert (len(psx) == len(labels))
+    # Convert bool mask to index mask
+    label_issues_idx = np.arange(len(labels))[label_issues_mask]
+    # self-confidence is the holdout probability that an example belongs to its given class label
+    self_confidence = np.array(
```
JohnsonKuan, line 65: Can reuse get_self_confidence_for_each_label() function below to get self_confidence.

cgnorthcutt: Good point. Rewrote entire function so now its easily extensible to any new `label quality score` ranking function.

e.g.

```python
    assert (len(pred_probs) == len(labels))
    # Convert bool mask to index mask
    label_issues_idx = np.arange(len(labels))[label_issues_mask]
    pred_probs_er, labels_er = pred_probs[label_issues_mask], labels[label_issues_mask]
    if rank_by == "self_confidence":
        label_quality_scores = get_self_confidence_for_each_label(labels_er, pred_probs_er)    
    elif rank_by == "normalized_margin":        
        label_quality_scores = get_normalized_margin_for_each_label(labels_er, pred_probs_er)
    else:
        raise ValueError(
            'rank_by must be "self_confidence" or "normalized_margin", '
            'but is "' + rank_by + '".'
        )
    return label_issues_idx[np.argsort(label_quality_scores)]
```

cleanlab/filter.py
jwmueller: I like raising that warning when `frac_noise` is detected to be > 1

cleanlab/models/fasttext.py
jwmueller: This would be API breaking change though, so we'd need to make it in 2.0.
We can just do it in a followup PR after this one?

cleanlab/count.py
jwmueller: This part of the docstring may make user feel as if they're missing something:
"This is the canonical way to find errors
simply by combining a ranking/scoring function from `rank.py` with `num_label_issues()`"

Where is an example of this "canonical way" in action?

jwmueller: maybe it's clearer as: 
"Another way to find label issues is to simply let `N = num_label_issues()` and then report the top `N`-ranked results based on label quality score from `rank.py`. The advantages/disadvantages of this alternative vs `find_label_issues()` are ..."

But even then it seems confusing to present two ways of doing the same thing, especially when one way has the canonical name: `find_label_issues()`

cleanlab/filter.py
jwmueller: `pred_probs` is good for me as well. 

Just FYI sklearn uses `y_pred`, cf.  https://scikit-learn.org/stable/modules/generated/sklearn.metrics.log_loss.html

cleanlab/util.py
```diff
@@ -330,11 +327,11 @@ def confusion_matrix(true, pred):
     ----------
     y : np.array 1d
       Contains labels.
-      Assumes s and y contains the same distinct set of labels.
+      Assumes labels and y contains the same distinct set of labels.
 
-    s : np.array 1d
-      Contains labels.
-      Assumes s and y contains the same distinct set of labels.
+    labels : np.array 1d
```
calebchiam, line 332: L316: parameter and doctoring mismatch: (true, pred) -> (y, labels)
L319: typo: represtented

cleanlab/noise_generation.py
```diff
@@ -76,8 +74,8 @@ def noise_matrix_is_valid(noise_matrix, py, verbose=False):
 
 
 def generate_noisy_labels(y, noise_matrix, verbose=False):
```
calebchiam, line 76: 'verbose' argument isn't used

cleanlab/count.py
calebchiam: Yeah, I agree with Jonas that this is potentially confusing. I think the default should be that `num_label_issues() = sum(find_label_issues(...))`. A user might check the number of label issues in a dataset to see if it's high enough to be worth investigating further, and then run `find_label_issues`, only to find that the number of 'promised' issues is fewer than expected.

Perhaps, `num_label_issues` can have a toggle argument `use_trace_joint = False`? That way, experienced users can still get the most accurate count if it's important to them, but the average user won't see a mismatch between `num_label_issues()` and `sum(find_label_issues(...))`.

cleanlab/count.py
jwmueller: I'd rather consider changing `y`, since in other ML packages, `y` would refer to the given labels as well. 

Here are various options off top of my head: 
1. keep `labels` as is (or change to `given_labels`, `y` -> `true_labels`)
2. keep `labels` as is (or change to `labels_given`, `y` -> `labels_true`)
3. ``labels` -> `y_given`, `y` -> `y_true` (for the mathematical helper functions only, keep `labels` as arg-name for eg. `num_label_errors()`).
4. `labels` -> `y_observed`, `y` -> `y_actual` (for the mathematical helper functions only, keep `labels` as arg-name for eg. `num_label_errors()`).

cleanlab/latent_algebra.py
jwmueller: I would move it now to minimize API breaking changes going forward.  Especially if it's just intended for researchers, who you can still point to this functionality from the confidentlearning-reproduce repo README.

Unless there are useful things a data scientist can do based on the `latent_algebra` functionality?  If so, these use-cases would need to be made more clear in the docstrings for this to remain at the top-level and be something the average user can actually leverage.  Actually even if moving these to lower `utils/` level, it could be helpful to add one extra sentence to each docstring explaining what purpose of the function is (ie. why it might be useful).

README.md
```diff
@@ -335,17 +335,17 @@ est_py, est_nm, est_inv, confident_joint = estimate_py_and_noise_matrices_from_p
...
+    labels=noisy_labels,
     psx=probabilities,
     confident_joint=None,  # Provide if you have it already
 )
 ```
 
 ## The Polyplex
 
-The key to learning in the presence of label errors is estimating the joint distribution between the actual, hidden labels ‘*y*’ and the observed, noisy labels ‘*s*’. Using `cleanlab` and the theory of confident learning, we can completely characterize the trace of the latent joint distribution, *trace(P(s,y))*, given *p(y)*, for any fraction of label errors, i.e. for any trace of the noisy channel, *trace(P(s|y))*.
+The key to learning in the presence of label issues is estimating the joint distribution between the actual, hidden labels ‘*y*’ and the observed, noisy labels ‘*s*’. Using `cleanlab` and the theory of confident learning, we can completely characterize the trace of the latent joint distribution, *trace(P(s,y))*, given *p(y)*, for any fraction of label errors, i.e. for any trace of the noisy channel, *trace(P(s|y))*.
```
jwmueller, line 348: ff

README.md
```diff
@@ -335,17 +335,17 @@ est_py, est_nm, est_inv, confident_joint = estimate_py_and_noise_matrices_from_p
...
 
 You can check out how to do this yourself here:
 1. [Drawing Polyplices](https://github.com/cleanlab/examples/blob/master/drawing_polyplices.ipynb)
```
jwmueller, line 351: Move polyplex tutorial/discussion to confidentlearning-reproduce repo?  Or into internal readme page, eg. in the `utils/` folder?

cleanlab/classification.py
```diff
@@ -73,7 +74,7 @@ def score(self, X, y, sample_weight = None):
 Note
 ----
 
-* `s` - denotes *noisy labels*. This is just dataset labels, maybe with errors.
+* `labels` - denotes *noisy labels*. This is just dataset labels, maybe with errors.
```
jwmueller, line 77: "dataset labels"
-> 
"given labels in the original dataset"

cleanlab/classification.py
```diff
@@ -100,14 +101,14 @@ def score(self, X, y, sample_weight = None):
...
     compute_noise_matrix_from_inverse,
 )
-from cleanlab.pruning import get_noise_indices
+from cleanlab.filter import find_label_issues
 
 
 class LearningWithNoisyLabels(BaseEstimator):  # Inherits sklearn classifier
     """Automated learning with noisy labels using any model.
 
     Confident Learning is the state-of-the-art (Northcutt et al., 2021) for
```
jwmueller, line 109: This docstring seems like an odd place to define confident learning and talk about its miscellaneous capabilities for eg. finding label issues, uncertainty estimation, etc.  Shouldn't the docstring describe what this class does instead? Ie. it produces a new version of the same model as `clf` which has been trained in a way that is more robust to label noise in the data.

cleanlab/classification.py
```diff
@@ -244,28 +250,28 @@ def fit(
         """This method implements the confident learning. It counts examples
         that are likely labeled correctly and incorrectly and uses their ratio
         to create a predicted confusion matrix.
-        This function fits the classifier (self.clf) to (X, s) accounting for
+        This function fits the classifier (self.clf) to (X, labels) accounting for
         the noise in both the positive and negative sets.
```
jwmueller, line 254: After the first sentence, it's again a bit unclear that the latter docstring sentences are describing what this method does. Should docstring instead explain what this function does procedurally? (i.e. train `clf` on given data, run CL to find label issues, delete data with label issues, retrain `clf` on cleaned data to produce more robust version of the same classifier).

jwmueller: Also unclear why "positive and negative sets" is mentioned in docstring. For some reason lots of the documentation seems to specifically mention binary (eg. "a binary vector") when this function works for multiclass classification.

cleanlab/classification.py
```diff
@@ -244,28 +250,28 @@ def fit(
...
 
         Parameters
         ----------
         X : :obj:`np.array`
           Input feature matrix (N, D), 2D numpy array
 
-        s : :obj:`np.array`
-          A binary vector of labels, s, which may contain mislabeling.
+        labels : :obj:`np.array`
+          A binary vector of labels, labels, which may contain mislabeling.
```
jwmueller: why binary

cleanlab/classification.py
```diff
@@ -121,10 +122,10 @@ class LearningWithNoisyLabels(BaseEstimator):  # Inherits sklearn classifier
     for their training label.
 
     Given any classifier having the predict_proba() method, an input feature
-    matrix, X, and a discrete vector of labels, s, which may contain
+    matrix, X, and a discrete vector of labels, labels, which may contain
     mislabeling, Confident Learning estimates the classifications that would
     be obtained if the hidden, true labels, y, had instead been provided to
```
jwmueller: Elsewhere you sometimes refer to true label as "hidden label" (eg. L279). 
I think this is confusing and would replace these with "true label" instead of "hidden label"

cleanlab/util.py
anishathalye: This looks like it was search-and-replaced by a bad refactoring tool (lots of `s` -> `labels` in comments)

cleanlab/example_models/README.md
```diff
@@ -9,7 +9,7 @@ The dependencies are as follows
 * mnist_pytorch.py - used for training a simplified AlexNet on MNIST using PyTorch
    - pytorch
    - torchvision
-* fastext.py - used for supervised learning on text classification using FastText
-   - fastext
+* fasttext.py - used for supervised learning on text classification using FastText
+   - fasttext
 
-You must install these dependencies on your own if you wish to use them. cleanlab will not force you to install these large deep learning frameworks because that may potentially clobber your code base. For example if you have both TensorFlow and PyTorch installed, issues sometims arise.
+You must install these dependencies on your own if you wish to use them. cleanlab will not force you to install these large deep learning frameworks because that may potentially clobber your code base. For example if you have both TensorFlow and PyTorch installed, issues sometimes arise.
```
anishathalye, line 15: nit: I think "clobber" isn't the right word to use here.
"may potentially clobber your code base." -> "may potentially install conflicting dependencies."

cleanlab/filter.py
```diff
@@ -0,0 +1,687 @@
...
+
+    w = '''To see estimated completion times for methods in cleanlab.filter, "pip install tqdm".'''
+    warnings.warn(w)
+
+# Leave at least 1 example in each class after filtering/pruning, even if noise estimates are larger
+MIN_NUM_PER_CLASS = 1
+
+# For python 2/3 compatibility, define pool context manager
+# to support the 'with' statement in Python 2
+if sys.version_info[0] == 2:
```
anishathalye, line 60: Don't need this block anymore; can get rid of the whole if/else and replace `multiprocessing_context` with `multiprocessing.Pool`

anishathalye (commented):
In some functions, we are switching parameter order in this release. Related to that: can we switch to using kw-only arguments for a lot of functionality where we don't expect people to use positional arguments? E.g. the signature of `find_label_issues` can be `def find_label_issues(labels, pred_probs, *, confident_joint=None, ...)`. This way, we'll have freedom to move things around / insert parameters as we please, and we won't break any client code.

There's a bunch of weird issues in the diff, probably from a search-and-replace or from an imprecise refactoring tool. I highlighted a couple of them. As we get closer to finalizing this PR, we should take a close line-by-line look at the diff to make sure we haven't missed any.

cleanlab/classification.py
```diff
@@ -447,14 +449,17 @@ def score(self, X, y, sample_weight=None):
 
         if hasattr(self.clf, 'score'):
 
-            # Check if sample_weight in clf.score(). Compatible with Python 2/3.
+            # Check if sample_weight in clf.score(). Not compatible with Python 2.
             if hasattr(inspect, 'getfullargspec') and 'sample_weight' in \
```
anishathalye, line 453: We no longer need this if/else, because Python 3.6+ has `getfullargspec`.

tests/test_classification.py
```diff
@@ -30,14 +25,12 @@
 import warnings
 import pytest
 
-
 # Ignore this warning. getargspec is necessary to support old versions.
 ARGSPEC_WARNING = "DeprecationWarning: inspect.getargspec() is deprecated" \
                   " since Python 3.0, use inspect.signature() or" \
                   " inspect.getfullargspec()"
 warnings.filterwarnings("ignore", message=ARGSPEC_WARNING)
```
anishathalye, line 32: Don't need this anymore, we can migrate to `getfullargspec`.

cleanlab/classification.py
```diff
@@ -328,84 +338,76 @@ class label.
...
-        # Check if sample_weight in clf.fit(). Compatible with Python 2/3.
+        self.label_issues_mask = find_label_issues(labels, pred_probs, confident_joint=self.confident_joint,
+                                            filter_by=self.filter_by, n_jobs=self.n_jobs)
+
+        x_mask = ~self.label_issues_mask
+        x_cleaned = X[x_mask]
+        labels_cleaned = labels[x_mask]
+
+        # Check if sample_weight in clf.fit(). Not compatible with Python 2.
         if hasattr(inspect, 'getfullargspec') and \
```
anishathalye, line 397: On 3.6+, this check will always return true.

cleanlab/count.py
cgnorthcutt: Leaving as is for the following reasons:
* default behavior should be the most accurate, which it currently is.
* default behavior should try to be the fastest approach, which it currently is
* `cleanlab` supports multiple approaches to things throughout (e.g. rank_methods, filter_methods, etc). We can also support multiple approaches to find the threshold cut-off for number of label errors in the dataset.

cleanlab/count.py
cgnorthcutt: I didn't use `y` because `y` implies `true labels` in most packages and throughout some of the literature.

cleanlab/count.py
cgnorthcutt: There are two possible workflows:
1. 
```
from cleanlab import filter
label_issues_bool = find_label_issues()
# Do whatever
```

2. 
```
from cleanlab import rank, count
label_quality_scores = rank.get_normalized_margin_for_each_label(labels, probs)
num_errors = count.num_label_issues()
# Label issues are the sorted top num_errors lowest scores

File: cleanlab/latent_algebra.py (hunk 38 above)
Comment (cgnorthcutt) [reply cleanlab/latent_algebra.py:18]: done

File: cleanlab/util.py (hunk 42 above)
Comment (cgnorthcutt) [reply cleanlab/util.py:332]: nice catch. thanks

File: cleanlab/count.py (hunk 37 above)
Comment (cgnorthcutt) [reply cleanlab/count.py]: changing to `p(label=k)` instead of `p(labels=k)` and `p(true_label=k)` instead of `p(true_labels=k)` in mathematical definitions like these. in functions i kept as `labels` and in the very few functions with `y` i changed to `true_labels` (plural).

Comment (cgnorthcutt) [reply cleanlab/count.py]: I think this is done. Worth double checking.

File: README.md (hunk 45 above)
Comment (cgnorthcutt) [reply README.md:350]: just removed

File: cleanlab/classification.py (hunk 47 above)
Comment (cgnorthcutt) [reply cleanlab/classification.py:109]: done

File: cleanlab/classification.py (hunk 46 above)
Comment (cgnorthcutt) [reply cleanlab/classification.py:77]: done

File: cleanlab/classification.py (hunk 48 above)
Comment (cgnorthcutt) [reply cleanlab/classification.py:254]: originally wrote this for binary. fixing and added more xplanation

File: cleanlab/classification.py (hunk 49 above)
Comment (cgnorthcutt) [reply cleanlab/classification.py]: old stuff from rankpruning left in here. removed.

File: README.md [hunk 56]
```diff
@@ -65,24 +65,24 @@ Past release notes and **future features planned** is available [here](https://g
...
+2.  **fast** - Non-iterative, parallelized algorithms (e.g. < 1 second to find label issues in ImageNet with pre-computed probabilities)
 3.  **general** - Works with any dataset, model, and framework, e.g., Tensorflow, PyTorch, sklearn, xgboost, etc.
 
-## Find label errors with PyTorch, Tensorflow, sklearn, xgboost, etc. in 1 line of code
+## Find label issues with PyTorch, Tensorflow, sklearn, xgboost, etc. in 1 line of code
 
 ```python
-# Compute psx (n x m matrix of predicted probabilities) on your own, with any classifier.
-# Here is an example that shows in detail how to compute psx on CIFAR-10:
+# Compute pred_probs (n x m matrix of predicted probabilities) on your own, with any classifier.
```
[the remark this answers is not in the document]
cgnorthcutt: oos_pred_probs? or pred_probs_cv? or heldout_pred_probs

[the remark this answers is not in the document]
cgnorthcutt: i might want to skip this change. The comment just below that line says
`# Be sure you compute probs in a holdout/out-of-sample manner (e.g. via cross-validation)`

README.md
```diff
@@ -65,24 +65,24 @@ Past release notes and **future features planned** is available [here](https://g
...
-    psx=numpy_array_of_predicted_probabilities,
-    sorted_index_method='normalized_margin', # Orders label errors
+# Now finding label issues is trivial with cleanlab... its one line of code.
+# label issues are ordered by likelihood of being an error. First index is most likely error.
+from cleanlab.filter import find_label_issues
+
+ordered_label_issues = find_label_issues(
+    labels=numpy_array_of_noisy_labels,
+    pred_probs=numpy_array_of_predicted_probabilities,
+    return_indices_ranked_by='normalized_margin', # Orders label issues
```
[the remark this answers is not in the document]
cgnorthcutt: In our previous discussion, we agreed that we should mention that this changes the format of the return output from a bool mask to indices and hence it was a good idea to include `return_indices...ranked_by` versus just `rank_by` which doesn't indicate clearly the return type is changing (and will be a different length as well)

cleanlab/coteaching.py
```diff
@@ -63,7 +58,7 @@ def loss_coteaching(
       List of Noisy Labels (t means targets)
 
     forget_rate : float
-      Decimal between 0 and 1 for how quickly the models forget what they learn.
+      Decimal between 0 and 1 for how quickly the example_models forget what they learn.
```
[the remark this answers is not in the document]
cgnorthcutt: fixed everywhere

PR #129, opened 2022-03-12
Add EditorConfig

anishathalye:
This will help maintain consistent coding styles across various editors and IDEs. See <https://editorconfig.org/>.

It helps with little things like making sure Windows users don't commit CRLF line endings and making sure people don't have extra whitespace at the end of lines.


jwmueller (approved):
LGTM

PR #131, opened 2022-03-18
Add label quality scoring functions and user API to choose the method

JohnsonKuan:
This PR adds new label quality scoring capabilities:

- Add "confidence_weighted_entropy" as new label quality score that performed well in internal benchmarks to detect label issues.
- Add option for users to adjust pred_probs by subtracting class thresholds (as defined in confident learning paper) and renormalizing. This improved label issue detection in internal benchmarks.
- Add user API to score label quality by choosing the scoring method.
- Add user API to score label quality for ensemble models where we have a list of pred_probs.
- Add rank_by_kwargs dict to accept keyword args for find_label_issues() and order_label_issues(). This allows us to pass the optional adj_pred_probs arg to adjust pred_probs by subtracting class thresholds. This also allows us to add more optional args in the future for scoring and ranking.

This PR also includes:

- New tests for CI to test the new label quality scoring functions
- Refactor order_label_issues()

**Context:**

Originally was planning to add more scores from the internal benchmark study. However, it was decided that we would exclude the active learning scores for now due to poor performance in internal benchmarks. We can create a separate PR in the future to add the active learning scores.

### DIFF
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -534,7 +534,7 @@ def find_label_issues(
             labels=labels,
             pred_probs=pred_probs,
             rank_by=return_indices_ranked_by,
-            **rank_by_kwargs,
+            rank_by_kwargs=rank_by_kwargs,
         )
         return er
     confident_joint
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -342,13 +342,9 @@ def find_label_issues(
       If 0, no print statements. If 1, prints when multiprocessing happens.
 
     rank_by_kwargs : dict
-      Optional keyword arguments to pass into scoring functions for ranking.
+      Optional keyword arguments to pass into `rank.score_label_quality()` method.
       Accepted args include:
         adj_pred_probs : bool, default = False
-          Adjust predicted probabilities by subtracting the class confident thresholds and renormalizing.
-          The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
-          See paper "Confident Learning: Estimating Uncertainty in Dataset Labels" by Northcutt et al.
-          https://arxiv.org/abs/1911.00068
 
     Returns
     -------
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -243,6 +243,7 @@ def find_label_issues(
     num_to_remove_per_class=None,
     n_jobs=None,
     verbose=0,
+    **rank_by_kwargs,
 ):
     """By default, this method returns a boolean mask for the entire dataset where True represents
     a label issue and False represents an example that is confidently/accurately labeled.
@@ -523,6 +524,7 @@ def find_label_issues(
             labels=labels,
             pred_probs=pred_probs,
             rank_by=return_indices_ranked_by,
+            **rank_by_kwargs,
         )
         return er
     confident_joint
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -308,7 +308,7 @@ def find_label_issues(
 
     rank_by_kwargs : dict
       Optional keyword arguments to pass into scoring functions for ranking by label quality score
-      (see: `score_label_quality()` in cleanlab/rank.py).
+      (see: `get_label_quality_scores()` in cleanlab/rank.py).
       Accepted args include:
         adj_pred_probs : bool, default = False
 
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -310,7 +310,7 @@ def find_label_issues(
       Optional keyword arguments to pass into scoring functions for ranking by label quality score
       (see: `get_label_quality_scores()` in cleanlab/rank.py).
       Accepted args include:
-        adj_pred_probs : bool, default = False
+      adj_pred_probs : bool, default = False
 
     multi_label : bool
       If true, labels should be an iterable (e.g. list) of iterables, containing a
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -243,7 +243,7 @@ def find_label_issues(
     num_to_remove_per_class=None,
     n_jobs=None,
     verbose=0,
-    **rank_by_kwargs,
+    rank_by_kwargs={},
 ):
     """By default, this method returns a boolean mask for the entire dataset where True represents
     a label issue and False represents an example that is confidently/accurately labeled.
@@ -339,6 +339,11 @@ def find_label_issues(
 
     verbose : int
       If 0, no print statements. If 1, prints when multiprocessing happens.
+
+    rank_by_kwargs : dict
+      Optional keyword arguments to pass into scoring functions for ranking.
+      Accepted args includes:
+        adj_pred_probs : bool, default = False
 
     Returns
     -------
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -238,12 +238,12 @@ def find_label_issues(
     confident_joint=None,
     filter_by="prune_by_noise_rate",
     return_indices_ranked_by=None,
+    rank_by_kwargs={},
     multi_label=False,
     frac_noise=1.0,
     num_to_remove_per_class=None,
     n_jobs=None,
     verbose=0,
-    rank_by_kwargs={},
 ):
     """By default, this method returns a boolean mask for the entire dataset where True represents
     a label issue and False represents an example that is confidently/accurately labeled.
@@ -306,6 +306,12 @@ def find_label_issues(
       ``'self_confidence' := [pred_probs[i][labels[i]] for i in label_issues_idx]``
       ``'confidence_weighted_entropy' := entropy(pred_probs) / self_confidence``
 
+    rank_by_kwargs : dict
+      Optional keyword arguments to pass into scoring functions for ranking by label quality score
+      (see: `score_label_quality()` in cleanlab/rank.py).
+      Accepted args include:
+        adj_pred_probs : bool, default = False
+
     multi_label : bool
       If true, labels should be an iterable (e.g. list) of iterables, containing a
       list of labels for each example, instead of just a single label.
@@ -340,11 +346,6 @@ def find_label_issues(
 
     verbose : int
       If 0, no print statements. If 1, prints when multiprocessing happens.
-
-    rank_by_kwargs : dict
-      Optional keyword arguments to pass into `rank.score_label_quality()` method.
-      Accepted args include:
-        adj_pred_probs : bool, default = False
 
     Returns
     -------
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -298,12 +298,13 @@ def find_label_issues(
       5. 'predicted_neq_given': Find examples where the predicted class
       (i.e. argmax of the predicted probabilities) does not match the given label.
 
-    return_indices_ranked_by : {:obj:`None`, :obj:`self_confidence`, :obj:`normalized_margin`}
+    return_indices_ranked_by : {:obj:`None`, :obj:`self_confidence`, :obj:`normalized_margin`, :obj:`confidence_weighted_entropy`}
       If None, returns a boolean mask (true if example at index is label error)
       If not None, returns an array of the label error indices
       (instead of a bool mask) where error indices are ordered by the either:
       ``'normalized_margin' := normalized margin (p(label = k) - max(p(label != k)))``
       ``'self_confidence' := [pred_probs[i][labels[i]] for i in label_issues_idx]``
+      ``'confidence_weighted_entropy' := entropy(pred_probs) / self_confidence``
 
     multi_label : bool
       If true, labels should be an iterable (e.g. list) of iterables, containing a
@@ -342,8 +343,12 @@ def find_label_issues(
 
     rank_by_kwargs : dict
       Optional keyword arguments to pass into scoring functions for ranking.
-      Accepted args includes:
+      Accepted args include:
         adj_pred_probs : bool, default = False
+          Adjust predicted probabilities by subtracting the class confident thresholds and renormalizing.
+          The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
+          See paper "Confident Learning: Estimating Uncertainty in Dataset Labels" by Northcutt et al.
+          https://arxiv.org/abs/1911.00068
 
     Returns
     -------
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -310,7 +310,7 @@ def find_label_issues(
       Optional keyword arguments to pass into scoring functions for ranking by label quality score
       (see: `get_label_quality_scores()` in cleanlab/rank.py).
       Accepted args include:
-      adj_pred_probs : bool, default = False
+      adjust_pred_probs : bool, default = False
 
     multi_label : bool
       If true, labels should be an iterable (e.g. list) of iterables, containing a
--- a/cleanlab/rank.py
+++ b/cleanlab/rank.py
@@ -43,7 +43,6 @@ def order_label_issues(
 
     Default label quality score is "normalized margin".
     See https://arxiv.org/pdf/1810.05369.pdf (eqn 2.2)
-    e.g. normalized_margin = prob_label - max_prob_not_label
 
     Parameters
     ----------
@@ -68,19 +67,17 @@ def order_label_issues(
         'confidence_weighted_entropy' := entropy(pred_probs) / self_confidence
 
     rank_by_kwargs : dict
-      Optional keyword arguments to pass into scoring functions for ranking.
+      Optional keyword arguments to pass into `score_label_quality()` method.
       Accepted args include:
         adj_pred_probs : bool, default = False
-          Adjust predicted probabilities by subtracting the class confident thresholds and renormalizing.
-          The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
-          See paper "Confident Learning: Estimating Uncertainty in Dataset Labels" by Northcutt et al.
-          https://arxiv.org/abs/1911.00068
 
     Returns
     -------
     label_issues_idx : np.array (int)
-      Return the index integers of the label issues, ordered by
-      the normalized margin."""
+      Return the index integers of the label issues, ordered by the label-quality scoring method
+      passed to rank_by.
+
+    """
 
     assert len(pred_probs) == len(labels)
 
@@ -96,243 +93,20 @@ def order_label_issues(
     return label_issues_idx[np.argsort(label_quality_scores_issues)]
 
 
-def get_self_confidence_for_each_label(
-    labels: np.array,
-    pred_probs: np.array,
-) -> np.array:
-    """Returns the "self-confidence" for every example in the dataset associated pred_probs
-    and labels. The self-confidence is the holdout probability that an example belongs to
-    its given class label.
-
-    Score is between 0 and 1.
-    1 - clean label (not an error).
-    0 - dirty label (label error).
-
-    Parameters
-    ----------
-    labels : np.array
-      A discrete vector of noisy labels, i.e. some labels may be erroneous.
-      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
-
-    pred_probs : np.array (shape (N, K))
-      P(label=k|x) is a matrix with K model-predicted probabilities.
-      Each row of this matrix corresponds to an example x and contains the model-predicted
-      probabilities that x belongs to each possible class.
-      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
-      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
-
-    Returns
-    -------
-    self_confidence : np.array (float)
-      Return the holdout probability that each example in pred_probs belongs to its
-      label. Assumes pred_probs is computed holdout/out-of-sample."""
-
-    # np.mean is used so that this works for multi-labels (list of lists)
-    self_confidence = np.array([np.mean(pred_probs[i, l]) for i, l in enumerate(labels)])
-    return self_confidence
-
-
-def get_normalized_margin_for_each_label(
-    labels: np.array,
-    pred_probs: np.array,
-) -> np.array:
-    """Returns the "normalized margin" for every example associated pred_probs and
-    labels.
-    The normalized margin is (p(label = k) - max(p(label != k))), i.e. the probability
-    of the given label minus the probability of the argmax label that is not
-    the given label. This gives you an idea of how likely an example is BOTH
-    its given label AND not another label, and therefore, scores its likelihood
-    of being a good label or a label error.
-
-    Score is between 0 and 1.
-    1 - clean label (not an error).
-    0 - dirty label (label error).
-
-    Parameters
-    ----------
-
-    labels : np.array
-      A discrete vector of noisy labels, i.e. some labels may be erroneous.
-      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
-
-    pred_probs : np.array (shape (N, K))
-      P(label=k|x) is a matrix with K model-predicted probabilities.
-      Each row of this matrix corresponds to an example x and contains the model-predicted
-      probabilities that x belongs to each possible class.
-      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
-      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
-
-    Returns
-    -------
-    normalized_margin : np.array (float)
-      Return a score (between 0 and 1) for each example of its likelihood of
-      being correctly labeled. Assumes pred_probs is computed holdout/out-of-sample.
-      normalized_margin = prob_label - max_prob_not_label"""
-
-    self_confidence = get_self_confidence_for_each_label(labels, pred_probs)
-    max_prob_not_label = np.array(
-        [max(np.delete(pred_probs[i], l, -1)) for i, l in enumerate(labels)]
-    )
-    normalized_margin = (self_confidence - max_prob_not_label + 1) / 2
-    return normalized_margin
-
-
-def get_confidence_weighted_entropy_for_each_label(
-    labels: np.array, pred_probs: np.array
-) -> np.array:
-    """Returns the normalized entropy divided by "self-confidence".
-
-    Score is between 0 and 1.
-    1 - clean label (not an error).
-    0 - dirty label (label error).
-
-    Parameters
-    ----------
-    labels : np.array
-      A discrete vector of noisy labels, i.e. some labels may be erroneous.
-      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
-
-    pred_probs : np.array (shape (N, K))
-      P(label=k|x) is a matrix with K model-predicted probabilities.
-      Each row of this matrix corresponds to an example x and contains the model-predicted
-      probabilities that x belongs to each possible class.
-      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
-      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
-
-    Returns
-    -------
-    confidence_weighted_entropy : np.array (float)
-      Return a score (between 0 and 1) for each example of its likelihood of
-      being correctly labeled. Assumes pred_probs is computed holdout/out-of-sample.
-
-    """
-
-    # Get confidence of given label
-    self_confidence = get_self_confidence_for_each_label(labels, pred_probs)
-
-    # Divide entropy by self confidence
-    confidence_weighted_entropy = get_entropy(**{"pred_probs": pred_probs}) / self_confidence
-
-    # Rescale
-    confidence_weighted_entropy = (
-        np.log(confidence_weighted_entropy + 1) / confidence_weighted_entropy
-    )
-
-    return confidence_weighted_entropy
-
-
-def get_entropy(pred_probs: np.array) -> np.array:
-    """Returns the normalized entropy of pred_probs.
-
-    Read more about normalized entropy here: https://en.wikipedia.org/wiki/Entropy_(information_theory)
-
-    Normalized entropy is used in active learning for uncertainty sampling: https://towardsdatascience.com/uncertainty-sampling-cheatsheet-ec57bc067c0b
-
-    Normalized entropy is between 0 and 1. Higher values of entropy indicate higher uncertainty in the model's prediction of the correct label.
-
-    Parameters
-    ----------
-    pred_probs : np.array (shape (N, K))
-      P(label=k|x) is a matrix with K model-predicted probabilities.
-      Each row of this matrix corresponds to an example x and contains the model-predicted
-      probabilities that x belongs to each possible class.
-      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
-      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
-
-    Returns
-    -------
-    entropy : np.array (float)
-
-    """
-
-    num_classes = pred_probs.shape[1]
-
-    # Note that dividing by log(num_classes) changes the base of the log which rescales entropy to 0-1 range
-    return -np.sum(pred_probs * np.log(pred_probs), axis=1) / np.log(num_classes)
-
-
-def subtract_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array:
-    """Returns adjusted predicted probabilities by subtracting the class confident thresholds and renormalizing.
-
-    The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
-
-    See paper "Confident Learning: Estimating Uncertainty in Dataset Labels" by Northcutt et al.
-    https://arxiv.org/abs/1911.00068
-
-    The purpose of this adjustment is to handle class imbalance.
-
-    Parameters
-    ----------
-    labels : np.array
-      A discrete vector of noisy labels, i.e. some labels may be erroneous.
-      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
-
-    pred_probs : np.array (shape (N, K))
-      P(label=k|x) is a matrix with K model-predicted probabilities.
-      Each row of this matrix corresponds to an example x and contains the model-predicted
-      probabilities that x belongs to each possible class.
-      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
-      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
-
-    Returns
-    -------
-    pred_probs_adj : np.array (float)
-    """
-
-    # Get expected (average) self-confidence for each class
-    confident_thresholds = get_confident_thresholds(labels, pred_probs)
-
-    # Subtract the class confident thresholds
-    pred_probs_adj = pred_probs - confident_thresholds
-
-    # Renormalize by shifting data to take care of negative values from the subtraction
-    pred_probs_adj += 1
-    pred_probs_adj /= pred_probs_adj.sum(axis=1)[:, None]
-
-    return pred_probs_adj
-
-
-def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array:
-    """Returns expected (average) "self-confidence" for each class.
-
-    The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
-
-    See paper "Confident Learning: Estimating Uncertainty in Dataset Labels" by Northcutt et al.
-    https://arxiv.org/abs/1911.00068
-
-    Parameters
-    ----------
-    labels : np.array
-      A discrete vector of noisy labels, i.e. some labels may be erroneous.
-      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
-
-    pred_probs : np.array (shape (N, K))
-      P(label=k|x) is a matrix with K model-predicted probabilities.
-      Each row of this matrix corresponds to an example x and contains the model-predicted
-      probabilities that x belongs to each possible class.
-      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
-      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
-
-    Returns
-    -------
-    confident_thresholds : np.array (shape (K,))
-
-    """
-    confident_thresholds = np.array(
-        [np.mean(pred_probs[:, k][labels == k]) for k in range(pred_probs.shape[1])]
-    )
-    return confident_thresholds
-
-
 def score_label_quality(
     labels: np.array,
     pred_probs: np.array,
     method: str = "self_confidence",
     adj_pred_probs: bool = False,
 ) -> np.array:
-    """Returns the label quality scores.
+    """Returns label quality scores for each datapoint.
 
-    The scoring methods are heuristics to enable users to quickly rank order data to find potential label quality issues.
+    This is a function to compute label-quality scores for classification datasets,
+    where lower scores indicate labels less likely to be correct.
+
+    Score is between 0 and 1.
+    1 - clean label (given label is likely correct).
+    0 - dirty label (given label is likely incorrect).
 
     Parameters
     ----------
@@ -356,14 +130,17 @@ def score_label_quality(
         :func:`confidence_weighted_entropy`
 
     adj_pred_probs : bool, default = True
-      Adjust predicted probabilities by subtracting the class confident thresholds and renormalizing.
-      The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
+      Account for class-imbalance in the label-quality scoring (by adjusting predicted probabilities
+      via subtraction of class confident thresholds and renormalization).
+      Set this = False if you prefer to emphasize label issues in classes that are more common
+      throughout the dataset
       See paper "Confident Learning: Estimating Uncertainty in Dataset Labels" by Northcutt et al.
       https://arxiv.org/abs/1911.00068
 
     Returns
     -------
     label_quality_scores : np.array (float)
+      Scores are between 0 and 1 where lower scores indicate labels less likely to be correct.
 
     See Also
     --------
@@ -411,25 +188,25 @@ def score_label_quality_ensemble(
     method: str = "self_confidence",
     adj_pred_probs: bool = False,
 ) -> np.array:
-    """Returns the ensemble label quality scores.
+    """Returns label quality scores based on predictions from an ensemble of models.
 
     This requires a list of pred_probs from each model in the ensemble.
 
-    The scoring methods are heuristics to enable users to quickly rank order data to find potential label quality issues.
+    This is a function to compute label-quality scores for classification datasets,
+    where lower scores indicate labels less likely to be correct.
+
+    Score is between 0 and 1.
+    1 - clean label (given label is likely correct).
+    0 - dirty label (given label is likely incorrect).
 
     Parameters
     ----------
     labels : np.array
-      A discrete vector of noisy labels, i.e. some labels may be erroneous.
-      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
+      Labels in the same format expected by the `score_label_quality()` method.
 
     pred_probs_list : List of np.array (shape (N, K))
-      List of pred_probs from each model in the ensemble.
-      P(label=k|x) is a matrix with K model-predicted probabilities.
-      Each row of this matrix corresponds to an example x and contains the model-predicted
-      probabilities that x belongs to each possible class.
-      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
-      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
+      Each element in this list should be an array of pred_probs in the same format
+      expected by the `score_label_quality()` method
 
     method : {"self_confidence", "normalized_margin", "confidence_weighted_entropy"}, default="self_confidence"
       Label quality scoring method. Default is "self_confidence".
@@ -440,10 +217,7 @@ def score_label_quality_ensemble(
         :func:`confidence_weighted_entropy`
 
     adj_pred_probs : bool, default = True
-      Adjust predicted probabilities by subtracting the class confident thresholds and renormalizing.
-      The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
-      See paper "Confident Learning: Estimating Uncertainty in Dataset Labels" by Northcutt et al.
-      https://arxiv.org/abs/1911.00068
+      Adj_pred_probs in the same format expected by the `score_label_quality()` method.
 
     Returns
     -------
@@ -481,3 +255,211 @@ def score_label_quality_ensemble(
     label_quality_scores = (np.vstack(scores_list).T * weights).sum(axis=1)
 
     return label_quality_scores
+
+
+def get_self_confidence_for_each_label(
+    labels: np.array,
+    pred_probs: np.array,
+) -> np.array:
+    """Returns the self-confidence label-quality score for each datapoint.
+
+    This is a function to compute label-quality scores for classification datasets,
+    where lower scores indicate labels less likely to be correct.
+
+    The self-confidence is the holdout probability that an example belongs to
+    its given class label.
+
+    Parameters
+    ----------
+    labels : np.array
+      Labels in the same format expected by the `score_label_quality()` method.
+
+    pred_probs : np.array (shape (N, K))
+      Predicted-probabilities in the same format expected by the `score_label_quality()` method.
+
+    Returns
+    -------
+    label_quality_scores : np.array (float)
+      Return the holdout probability that each example in pred_probs belongs to its
+      label.
+
+    """
+
+    # np.mean is used so that this works for multi-labels (list of lists)
+    label_quality_scores = np.array([np.mean(pred_probs[i, l]) for i, l in enumerate(labels)])
+    return label_quality_scores
+
+
+def get_normalized_margin_for_each_label(
+    labels: np.array,
+    pred_probs: np.array,
+) -> np.array:
+    """Returns the "normalized margin" label-quality score for each datapoint.
+
+    This is a function to compute label-quality scores for classification datasets,
+    where lower scores indicate labels less likely to be correct.
+
+    The normalized margin is (p(label = k) - max(p(label != k))), i.e. the probability
+    of the given label minus the probability of the argmax label that is not
+    the given label. This gives you an idea of how likely an example is BOTH
+    its given label AND not another label, and therefore, scores its likelihood
+    of being a good label or a label error.
+
+    Parameters
+    ----------
+
+    labels : np.array
+      Labels in the same format expected by the `score_label_quality()` method.
+
+    pred_probs : np.array (shape (N, K))
+      Predicted-probabilities in the same format expected by the `score_label_quality()` method.
+
+    Returns
+    -------
+    label_quality_scores : np.array (float)
+      Return a score (between 0 and 1) for each example of its likelihood of
+      being correctly labeled.
+      normalized_margin = prob_label - max_prob_not_label
+
+    """
+
+    self_confidence = get_self_confidence_for_each_label(labels, pred_probs)
+    max_prob_not_label = np.array(
+        [max(np.delete(pred_probs[i], l, -1)) for i, l in enumerate(labels)]
+    )
+    label_quality_scores = (self_confidence - max_prob_not_label + 1) / 2
+    return label_quality_scores
+
+
+def get_confidence_weighted_entropy_for_each_label(
+    labels: np.array, pred_probs: np.array
+) -> np.array:
+    """Returns the "confidence weighted entropy" label-quality score for each datapoint.
+
+    This is a function to compute label-quality scores for classification datasets,
+    where lower scores indicate labels less likely to be correct.
+
+    "confidence weighted entropy" is the normalized entropy divided by "self-confidence".
+
+    Parameters
+    ----------
+    labels : np.array
+      Labels in the same format expected by the `score_label_quality()` method.
+
+    pred_probs : np.array (shape (N, K))
+      Predicted-probabilities in the same format expected by the `score_label_quality()` method.
+
+    Returns
+    -------
+    label_quality_scores : np.array (float)
+      Return a score (between 0 and 1) for each example of its likelihood of
+      being correctly labeled.
+
+    """
+
+    self_confidence = get_self_confidence_for_each_label(labels, pred_probs)
+
+    # Divide entropy by self confidence
+    label_quality_scores = get_entropy(**{"pred_probs": pred_probs}) / self_confidence
+
+    # Rescale
+    label_quality_scores = np.log(label_quality_scores + 1) / label_quality_scores
+
+    return label_quality_scores
+
+
+def get_entropy(pred_probs: np.array) -> np.array:
+    """Returns the normalized entropy of pred_probs.
+
+    Read more about normalized entropy here: https://en.wikipedia.org/wiki/Entropy_(information_theory)
+
+    Normalized entropy is used in active learning for uncertainty sampling: https://towardsdatascience.com/uncertainty-sampling-cheatsheet-ec57bc067c0b
+
+    Normalized entropy is between 0 and 1. Higher values of entropy indicate higher uncertainty in the model's prediction of the correct label.
+
+    Parameters
+    ----------
+    pred_probs : np.array (shape (N, K))
+      P(label=k|x) is a matrix with K model-predicted probabilities.
+      Each row of this matrix corresponds to an example x and contains the model-predicted
+      probabilities that x belongs to each possible class.
+      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
+      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
+
+    Returns
+    -------
+    entropy : np.array (float)
+
+    """
+
+    num_classes = pred_probs.shape[1]
+
+    # Note that dividing by log(num_classes) changes the base of the log which rescales entropy to 0-1 range
+    return -np.sum(pred_probs * np.log(pred_probs), axis=1) / np.log(num_classes)
+
+
+def subtract_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array:
+    """Returns adjusted predicted probabilities by subtracting the class confident thresholds and renormalizing.
+
+    The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
+
+    The purpose of this adjustment is to handle class imbalance.
+
+    Parameters
+    ----------
+    labels : np.array
+      A discrete vector of noisy labels, i.e. some labels may be erroneous.
+      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
+
+    pred_probs : np.array (shape (N, K))
+      P(label=k|x) is a matrix with K model-predicted probabilities.
+      Each row of this matrix corresponds to an example x and contains the model-predicted
+      probabilities that x belongs to each possible class.
+      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
+      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
+
+    Returns
+    -------
+    pred_probs_adj : np.array (float)
+    """
+
+    # Get expected (average) self-confidence for each class
+    confident_thresholds = get_confident_thresholds(labels, pred_probs)
+
+    # Subtract the class confident thresholds
+    pred_probs_adj = pred_probs - confident_thresholds
+
+    # Renormalize by shifting data to take care of negative values from the subtraction
+    pred_probs_adj += 1
+    pred_probs_adj /= pred_probs_adj.sum(axis=1)[:, None]
+
+    return pred_probs_adj
+
+
+def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array:
+    """Returns expected (average) "self-confidence" for each class.
+
+    The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
+
+    Parameters
+    ----------
+    labels : np.array
+      A discrete vector of noisy labels, i.e. some labels may be erroneous.
+      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
+
+    pred_probs : np.array (shape (N, K))
+      P(label=k|x) is a matrix with K model-predicted probabilities.
+      Each row of this matrix corresponds to an example x and contains the model-predicted
+      probabilities that x belongs to each possible class.
+      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
+      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
+
+    Returns
+    -------
+    confident_thresholds : np.array (shape (K,))
+
+    """
+    confident_thresholds = np.array(
+        [np.mean(pred_probs[:, k][labels == k]) for k in range(pred_probs.shape[1])]
+    )
+    return confident_thresholds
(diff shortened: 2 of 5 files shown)


tests/test_rank.py
```diff
@@ -160,3 +160,105 @@ def test_order_label_issues_using_self_confidence_ranking():
...
+    )
+    scores = rank.get_confidence_weighted_entropy_for_each_label(data["labels"], data["pred_probs"])
+    indices = np.arange(len(scores))[data["label_errors_mask"]]
+    scores = scores[data["label_errors_mask"]]
+    score_idx = sorted(list(zip(scores, indices)), key=lambda y: y[0])  # sort indices by score
+    label_issues_indices2 = [z[1] for z in score_idx]
+    assert all(label_issues_indices == label_issues_indices2)
+
+
+def test_order_label_issues_using_scoring_func_ranking():
```
jwmueller, line 180: It seems to me we can easily include the above 2 tests inside of this function as well:
`test_order_label_issues_using_confidence_weighted_entropy_ranking()`, `test_order_label_issues_using_self_confidence_ranking()`.   What I mean is inside this function, you can create a dict object:
```
scoring_methods_dict = {"self_confidence": rank.get_self_confidence_for_each_label, "normalized_margin":  rank.get_normalized_margin_for_each_label, "confidence_weighted_entropy": rank.get_confidence_weighted_entropy_for_each_label}

scoring_methods = scoring_methods_dict.keys()
```
and then add something like this after L197s `assert all(label_issues_indices == label_issues_indices2)
`:

```
score_func = scoring_methods_dict[]
scores = score_func(data["labels"], data["pred_probs"])
 indices = np.arange(len(scores))[data["label_errors_mask"]]  # actually only needs to be called once outside the for loop.
    scores = scores[data["label_errors_mask"]]
    score_idx = sorted(list(zip(scores, indices)), key=lambda y: y[0])  # sort indices by score
    label_issues_indices3 = [z[1] for z in score_idx]
    assert all(label_issues_indices == label_issues_indices3)
```

I prefer doing it that way to reduce the amount of duplicated copy/paste code, and it will be easier to add tests for more scoring functions.

tests/test_rank.py
```diff
@@ -160,3 +160,105 @@ def test_order_label_issues_using_self_confidence_ranking():
...
+            labels=data["labels"],
+            pred_probs=data["pred_probs"],
+            rank_by=method,
+        )
+        scores = rank.score_label_quality(data["labels"], data["pred_probs"], method=method)
+        indices = np.arange(len(scores))[data["label_errors_mask"]]
+        scores = scores[data["label_errors_mask"]]
+        score_idx = sorted(list(zip(scores, indices)), key=lambda y: y[0])  # sort indices by score
+        label_issues_indices2 = [z[1] for z in score_idx]
+        assert all(label_issues_indices == label_issues_indices2)
```
jwmueller, line 197: when this fails, would be helpful to print which `scoring_method` it failed on.

tests/test_rank.py
```diff
@@ -160,3 +160,105 @@ def test_order_label_issues_using_self_confidence_ranking():
...
+
+
+def test_order_label_issues_using_scoring_func_adj_pred_probs_ranking():
```
jwmueller, line 200: To me, this seems like similar test function as: `test_order_label_issues_using_confidence_weighted_entropy_ranking()` except just need to pass in special `rank_by_kwargs`.  Could we reduce copy/pasted code, by defining one helper function which can optionally take in `rank_by_kwargs` and pass it where it's needed, and then just have both this test and `test_order_label_issues_using_confidence_weighted_entropy_ranking()` call the helper function with different args?  Or is it better to keep these two tests as entirely different functions without any shared helper method?

tests/test_rank.py
```diff
@@ -160,3 +160,105 @@ def test_order_label_issues_using_self_confidence_ranking():
...
+            },  # adjust predicted probabilities by subtracting class thresholds (as defined in confident learning paper)
+        )
+        scores = rank.score_label_quality(
+            data["labels"], data["pred_probs"], method=method, adj_pred_probs=True
+        )
+        indices = np.arange(len(scores))[data["label_errors_mask"]]
+        scores = scores[data["label_errors_mask"]]
+        score_idx = sorted(list(zip(scores, indices)), key=lambda y: y[0])  # sort indices by score
+        label_issues_indices2 = [z[1] for z in score_idx]
+        assert all(label_issues_indices == label_issues_indices2)
```
jwmueller, line 222: when this fails, would be helpful to print which scoring_method it failed on.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+
+    """
+
+    num_classes = pred_probs.shape[1]
+
+    # Note that dividing by log(num_classes) changes the base of the log which rescales entropy to 0-1 range
+    return -np.sum(pred_probs * np.log(pred_probs), axis=1) / np.log(num_classes)
+
+
+def subtract_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array:
```
jwmueller, line 254: I think this is more of a helper method and not intended to be user-facing (will clutter the documentation IMO). 
So I'd prefer moving this method and other helper functions to a new module named something like `cleanlab/utils/label_quality_utils.py`
whose header says: `"""Helper functions for computing label quality scores"""`

That is unless you think beginner users would also find this `subtract_confident_thresholds()` method useful and it should be prominently displayed in the top-level documentation?

cleanlab/rank.py
```diff
@@ -111,19 +123,19 @@ def get_self_confidence_for_each_label(
...
 
 
 def get_normalized_margin_for_each_label(
-    labels,
-    pred_probs,
-):
+    labels: np.array,
+    pred_probs: np.array,
+) -> np.array:
     """Returns the "normalized margin" for every example associated pred_probs and
```
jwmueller, line 139: Can we make it clear: This is a function to compute label-quality scores, where lower scores indicate labels less likely to be correct.

In general, the docstring for every label-quality scoring function should make it clear that this is what it is.

cleanlab/rank.py
```diff
@@ -57,36 +61,45 @@ def order_label_issues(
...
 
 
 def get_self_confidence_for_each_label(
-    labels,
-    pred_probs,
-):
+    labels: np.array,
+    pred_probs: np.array,
+) -> np.array:
     """Returns the "self-confidence" for every example in the dataset associated pred_probs
```
jwmueller, line 103: See comment below about making it clear: This is a function to compute label-quality scores, ...

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
     )
-    label_quality_scores = (self_confidence - max_prob_not_label + 1) / 2
+    normalized_margin = (self_confidence - max_prob_not_label + 1) / 2
+    return normalized_margin
+
+
+def get_confidence_weighted_entropy_for_each_label(
+    labels: np.array, pred_probs: np.array
+) -> np.array:
+    """Returns the normalized entropy divided by "self-confidence".
```
jwmueller, line 183: See comment above about making it clear: This is a function to compute label-quality scores, ...

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+
+    Score is between 0 and 1.
+    1 - clean label (not an error).
+    0 - dirty label (label error).
```
jwmueller, line 187: This should live in the `score_label_quality()` docstring as behavior that applies to all scoring functions. So it should not be in the docstring for individual scoring functions.  In the individual scoring functions' docstrings, I'd just put something simpler like:

"This is a function to compute label-quality scores, where lower scores indicate labels less likely to be correct."

jwmueller: I'd also change the language to:
```
 1 - clean label (given label is likely correct).
 0 - dirty label (given label is likely incorrect).
```

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+    pred_probs_adj = pred_probs - confident_thresholds
+
+    # Renormalize by shifting data to take care of negative values from the subtraction
+    pred_probs_adj += 1
+    pred_probs_adj /= pred_probs_adj.sum(axis=1)[:, None]
+
+    return pred_probs_adj
+
+
+def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array:
```
jwmueller, line 295: Maybe this is also helper function that belongs in something like: `cleanlab/utils/label_quality_utils.py`?

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+    confident_thresholds : np.array (shape (K,))
+
+    """
+    confident_thresholds = np.array(
+        [np.mean(pred_probs[:, k][labels == k]) for k in range(pred_probs.shape[1])]
+    )
+    return confident_thresholds
+
+
+def score_label_quality(
```
jwmueller, line 327: [Nit] I'd prefer to have this function appear above all the individual scoring functions in the file. That'll make it more clear that each scoring function is a special instance of this function.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+    labels: np.array,
+    pred_probs: np.array,
+    method: str = "self_confidence",
+    adj_pred_probs: bool = False,
+) -> np.array:
+    """Returns the label quality scores.
```
jwmueller, line 333: "Returns the label quality scores."
->
"Returns label quality scores for each datapoint".

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+
+    The scoring methods are heuristics to enable users to quickly rank order data to find potential label quality issues.
```
jwmueller, line 335: would delete this sentence (especially 'heuristics' does not sound good). And replace with explanation above of what higher-vs-lower scores mean and intended range of scores.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+      A discrete vector of noisy labels, i.e. some labels may be erroneous.
+      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
+
+    pred_probs : np.array (shape (N, K))
+      P(label=k|x) is a matrix with K model-predicted probabilities.
+      Each row of this matrix corresponds to an example x and contains the model-predicted
+      probabilities that x belongs to each possible class.
+      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
+      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
+
```
jwmueller, line 201: We probably don't want to copy/paste argument definitions for each individual scoring function, as then updating them will be error-prone. Instead, can we just say:
```
pred_probs : np.array (shape (N, K))
Predicted-probabilities in the same format expected by the `score_label_quality()` method.
```
We'd want to use this same brief explanation in docstring of each the individual scoring functions. Then `score_label_quality()` can host the 'master' documentation for the `pred_probs` arg with all the gory details in one place.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+    Score is between 0 and 1.
+    1 - clean label (not an error).
+    0 - dirty label (label error).
+
+    Parameters
+    ----------
+    labels : np.array
+      A discrete vector of noisy labels, i.e. some labels may be erroneous.
+      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
```
jwmueller, line 193: Same comment as the one below for `pred_probs()`.
Avoid copy/pasting details about the format of the argument across the individual scoring functions and instead have one place the user can refer to details, which should be docstring of `score_label_quality()`.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+    pred_probs : np.array (shape (N, K))
+      P(label=k|x) is a matrix with K model-predicted probabilities.
+      Each row of this matrix corresponds to an example x and contains the model-predicted
+      probabilities that x belongs to each possible class.
+      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
+      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
+
+    Returns
+    -------
+    confidence_weighted_entropy : np.array (float)
```
jwmueller, line 204: For all the individual label quality scoring functions, I think return arg should be named: 
`label_quality_scores`

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+      Return a score (between 0 and 1) for each example of its likelihood of
+      being correctly labeled. Assumes pred_probs is computed holdout/out-of-sample.
```
jwmueller, line 206: seems irrelevant to the return type:
"Assumes pred_probs is computed holdout/out-of-sample."

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+
+    """
+
+    # Get confidence of given label
```
jwmueller, line 210: [Nit] comment seems unnecessary given the very clear code

cleanlab/rank.py
```diff
@@ -28,16 +28,20 @@
...
-):
-    """Sorts label issues by normalized margin.
+    rank_by: str = "normalized_margin",
+    rank_by_kwargs: dict = {},
+) -> np.array:
+    """Sorts label issues by label quality score.
+
+    Default label quality score is "normalized margin".
     See https://arxiv.org/pdf/1810.05369.pdf (eqn 2.2)
     e.g. normalized_margin = prob_label - max_prob_not_label
```
jwmueller, line 46: Let's delete: "e.g. normalized_margin = prob_label - max_prob_not_label" and try to reduce redundancy between the docstrings now that there are many more of them.

cleanlab/rank.py
```diff
@@ -57,36 +61,45 @@ def order_label_issues(
...
+          https://arxiv.org/abs/1911.00068
 
     Returns
     -------
-      label_issues_idx : np.array (int)
-        Return the index integers of the label issues, ordered by
-        the normalized margin."""
+    label_issues_idx : np.array (int)
+      Return the index integers of the label issues, ordered by
+      the normalized margin."""
```
jwmueller, line 83: "ordered by
      the normalized margin" is wrong, since rank_by can be specified.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+      .. seealso::
+        :func:`self_confidence`
+        :func:`normalized_margin`
+        :func:`confidence_weighted_entropy`
+
+    adj_pred_probs : bool, default = True
+      Adjust predicted probabilities by subtracting the class confident thresholds and renormalizing.
+      The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
+      See paper "Confident Learning: Estimating Uncertainty in Dataset Labels" by Northcutt et al.
+      https://arxiv.org/abs/1911.00068
```
jwmueller, line 362: This docstring can probably be shortened to:
"Account for class-imbalance in the label-quality scoring  (by adjusting predicted probabilities via subtraction of class confident thresholds and renormalization). Set this `= False` if you prefer to emphasize label issues in classes that are more common throughout the dataset.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
```
jwmueller, line 446: This is duplicated in far too many docstrings IMO. Should be just mentioned once in the score_label_quality docstring.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+    return label_quality_scores
+
+
+def score_label_quality_ensemble(
+    labels: np.array,
+    pred_probs_list: List[np.array],
+    method: str = "self_confidence",
+    adj_pred_probs: bool = False,
+) -> np.array:
+    """Returns the ensemble label quality scores.
```
jwmueller, line 414: "Returns the ensemble label quality scores"
-> 
"Returns label quality scores based on predictions from an ensemble of models"

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+
+    This requires a list of pred_probs from each model in the ensemble.
+
+    The scoring methods are heuristics to enable users to quickly rank order data to find potential label quality issues.
```
jwmueller, line 418: delete this sentence

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+
+    Parameters
+    ----------
+    labels : np.array
+      A discrete vector of noisy labels, i.e. some labels may be erroneous.
+      *Format requirements*: for dataset with K classes, labels must be in {0,1,...,K-1}.
```
jwmueller, line 424: again tell user to refer to `score_label_quality()` doc, where specify the precise formatting requirement details in one place only.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+
+    pred_probs_list : List of np.array (shape (N, K))
+      List of pred_probs from each model in the ensemble.
+      P(label=k|x) is a matrix with K model-predicted probabilities.
+      Each row of this matrix corresponds to an example x and contains the model-predicted
+      probabilities that x belongs to each possible class.
+      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
+      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
```
jwmueller, line 432: Specify that each element in this list should be array of pred_probs in same format as for `score_label_quality()` (don't mention details of the formatting to reduce duplication across docstrings).

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+    method : {"self_confidence", "normalized_margin", "confidence_weighted_entropy"}, default="self_confidence"
+      Label quality scoring method. Default is "self_confidence".
+
+      .. seealso::
+        :func:`self_confidence`
+        :func:`normalized_margin`
+        :func:`confidence_weighted_entropy`
+
+    adj_pred_probs : bool, default = True
+      Adjust predicted probabilities by subtracting the class confident thresholds and renormalizing.
```
jwmueller, line 443: See above comment on this arg description. Also try to reduce redundancy of this description with its description in `score_label_quality()` docstring.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+            pred_probs=pred_probs,
+            method=method,
+            adj_pred_probs=adj_pred_probs,
+        )
+        accuracy = (pred_probs.argmax(axis=1) == labels).mean()
+        scores_list.append(scores)
+        accuracy_list.append(accuracy)
+
+    # Weight by accuracy
+    weights = np.array(accuracy_list) / sum(accuracy_list)
```
jwmueller, line 478: can we make this a kwarg to the method?
Eg. 
`weight_ensemble_members_by = "accuracy"`
where another option could be "uniform" to apply uniform weights.
Could add TODO that we eventually want to support other metrics like "balanced_accuracy", "logloss", "calibration", etc.

Could add TODO that we eventually want to support "custom" option where user provides np.array or list of weights that they selected themselves (eg. if they already know how trustworthy each model is).

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+    for pred_probs in pred_probs_list:
+
+        # Calculate scores and accuracy
+        scores = score_label_quality(
+            labels=labels,
+            pred_probs=pred_probs,
+            method=method,
+            adj_pred_probs=adj_pred_probs,
+        )
+        accuracy = (pred_probs.argmax(axis=1) == labels).mean()
```
jwmueller, line 473: only compute this if the user has specified to weight by accuracy (see below comment).

cleanlab/rank.py
jwmueller: If the user specified to weight by accuracy, could be nice to print out the accuracy of each ensemble member and that how they will be weighted.

Consider introducing a `verbosity` kwarg that enables/disables print statements.

jwmueller (changes requested):
Code looks mostly good but left lots of documentation/style comments. Main thing is to try and reduce duplication (copy/paste) in code, tests, and docstrings.  With too much duplication, future updates will be cumbersome and error-prone.

tests/test_rank.py
cgnorthcutt: The lack of equivalency is the failure in this case. Any difference is failure for this test. Its unspecified an exactness for what the indices should be, only their equivalency is expected.

tests/test_rank.py
cgnorthcutt: @JohnsonKuan one way to do this is to use @paramterize and pass parameters to create separate tests for every method without re-writing code. Example of how to do that here: https://github.com/cleanlab/cleanlab/blob/master/tests/test_filter_count.py#L332

tests/test_rank.py
cgnorthcutt: The way to do this @JohnsonKuan  is not use a for-loop. This should be rewritten using @parameterize. See https://github.com/cleanlab/cleanlab/blob/master/tests/test_filter_count.py#L332 for an example.

JohnsonKuan: @cgnorthcutt Nice! That is a convenient decorator. Thanks for sharing. I rewrote the tests to use @parameterize

@jwmueller Agree - added messages to asserts so tests will print the scoring method if assertions fails

315e956b18f489a90b90a2ca62016b34176997fc
b56d2d8d969a1e8e8f91fd29eca706b2932edeb9

tests/test_rank.py
JohnsonKuan: Agree, good catch. I combined the tests and also replaced for-loops with @pytest.mark.parametrize. I believe this should take care of the redundancies and make it easier to add tests for more scoring functions in the future.

67612f40b636d2794bfe4b9c3674b19ce4087501
315e956b18f489a90b90a2ca62016b34176997fc
2c83638a255f79b4fea3f6454894bff55c0399b0

tests/test_rank.py
JohnsonKuan: Agree, good catch. I removed this test function. This is now taken care of in the other test function, `test_order_label_issues_using_scoring_func_ranking`, where I added @parameterize decorator to pass the rank_by arg.

Thanks @cgnorthcutt for the tip on @parameterize

315e956b18f489a90b90a2ca62016b34176997fc

tests/test_rank.py
JohnsonKuan: Agree, good call. I committed the change to print the scoring method when assertion fails.

cleanlab/filter.py
```diff
@@ -339,6 +341,15 @@ def find_label_issues(
...
       If 0, no print statements. If 1, prints when multiprocessing happens.
 
+    rank_by_kwargs : dict
+      Optional keyword arguments to pass into scoring functions for ranking.
+      Accepted args include:
+        adj_pred_probs : bool, default = False
+          Adjust predicted probabilities by subtracting the class confident thresholds and renormalizing.
+          The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
+          See paper "Confident Learning: Estimating Uncertainty in Dataset Labels" by Northcutt et al.
+          https://arxiv.org/abs/1911.00068
```
jwmueller, line 351: would specify exactly which method these kwargs will be passed into. Then refer user to that method's documentation rather than trying to explain all the kwargs here. This reduces redundancy in the docstrings.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+
+    adj_pred_probs : bool, default = True
+      Adjust predicted probabilities by subtracting the class confident thresholds and renormalizing.
+      The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
+      See paper "Confident Learning: Estimating Uncertainty in Dataset Labels" by Northcutt et al.
+      https://arxiv.org/abs/1911.00068
+
+    Returns
+    -------
+    label_quality_scores : np.array (float)
```
jwmueller, line 366: IMO this is one place some redundancy would be good. I'd again explain here what the scores should look like and what higher values mean

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+    normalized_margin
+    confidence_weighted_entropy
+    score_label_quality
+
+    """
+
+    # Generate scores for each model's pred_probs
+    scores_list = []
+    accuracy_list = []
+    for pred_probs in pred_probs_list:
```
jwmueller, line 464: A common user error will be `pred_probs_list` is np.array. Can we add explicit error check for this?
Also consider error if `pred_probs_list` is list but len = 0, and warning if its len = 1.

cleanlab/rank.py
JohnsonKuan: Thx good call. Done: 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: Done: 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: Done: 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: Agree, sounds good. Done: 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: Good call, I agree. Done: 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: Done: 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: Done: 100783d9199110a66737f2d09e2fce762b500c45

Thanks

cleanlab/rank.py
JohnsonKuan: Agree, done: 100783d9199110a66737f2d09e2fce762b500c45

Thanks

cleanlab/rank.py
JohnsonKuan: Done: 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: I'm cool with that. I don't have a strong preference on this one.

For now, I made the changes: 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: Done: 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: Agree, removed comment

cleanlab/rank.py
JohnsonKuan: Done: 100783d9199110a66737f2d09e2fce762b500c45

Thanks

cleanlab/rank.py
JohnsonKuan: Good catch. Fixed 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: That sounds good. Changed docstring to this shortened version.

100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: Agree. Removed it everywhere except in score_label_quality.

100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
JohnsonKuan: Done, thanks

cleanlab/rank.py
JohnsonKuan: Done, thanks

cleanlab/rank.py
JohnsonKuan: Done, thanks

cleanlab/rank.py
JohnsonKuan: Done, thanks

cleanlab/rank.py
JohnsonKuan: Done, thanks

cleanlab/filter.py
JohnsonKuan: Agree. Done, thanks

cleanlab/rank.py
JohnsonKuan: Agree, done: 100783d9199110a66737f2d09e2fce762b500c45

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+      .. seealso::
+        :func:`self_confidence`
+        :func:`normalized_margin`
+        :func:`confidence_weighted_entropy`
+
+    adj_pred_probs : bool, default = True
+      Account for class-imbalance in the label-quality scoring (by adjusting predicted probabilities
+      via subtraction of class confident thresholds and renormalization).
+      Set this = False if you prefer to emphasize label issues in classes that are more common
+      throughout the dataset
```
jwmueller, line 138: Nit: punctuation

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
 
+    method : {"self_confidence", "normalized_margin", "confidence_weighted_entropy"}, default="self_confidence"
+      Label quality scoring method. Default is "self_confidence".
+
+      .. seealso::
+        :func:`self_confidence`
+        :func:`normalized_margin`
+        :func:`confidence_weighted_entropy`
+
+    adj_pred_probs : bool, default = True
```
jwmueller, line 134: The default is `False` currently. If False is kept as default, then docstring should explain "Set this = True if you prefer..." rather than "Set this = False if ..."

cleanlab/rank.py
```diff
@@ -57,47 +62,56 @@ def order_label_issues(
...
-    its given class label.
+    # Get label quality scores for label issues
+    label_quality_scores_issues = label_quality_scores[label_issues_mask]
+
+    return label_issues_idx[np.argsort(label_quality_scores_issues)]
+
+
+def score_label_quality(
+    labels: np.array,
+    pred_probs: np.array,
```
jwmueller, line 100: Can we update the signature of this func to be:
```
def score_label_quality(labels: np.array,
    pred_probs: np.array, *, method: str = "self_confidence",
    adj_pred_probs: bool = False,
)
```
with the `, *,` separating the keyword args from the positional args. That'll give us more flexibility for api changes going forward without breaking users' existing code.

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+    # Pass keyword arguments for scoring function
+    input = {"labels": labels, "pred_probs": pred_probs}
+
+    # Calculate scores
+    label_quality_scores = scoring_func(**input)
+
+    return label_quality_scores
+
+
+def score_label_quality_ensemble(
```
jwmueller, line 187: can we similarly update this function's signature as requested for the `score_label_quality()` signature update in my comment above?

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+    0 - dirty label (given label is likely incorrect).
+
+    Parameters
+    ----------
+    labels : np.array
+      Labels in the same format expected by the `score_label_quality()` method.
+
+    pred_probs_list : List of np.array (shape (N, K))
+      Each element in this list should be an array of pred_probs in the same format
+      expected by the `score_label_quality()` method
```
jwmueller, line 216: Add extra sentence: 

Each element of `pred_probs_list` corresponds to the predictions from one model for all datapoints.

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+    """Returns label quality scores based on predictions from an ensemble of models.
+
+    This is a function to compute label-quality scores for classification datasets,
+    where lower scores indicate labels less likely to be correct.
+
+    Ensemble scoring requires a list of pred_probs from each model in the ensemble.
+
+    For each pred_probs in list, compute label quality score.
+    Take the average of the scores with the chosen weighting scheme determined by weight_ensemble_members_by.
+    Default weight_ensemble_members_by is "uniform" which computes the simple average of scores.
```
jwmueller, line 203: I wouldn't mention defaults up here but rather where the arg is explained under "Parameters" (easy to later forget to change this docstring if it's up here).

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+
+      .. seealso::
+        :func:`self_confidence`
+        :func:`normalized_margin`
+        :func:`confidence_weighted_entropy`
+
+    adj_pred_probs : bool, default = True
+      Adj_pred_probs in the same format expected by the `score_label_quality()` method.
+
+    weight_ensemble_members_by : {"uniform", "accuracy"}, default="uniform"
```
jwmueller, line 229: why not default to "accuracy" if it seemed better in benchmark? It also intuitively makes more sense to me.

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
-        Return the holdout probability that each example in pred_probs belongs to its
-        label. Assumes pred_probs is computed holdout/out-of-sample."""
+    label_quality_scores : np.array (float)
+
+    See Also
+    --------
+    self_confidence
+    normalized_margin
+    confidence_weighted_entropy
+    score_label_quality
```
jwmueller, line 246: I think See Also should only be `score_label_quality`. The other ones are all explained/linked in the `score_label_quality` documentation

cleanlab/rank.py
jwmueller: update LGTM!

cleanlab/rank.py
```diff
@@ -28,18 +28,23 @@
...
+from cleanlab.utils.label_quality_utils import subtract_confident_thresholds
 
 
 def order_label_issues(
-    label_issues_mask,
-    labels,
-    pred_probs,
+    label_issues_mask: np.array,
+    labels: np.array,
+    pred_probs: np.array,
```
jwmueller, line 39: Not your responsibility, but since you already made the change for the other functions, could you also update docstring for these args here (in `order_label_issues` docstring) to what you did for the other methods:
```
labels : np.array
      Labels in the same format expected by the `score_label_quality()` method.
    pred_probs : np.array (shape (N, K))
      Predicted-probabilities in the same format expected by the `score_label_quality()` method.
```

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+    return label_quality_scores
+
+
+def score_label_quality_ensemble(
+    labels: np.array,
+    pred_probs_list: List[np.array],
+    method: str = "self_confidence",
+    adj_pred_probs: bool = False,
+    weight_ensemble_members_by: str = "uniform",
+    verbose: int = 0,
```
jwmueller, line 192: verbose should typically be 1 by default, otherwise most users will never see the informative print statements

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+
+    adj_pred_probs : bool, default = True
+      Adj_pred_probs in the same format expected by the `score_label_quality()` method.
+
+    weight_ensemble_members_by : {"uniform", "accuracy"}, default="uniform"
+      Weighting scheme used to aggregate scores from each model:
+        "uniform": Take the simple average of scores
+        "accuracy": Take weighted average of scores, weighted by model accuracy
+
+    verbose : int
```
jwmueller, line 234: specify default value

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+      If 0, no print statements. If 1, print weighting scheme and accuracy of each model.
```
jwmueller, line 235: If you change default to 1, I'd just say:
"Set this = 0 to suppress all print statements"
No need to say what will be printed, since that is likely to change over time.

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+    # Aggregate scores with chosen weighting scheme
+    if weight_ensemble_members_by == "uniform":
+
+        # Uniform weights (simple average)
+        label_quality_scores = scores_ensemble.mean(axis=1)
+
+    elif weight_ensemble_members_by == "accuracy":
+
+        # Weight by accuracy
+        weights = np.array(accuracy_list) / sum(accuracy_list)
```
jwmueller, line 305: Let's also print out the aggregation weights that will be used when `verbosity > 0`

cleanlab/rank.py
jwmueller: nit: don't need to compute this except if weighting scheme = "accuracy"

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+
+
+def get_entropy(pred_probs: np.array) -> np.array:
+    """Returns the normalized entropy of pred_probs.
+
+    Read more about normalized entropy here: https://en.wikipedia.org/wiki/Entropy_(information_theory)
+
+    Normalized entropy is used in active learning for uncertainty sampling: https://towardsdatascience.com/uncertainty-sampling-cheatsheet-ec57bc067c0b
+
+    Normalized entropy is between 0 and 1. Higher values of entropy indicate higher uncertainty in the model's prediction of the correct label.
```
jwmueller, line 439: Can we make this the 2nd sentence of the docstring before the external links are stated. I'd also add one more sentence: 

"Unlike label-quality scores, entropy only depends on the model's predictions, not the given label".

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
+    # Divide entropy by self confidence
+    label_quality_scores = get_entropy(**{"pred_probs": pred_probs}) / self_confidence
+
+    # Rescale
+    label_quality_scores = np.log(label_quality_scores + 1) / label_quality_scores
+
+    return label_quality_scores
+
+
+def get_entropy(pred_probs: np.array) -> np.array:
```
jwmueller, line 432: For this function's return value, we currently have: higher values = more concerning, whereas for the label-quality scores, we have lower values = more concerning.  I think we need to standardize everything around lower = more concerning, so could change this score to `1 - normalized_entropy`.

What do you think?

JohnsonKuan: I think we should leave it as-is because entropy here is not used as a label quality score.

If we add it to the label quality scoring function in the future, we can rename it to "inv_entropy" or something similar and do the 1 - normalized_entropy.

cleanlab/rank.py
```diff
@@ -109,57 +123,335 @@ def get_self_confidence_for_each_label(
...
-    labels.
+    labels: np.array,
+    pred_probs: np.array,
+) -> np.array:
+    """Returns the "normalized margin" label-quality score for each datapoint.
+
+    This is a function to compute label-quality scores for classification datasets,
+    where lower scores indicate labels less likely to be correct.
+
     The normalized margin is (p(label = k) - max(p(label != k))), i.e. the probability
```
jwmueller, line 363: Not your responsibility, but can we add:
"Letting `k` denote the given label for a datapoint, the normalized margin is ..."

cleanlab/rank.py
jwmueller: I guess because this function is currently merely intended as a helper function for `get_confidence_weighted_entropy_for_each_label`?

If so, maybe we move it to `utils/` as well?
I think for all user-facing scoring functions, we want to be unified and always have: lower = more concerning. 

So later when we introduce a scoring function based on this, we could add the `1 - entropy()` in that function and have it import this as a helper from `utils/` as well.

JohnsonKuan: Right now it's more of a helper function for confidence_weighted_entropy. We can consider putting it in label_quality_utils.py

JohnsonKuan: Just saw your previous comment. I think we commented at the same time :)

Agree, will move it to utils and we can add 1 - entropy() later if we introduce this as a scoring function

cleanlab/rank.py
```diff
@@ -57,47 +62,56 @@ def order_label_issues(
...
+    # Get label quality scores for label issues
+    label_quality_scores_issues = label_quality_scores[label_issues_mask]
+
+    return label_issues_idx[np.argsort(label_quality_scores_issues)]
+
+
+def score_label_quality(
+    labels: np.array,
+    pred_probs: np.array,
+    method: str = "self_confidence",
```
jwmueller, line 101: Why does default here not match default for `order_label_issues`?  Shouldn't we be consistent throughout?

@cgnorthcutt

tests/test_rank.py
jwmueller: Given the switch to `parameterize` decorator based testing, do we still need the singleton tests:
`test_get_normalized_margin_for_each_label()`
`test_get_self_confidence_for_each_label()` ?

I suspect no, so let's delete these.  But if yes, then why is there no such test for the new scoring funcs?

jwmueller: @cgnorthcutt Will let you review the `parameterize` decorator based tests.

cleanlab/rank.py
JohnsonKuan: Done, thanks.
3dbbe7054aeae25b8bddd73a53838cce35b06fa9

cleanlab/rank.py
JohnsonKuan: Done, thanks
3dbbe7054aeae25b8bddd73a53838cce35b06fa9

cleanlab/rank.py
JohnsonKuan: Done, thanks
3dbbe7054aeae25b8bddd73a53838cce35b06fa9

cleanlab/rank.py
JohnsonKuan: Done 3dbbe7054aeae25b8bddd73a53838cce35b06fa9

cleanlab/rank.py
JohnsonKuan: Done 3dbbe7054aeae25b8bddd73a53838cce35b06fa9

cleanlab/rank.py
JohnsonKuan: Done 3dbbe7054aeae25b8bddd73a53838cce35b06fa9

cleanlab/rank.py
JohnsonKuan: Good call. Done, thanks 
3dbbe7054aeae25b8bddd73a53838cce35b06fa9

cleanlab/rank.py
JohnsonKuan: Yeah definitely. Done, thanks 
3dbbe7054aeae25b8bddd73a53838cce35b06fa9

cleanlab/rank.py
JohnsonKuan: Agree, removed the others so now it's only `score_label_quality`
3dbbe7054aeae25b8bddd73a53838cce35b06fa9

cleanlab/rank.py
JohnsonKuan: That's true. Initially, I set it to "uniform" prior to latest benchmark results. Now that we have the latest benchmark results where it outperforms, I changed the default now to "accuracy"

Thanks

3dbbe7054aeae25b8bddd73a53838cce35b06fa9

cleanlab/rank.py
JohnsonKuan: Agree, I removed this sentence

cleanlab/rank.py
JohnsonKuan: Done, thanks

cleanlab/rank.py
JohnsonKuan: Done, thanks

cleanlab/rank.py
JohnsonKuan: Done, thanks

cleanlab/rank.py
JohnsonKuan: Good catch. I changed the default to True.

cleanlab/rank.py
JohnsonKuan: Good catch. I changed the logic so it now only computes if user specifies to weight by "accuracy"

cleanlab/rank.py
jwmueller: Once we decide on default scoring method for the functions, we'll also want to make sure `adj_pred_probs` default matches across:  
 `order_label_issues()`, `score_label_quality()`, `score_label_quality_ensemble()`

JohnsonKuan: Confirming that I changed all defaults to:
`method="normalized_margin"`
`adj_pred_probs=False`

eb9c57a04b33412f6a547860e816d21b9d6c05ca

cleanlab/rank.py
JohnsonKuan: Per discussion on slack, changed default to `normalized_margin` everywhere.

cleanlab/rank.py
```diff
@@ -28,76 +28,86 @@
...
-      probabilities that x belongs to each possible class.
-      The columns must be ordered such that these probabilities correspond to class 0,1,2,...
-      `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
+      Predicted-probabilities in the same format expected by the `score_label_quality()` method.
 
-    rank_by : str ['normalized_margin', 'self_confidence']
+    rank_by : str ['normalized_margin', 'self_confidence', 'confidence_weighted_entropy']
       Method to order label error indices (instead of a bool mask), either:
         'normalized_margin' := normalized margin (p(label = k) - max(p(label != k)))
         'self_confidence' := [pred_probs[i][labels[i]] for i in label_issues_idx]
```
jwmueller, line 63: [not your responsibility but could be quickly addressed before merging this PR] This can be more clearly written as:
```
Score by which to order label error indices (in increasing order), either: 'normalized_margin', 'self_confidence', or 'confidence_weighted_entropy'. See `score_label_quality()` documentation for description of these scores.

Comment (jwmueller) [reply cleanlab/rank.py:63]: And then I'd move the mathematical defn of the scores to the `score_label_quality()` docstring for its `method` arg. I'd also rewrite it a bit more clearly as the following:
```
Letting `k := labels[i]` and `P := pred_probs[i]`  denote the given label and predicted class-probabilities for the `i`th datapoint, its score can either be: 
'normalized_margin' := P[k] - max_{k' != k}[ P[k'] ] 
'self_confidence' := P[k]
'confidence_weighted_entropy' := entropy(P) / self_confidence
```

**Review by jwmueller — APPROVED** [review]
LGTM besides minor comment on scoring functions docstring, which can be addressed before this is merged without another review

File: cleanlab/rank.py (hunk 50 above)
Comment (JohnsonKuan) [reply cleanlab/rank.py]: Sounds good! I made these changes. Thanks.

35624d6f3e43276c53967e531e123b123f3013a7

File: cleanlab/rank.py [hunk 51]
```diff
@@ -57,109 +118,328 @@ def order_label_issues(
       The columns must be ordered such that these probabilities correspond to class 0,1,2,...
       `pred_probs` should have been computed using 3 (or higher) fold cross-validation.
 
-    rank_by : str ['normalized_margin', 'self_confidence']
-      Method to order label error indices (instead of a bool mask), either:
-        'normalized_margin' := normalized margin (p(label = k) - max(p(label != k)))
-        'self_confidence' := [pred_probs[i][labels[i]] for i in label_issues_idx]
+    method : {"self_confidence", "normalized_margin", "confidence_weighted_entropy"}, default="normalized_margin"
+      Label quality scoring method. Default is "normalized_margin".
```
jwmueller, line 122: [Nit: this is unnecessary] 
Default is "normalized_margin".

cleanlab/rank.py
```diff
@@ -57,109 +118,328 @@ def order_label_issues(
...
+      for the `i`th datapoint, its score can either be:
+      'normalized_margin' := P[k] - max_{k' != k}[ P[k'] ]
+      'self_confidence' := P[k]
+      'confidence_weighted_entropy' := entropy(P) / self_confidence
+
+      Self-confidence works better for finding out-of-distribution (OOD) examples, weird examples, bad examples,
+      multi-label, and other types of label errors.
+
+      Normalized margin works better for finding class conditional label errors where
+      there is another label in the class that is better than the given label.
```
jwmueller, line 134: Language here can be more clear/precise as:

```
Let `C = {0,1,...,K}` denote the classification task's specified set of classes. The normalized_margin score works better for identifying class conditional label errors, i.e. datapoints for which another label in C is appropriate but the given label is not. The self_confidence score  works better for identifying alternative label issues corresponding to bad datapoints that are: not from any of the classes in C, well-described by 2 or more labels in C, or generally just out-of-distribution (ie. anomalous outliers).
```

cleanlab/rank.py
cgnorthcutt: I don't understand why you're computing this. Its in a for-loop but none of the inputs to accuracy change, so its just computing the same thing over and over. So, then what are the weights if all the weights are the same number? I'm confused.

jwmueller:
> While the package is still small, I prefer instead not breaking up the code more than necessary just yet. It makes the pieces harder to follow. But agree with moving to utils if we have several helper methods and growing. Fow now, I'd just change the name of the function to _subtract_confident_thresholds() - thoughts @jwmueller ?

I don't have much preference. Just don't want to commit to these functions' APIs remaining stable and don't want them to clutter the documentation.

cleanlab/rank.py
```diff
@@ -152,14 +164,320 @@ def get_normalized_margin_for_each_label(
...
+        raise ValueError(
+            f"""
+            {method} is not a valid scoring method for rank_by!
+            Please choose a valid rank_by: self_confidence, normalized_margin, confidence_weighted_entropy
+            """
+        )
+
+    # Adjust predicted probabilities
+    if adj_pred_probs:
+        pred_probs = subtract_confident_thresholds(labels, pred_probs)
```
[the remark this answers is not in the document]
JohnsonKuan: My thought for this design choice is to have `adjust_pred_probs` be an option to adjust pred_probs given that there could be multiple ways we can adjust pred_probs. Right now we only have the one option `subtract_confident_thresholds` but we can introduce more adjustment methods later as well.

cleanlab/rank.py
JohnsonKuan: The for-loop goes through each `pred_probs` in `pred_probs_list` (one `pred_probs` for each model in the ensemble).

The inputs to accuracy will change in the for-loop because `pred_probs` changes. The weights will change as the accuracy changes.

cleanlab/rank.py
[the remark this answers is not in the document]
JohnsonKuan: Curious to get @jwmueller thoughts on this

cleanlab/filter.py
```diff
@@ -339,6 +341,15 @@ def find_label_issues(
     verbose : int
       If 0, no print statements. If 1, prints when multiprocessing happens.
 
+    rank_by_kwargs : dict
+      Optional keyword arguments to pass into scoring functions for ranking.
```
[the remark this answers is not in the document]
JohnsonKuan: Good catch, done! Thanks :)

cleanlab/filter.py
```diff
@@ -243,6 +243,7 @@ def find_label_issues(
     num_to_remove_per_class=None,
     n_jobs=None,
     verbose=0,
+    rank_by_kwargs={},
```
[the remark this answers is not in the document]
JohnsonKuan: Good call. I made this change. Thanks!

cleanlab/rank.py
JohnsonKuan: Agree, that sounds better. I made this change. Thanks.

cleanlab/rank.py
JohnsonKuan: Agree, I removed it. Thanks.

cleanlab/filter.py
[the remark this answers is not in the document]
JohnsonKuan: Done, thanks!

cleanlab/utils/label_quality_utils.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+
+    # Get expected (average) self-confidence for each class
+    confident_thresholds = get_confident_thresholds(labels, pred_probs)
+
+    # Subtract the class confident thresholds
+    pred_probs_adj = pred_probs - confident_thresholds
+
+    # Renormalize by shifting data to take care of negative values from the subtraction
+    pred_probs_adj += 1
```
cgnorthcutt, line 79: Is this arbitrary? Why not use `(x-min(x))/(max(x)-min(x))`

cleanlab/utils/label_quality_utils.py
```diff
@@ -0,0 +1,114 @@
...
+    pred_probs_adj /= pred_probs_adj.sum(axis=1)[:, None]
```
cgnorthcutt, line 80: The right hand side is already 1-D, so what is `[:, None]` doing? if `[:, None]` is reshaping the matrix for some reason to make the `/=` operator work, use `.reshape()` instead of clarity.

cleanlab/rank.py
```diff
@@ -28,24 +28,85 @@
...
+    # Calculate label quality scores
+    label_quality_scores = score_label_quality(labels, pred_probs, method=rank_by, **rank_by_kwargs)
+
+    # Get label quality scores for label issues
+    label_quality_scores_issues = label_quality_scores[label_issues_mask]
+
+    return label_issues_idx[np.argsort(label_quality_scores_issues)]
+
+
+def score_label_quality(
```
cgnorthcutt, line 92: I think should be named `get_label_quality_scores()`

cleanlab/utils/label_quality_utils.py
JohnsonKuan: Similar to our decision to not use min-max norm (in [internal doc](https://www.notion.so/cleanlab/Evaluation-of-Label-Quality-Scores-to-Detect-Label-Issues-325f18b39e31466fa7aeb48e4da7441d)) for the scores, I decided not to use min-max norm here as well for adjusting pred_probs because I didn't want the renormalization to depend on characteristics of the input (e.g. min(x)).

cleanlab/utils/label_quality_utils.py
JohnsonKuan: The `[:, None]` adds a dimension, changing the array from 1D shape (n_samples,) to 2D shape (n_samples, 1). This is needed to make the `/=` operator work for broadcasting.

If we want to make adding a dimension explicit, I recommend using: `[:, np.newaxis]`

cleanlab/rank.py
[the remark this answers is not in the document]
jwmueller: I agree with @JohnsonKuan I think it's ok as is, as there could be other ways to do the adjustment in the future (it's also good to use different name for function vs the argname).  My only strong opinion is that  `subtract_confident_thresholds()` should not be a user-facing method, it should either be private method `_subtract_confident_thresholds()` in this file as Curtis suggested, or buried deep inside `utils/` like we currently have it.

cleanlab/rank.py
jwmueller: I'm fine with either way. My only strong opinion is that subtract_confident_thresholds() should not be a user-facing method, it should either be private method `_subtract_confident_thresholds()` in this file as Curtis suggested, or buried deep inside utils/ like we currently have it.

cleanlab/filter.py
```diff
@@ -339,6 +341,15 @@ def find_label_issues(
     verbose : int
       If 0, no print statements. If 1, prints when multiprocessing happens.
 
+    rank_by_kwargs : dict
+      Optional keyword arguments to pass into scoring functions for ranking.
+      Accepted args include:
```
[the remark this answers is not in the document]
JohnsonKuan: I'm new to sphinx and also unfamiliar with how increasingly indented format impacts the docs.

To be safe, I removed the extra indents in this commit: 76d2fff34db9e59e0b114582473e6eab62c39376

cleanlab/rank.py
cgnorthcutt: tl;dr
* change `subtract_confident_thresholds` to `_subtract_confident_thresholds`
* change `adj_pred_probs` to `adjust_pred_probs`
everywhere

cleanlab/rank.py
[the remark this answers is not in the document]
cgnorthcutt: tl;dr:

- change subtract_confident_thresholds to _subtract_confident_thresholds
- change adj_pred_probs to adjust_pred_probs

everywhere

cleanlab/utils/label_quality_utils.py
```diff
@@ -0,0 +1,114 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Helper functions for computing label quality scores"""
+
+import numpy as np
+
+
+def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array:
```
cgnorthcutt, line 22: Imo, this is a core function of the package (to compute the thresholds) and should be user facing. This should be in `count.py`

cleanlab/rank.py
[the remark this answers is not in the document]
jwmueller: could also consider term like `rebalance_` instead of `adjust_` if you don't like latter term.

cleanlab/utils/label_quality_utils.py
cgnorthcutt: Seems worth a comment to clarify what the code is doing then :)

cgnorthcutt:
FInal thoughts:

- I’m not convinced by the +1 numerator way of handling the `adjust_pred_probs` due to the skewing. I’ll let @jwmueller think more on it.
- btw avoid `adj` which could mean adjacent among other things).
- I see benefits of using min-max to normalize rankings. Sure it has downsides (global comparison, robustness to data removed), but I like that it prevents a classifier like random forests from maxing out at `prob = 0.3` and I like that for every dataset, for every type of issue, there is always a `score==0` example (the most susceptible to that issue) and `score==1 `issue (the least susceptible to that issue), with all other scores correctly scaled in between.
- Another option might be `+max(probs)` or `-min(probs)`… then divide by the sum. lots of ways.
- Just keep in mind, no matter what you do, you still divide by the sum and 
- Consider the following when making a decision about how to map scores between 0 and 1:
  - squashing of scores
  - interpretability and ease of understanding what’s the most likely issue and least likely issue within any dataset
  - accuracy of scores
  - interpretability of scores as probabilities
- I'm not convinced of inclusion for the entropy rank  yet — it’s computation seems a little wonky and I question if it can be interpreted probabilistically. I haven’t had time to benchmark it, so please make sure it adds value.
- As a final point, I only recommended using `subtract_confident_thresholds` instead of `adjust_pred_probs` because `adjust_pred_probs` is a boolean. If you do want to be able to swap out any method like this `adjust_pred_probs = "subtract_confident_thresholds"` it might make sense to make it a string to avoid the future API change? This is a minor point and more of a nit. (agree users don’t want to type a string compared to a bool, but if it remains a bool for years to come… then it should have just been  named `subtract_confident_thresholds` and if not, then it might as well be a string now… see what i mean?

Once the above is resolved, LGTM to merge.

cleanlab/rank.py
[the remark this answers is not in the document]
JohnsonKuan: Changed below everywhere. Thanks.

- Changed `subtract_confident_thresholds` to `_subtract_confident_thresholds`
- Changed `adj_pred_probs` to `adjust_pred_probs`

a5afbfb30d8123a795c0c544264244667243c52c

cleanlab/utils/label_quality_utils.py
JohnsonKuan: I added a comment to explain this line. Thanks.

50d6273176922daed8b78f60a514d0c5eab997d2

JohnsonKuan:
> FInal thoughts:
> 
> * I’m not convinced by the +1 numerator way of handling the `adjust_pred_probs` due to the skewing. I’ll let @jwmueller think more on it.
> * btw avoid `adj` which could mean adjacent among other things).
> * I see benefits of using min-max to normalize rankings. Sure it has downsides (global comparison, robustness to data removed), but I like that it prevents a classifier like random forests from maxing out at `prob = 0.3` and I like that for every dataset, for every type of issue, there is always a `score==0` example (the most susceptible to that issue) and `score==1 `issue (the least susceptible to that issue), with all other scores correctly scaled in between.
> * Another option might be `+max(probs)` or `-min(probs)`… then divide by the sum. lots of ways.
> * Just keep in mind, no matter what you do, you still divide by the sum and
> * Consider the following when making a decision about how to map scores between 0 and 1:
>   
>   * squashing of scores
>   * interpretability and ease of understanding what’s the most likely issue and least likely issue within any dataset
>   * accuracy of scores
>   * interpretability of scores as probabilities
> * I'm not convinced of inclusion for the entropy rank  yet — it’s computation seems a little wonky and I question if it can be interpreted probabilistically. I haven’t had time to benchmark it, so please make sure it adds value.
> * As a final point, I only recommended using `subtract_confident_thresholds` instead of `adjust_pred_probs` because `adjust_pred_probs` is a boolean. If you do want to be able to swap out any method like this `adjust_pred_probs = "subtract_confident_thresholds"` it might make sense to make it a string to avoid the future API change? This is a minor point and more of a nit. (agree users don’t want to type a string compared to a bool, but if it remains a bool for years to come… then it should have just been  named `subtract_confident_thresholds` and if not, then it might as well be a string now… see what i mean?
> 
> Once the above is resolved, LGTM to merge.

Thanks @cgnorthcutt for the latest round of comments. I addressed most of the comments with the latest commits:

a5afbfb30d8123a795c0c544264244667243c52c
50d6273176922daed8b78f60a514d0c5eab997d2

The key decision left before merging will be to decide on the renormalization logic for `adjust_pred_probs`. I'll defer to @jwmueller on final decision here.

Regarding inclusion of `confidence_weighted_entropy`: I included this because it performed well in this [internal benchmark](https://github.com/JohnsonKuan/dcai-benchmarks) using the Andrew Ng DCAI competition roman numeral dataset (note: I'm still in the process of cleaning up the repo because it has undergone many changes since we started benchmarking scores). Also, we've seen that entropy based scoring methods performed really well for out-of-distribution (OOD) detection tasks (more to come on this). For context, I've tried many different scoring methods (including variations of the entropy based methods) in the past month and have only included the ones that I believe will add value.

Regarding the last point on using `subtract_confident_thresholds` instead of `adjust_pred_probs`: I see what you mean. I don't have a strong preference here. I'm ok going either way.

Thanks

cgnorthcutt:
Thanks @JohnsonKuan the other commits/changes lgtm!

JohnsonKuan:
@jwmueller 

Per Slack discussion earlier, I added latest commits to:
- Raise ValueError message if `adjust_pred_probs=True` and `method="confidence_weighted_entropy"` because we've decided that `adjust_pred_probs` is not supported for `confidence_weighted_entropy`
- Change renormalization logic for `adjust_pred_probs` to use `pred_probs_adj += confident_thresholds.max()` instead of `pred_probs_adj += 1`

This is ready for your final review. Thanks!

jwmueller:
LGTM

jwmueller (approved):

PR #138, opened 2022-03-25
Change cleanlab version in Image Tutorial

weijinglok:
This PR introduces:

1. Updated Image Tutorial using `cleanlab`'s latest API (AKA the upcoming v2.0.0 API)
2. Additional steps and conditions in the CI workflow (`gh-pages.yaml`) to dynamically install different versions of `cleanlab` in the .ipynb notebooks based on whether it is triggered by a push to `master` branch or a new release.
3. The Colab hyperlink in the tutorial will now link to the executed version of the notebook from the `cleanlab-docs` repo.
3. Fix the `cannot import name 'escape' from 'jinja2'` [issue](https://github.com/pallets/jinja/issues/1626).

### DIFF
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -48,6 +48,33 @@ jobs:
       - name: Install dependencies
         run: python3 -m pip install -r docs/requirements.txt
 
+      - name: Find and replace "%pip install cleanlab" with master branch in .ipynb files
+        if: ${{ github.ref_type == 'branch' }}
+        uses: jacobtomlinson/gha-find-replace@v2
+        with:
+          find: "%pip install cleanlab  ${{ '#' }} for colab"
+          replace: "%pip install git+https://github.com/${{ github.repository }}.git@${{ github.sha }}"
+          include: "docs/source**.ipynb"
+
+      - name: Find and replace "%pip install cleanlab" with release tag in .ipynb files
+        if: ${{ github.ref_type == 'tag' }}
+        uses: jacobtomlinson/gha-find-replace@v2
+        with:
+          find: "%pip install cleanlab  ${{ '#' }} for colab"
+          replace: "%pip install cleanlab==${{ github.ref_name }}"
+          include: "docs/source**.ipynb"
+
+      - name: Add and Commit changes to .ipynb files
+        run: |
+          git config --global user.email "actions@github.com"
+          git config --global user.name "GitHub Actions"
+          git add -A && git commit -m "Change cleanlab ver in .ipynb files"
+          
+      - name: Change tag to the latest commit
+        if: ${{ github.ref_type == 'tag' }}
+        run: |
+          git tag -f $GITHUB_REF_NAME
+
       - name: Run Sphinx with sphinx-multiversion wrapper - branch trigger
         if: ${{ github.ref_type == 'branch' }}
         run: sphinx-multiversion docs/source cleanlab-docs -D smv_branch_whitelist=${{ github.ref_name }} -D smv_tag_whitelist=None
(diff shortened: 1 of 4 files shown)


jwmueller:
need to fix collab version before we merge

docs/source/conf.py
```diff
@@ -169,7 +169,7 @@
 )
 
 # Change this to "always" before running in the doc's CI/CD server
-nbsphinx_execute = "never"
+# nbsphinx_execute = "never"
```
jwmueller, line 172: This looks a bit sketchy, why not just delete this and the "Change this" comment above

.github/workflows/gh-pages.yaml
```diff
@@ -48,6 +48,13 @@ jobs:
       - name: Install dependencies
         run: python3 -m pip install -r docs/requirements.txt
 
+      - name: Find and replace "%pip install cleanlab" in .ipynb files
+        uses: jacobtomlinson/gha-find-replace@v2
+        with:
+          find: "%pip install cleanlab"
+          replace: "%pip install git+https://github.com/${{ github.repository }}.git@${{ github.sha }}"
```
jwmueller, line 55: Colab should only be installing from github when this is not a tagged version release.  If this is the latest stable version, then do not need to replace. If this is an older tagged version that is no longer the latest stable version, then we want to replace with:   
`"%pip install cleanlab==X.Y.Z" with X.Y.Z indicating the version number."

.github/workflows/gh-pages.yaml
```diff
@@ -48,6 +48,13 @@ jobs:
...
+          include: "**/*.ipynb"
```
jwmueller, line 56: should only be doing this auto find/replace inside `docs/source` right?

docs/source/conf.py
weijinglok: My bad, this was used in testing to speed up the build. I've removed it in the latest commit.

.github/workflows/gh-pages.yaml
weijinglok: done. added additional steps after the one above to handle this.

.github/workflows/gh-pages.yaml
weijinglok: done!

.github/workflows/gh-pages.yaml
```diff
@@ -48,6 +48,33 @@ jobs:
       - name: Install dependencies
         run: python3 -m pip install -r docs/requirements.txt
 
+      - name: Find and replace "%pip install cleanlab" with master branch in .ipynb files
+        if: ${{ github.ref_type == 'branch' }}
```
jwmueller, line 52: will this work for when we merge PRs into master also?

weijinglok: Yes, it should. It may look a little odd because there isn't an `if...else` statement here, so it's just 2 separate `if` statements - one for when the doc CI is triggered by a push to the `master` branch, another when it's triggered by a newly published release.

In pseudocode, the above-highlighted code snippet does the following:
1. If it's triggered by a push to a branch, i.e. the `master` branch in the `cleanlab/cleanlab` repo AKA our production repo
2. find `%pip install cleanlab  # for colab` in the `.ipynb` files and replace them with pip installing a specific commit hash of cleanlab. For example, if the doc build is run in the **`cleanlab/cleanlab`** repo and the merging commit hash is **`abc123`**, then the replacement will be:

`%pip install git+https://github.com/`**`cleanlab`**`/cleanlab.git@`**`abc123`**

if this is unnecessarily specific, setting it to download to a specific branch would work too, e.g. something like:

`%pip install git+https://github.com/`**`cleanlab`**`/cleanlab.git@`**`master`**

anishathalye (approved):

jwmueller (approved):
LGTM

PR #141, opened 2022-03-28
Utilites -> internal submodule refactor

jwmueller:
Renames `Utilies` submodule to `internal`.  Updates docs to display methods previously missing such as those in `label_quality_utils.py`. Clarifies missing steps in instructions to build docs locally and adds more information to contributing guidelines (eg. how to run unit tests).

PR #142, opened 2022-03-28
Fix NumPy deprecation warning

anishathalye:
`np.long` was deprecated in NumPy 1.20, and this was making the test output noisy with deprecation warnings. This patch replaces the type with `np.int32` (the labels are 0--9).


jwmueller (approved):
LGTM!  Will merge after https://github.com/cleanlab/cleanlab/pull/141

Issue #87, opened 2021-10-25
TypeError: cannot pickle '_thread.RLock' object


kothari1997narayan:
I tried implementing a simple **Keras** model, model description is mentioned below

>     input_img = Input(shape=(1350,))
>     x = Dense(units=32, activation='relu')(input_img)
>     x = Dropout(0.2)(x)
>     x = Dense(units=10, activation='relu')(x)
>     x = Dense(units=4, activation='softmax')(x)
>     model2 = keras.Model(input_img, x)
>     model2.compile(optimizer='adam', loss='categorical_crossentropy', metrics=['accuracy'])

Then wrapped it with the sklearn **BaseEstimator**
Also edited a line in utils.py to not check for shapes of X and Y(https://github.com/cleanlab/cleanlab/blob/master/cleanlab/util.py#L55)

>     from sklearn.base import BaseEstimator
> 
>     class YourFavoriteModel(BaseEstimator): # Inherits sklearn base classifier
>     
>         def __init__(self, model):
>             self.model = model
>         
>         def fit(self, X, y):
>             yt = to_categorical(y, num_classes=4)
>             self.model.fit(X,yt, epochs=20, shuffle=True)
>         
>         def predict(self, X):
>             return np.argmax(self.model.predict(X,axis=-1))
>     
>         def predict_proba(self, X):
>             return self.model.predict(X)
>                          
>         def score(self, X, y, sample_weight=None):
>             yt = to_categorical(y, num_classes=4)
>             return self.model.evaluate(X,yt)[1]
> 
>     from cleanlab.classification import LearningWithNoisyLabels
>     lnl = LearningWithNoisyLabels(clf=YourFavoriteModel(model2))
>     lnl.fit(X_train, y_train)

But while doing the fit operation getting **TypeError: cannot pickle '_thread.RLock' object**
Detailed error is printed below..

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-21-2af841381025> in <module>
     23 from cleanlab.classification import LearningWithNoisyLabels
     24 lnl = LearningWithNoisyLabels(clf=YourFavoriteModel(model2))
---> 25 lnl.fit(X_train, y_train)

~\AppData\Roaming\Python\Python38\site-packages\cleanlab\classification.py in fit(self, X, s, psx, thresholds, noise_matrix, inverse_noise_matrix)
    320                 self.py, self.noise_matrix, self.inverse_noise_matrix, \
[...]

PR #144, opened 2022-03-28
Fix sklearn estimator cloning

anishathalye:
Arbitrary scikit-learn estimators are not safe to clone via `copy.deepcopy()`, for example when they contain state that cannot be pickled, such as a Keras model that contains a `_thread.RLock` [[1]]. The correct way to clone an estimator is to use `sklearn.base.clone()` [[2]], which constructs a new unfitted estimator with the same parameters as the original by calling `get_params()` followed by creating a new object of the class, passing the parameters to `__init__()`. This patch switches from `copy.deepcopy` to `sklearn.base.clone`.

This patch also updates the MNIST PyTorch CNN example model to be compatible with `sklearn.base.clone`. Writing a custom `get_params` implementation usually wouldn't be necessary with a more standard estimator that inherits from `BaseEstimator`, but it was necessary to override the method in this case because of the nonstandard implementation that supports multiple datasets and does data loading within the estimator itself based on the `dataset` argument.

[1]: https://github.com/cleanlab/cleanlab/issues/87
[2]: https://scikit-learn.org/stable/modules/generated/sklearn.base.clone.html

### DIFF
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -31,8 +31,8 @@
 from sklearn.linear_model import LogisticRegression as LogReg
 from sklearn.model_selection import StratifiedKFold
 from sklearn.metrics import confusion_matrix
+import sklearn.base
 import numpy as np
-import copy
 
 from cleanlab.internal.util import (
     value_counts,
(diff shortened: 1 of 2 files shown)


jwmueller (approved):
LGTM

Issue #86, opened 2021-10-22
Allow dimension of X to be more than 2 (ValueError: Found array with dim 4. Estimator expected <= 2)


kothari1997narayan:
https://github.com/cleanlab/cleanlab/blob/e10d93a6af69882ad317af99926e66bde353db72/cleanlab/util.py#L55

If we use a CNN model as a classifier then it will require X to have more than 2 dimensions
But in this check_X_y of sklearn utils take default argument of **allow_nd(Whether to allow X.ndim > 2) as false**. 
So, it will throw an error( **ValueError: Found array with dim 4. Estimator expected <= 2**) and we won't be able to use any CNN models.

I think you should give this option to user or set this parameter to **True** in cleanlab/util.py#L55

anishathalye:
Great idea, thanks! Addressed in #148.

PR #148, opened 2022-03-29
Allow n-dim data in LearningWithNoisyLabels

anishathalye:
LearningWithNoisyLabels used `sklearn.utils.check_X_y` to enforce that `X` was 2D, to be in line with what sklearn's standard estimators expect. However, LearningWithNoisyLabels is dataset-agnostic: it doesn't look at the data points themselves. If the underlying classifier supports data in a different shape, there's no reason LearningWithNoisyLabels should prohibit it. Users have requested that we relax this unnecessary restriction [[1]] so LearningWithNoisyLabels will more naturally support e.g. image datasets and CNN models.

Thanks to @kothari1997narayan for suggesting this change.

[1]: https://github.com/cleanlab/cleanlab/issues/86

### DIFF
--- a/cleanlab/internal/util.py
+++ b/cleanlab/internal/util.py
@@ -54,6 +54,7 @@ def assert_inputs_are_valid(X, s, pred_probs=None):  # pragma: no cover
         dtype=None,
         force_all_finite=False,
         ensure_2d=False,
+        allow_nd=True,
     )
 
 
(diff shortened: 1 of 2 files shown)


jwmueller (approved):
LGTM!

jwmueller:
black complained though: https://github.com/cleanlab/cleanlab/runs/5730967872?check_suite_focus=true

PR #152, opened 2022-03-31
Enable use of find_label_issues_kwargs for hyper-parameter search

JohnsonKuan:
This PR is to move `find_label_issues_kwargs` parameter from the `fit()` method to the `__init__()` method of `LearningWithNoisyLabels()`.

Moving `find_label_issues_kwargs` to `__init__()` allows us to use it for hyper-parameter search. For example, scikit-learn [GridSearchCV](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html) requires hyper-parameters from `param_grid` to be passed to the the init of the estimator.


jwmueller (approved):
LGTM!

jwmueller (commented):
Good catch!

PR #157, opened 2022-04-02
added LearningWithNoisyLabels.find_label_issues instance method

jwmueller:
Adds instance method `LearningWithNoisyLabels.find_label_issues()`, which is now called as a helper method inside `LearningWithNoisyLabels.fit()`.  
`LearningWithNoisyLabels.fit()` still works exactly the same way as before from user's perspective. 

This change enables a new workflow:

Workflow 1:
```
clf = RandomForest()
lnl = LearningWithNoisyLabels(clf)
label_issues_mask = lnl.find_label_issues(X, y)
lnl.fit(X,y, label_issues_mask=label_issues_mask)  # will skip cross-validation and error-finding
lnl.fit(X,y)  # will not skip cross-validation and error-finding and will redo them.
```

Workflow 2:
```
clf = RandomForest()
lnl = LearningWithNoisyLabels(clf)
lnl.fit(X,y)  # will run cross-val and then filter.find_label_issues()
lnl.get_label_issues()  # instantly fetches stored errors  
lnl.find_label_issues(X,y)  # will be slow and recompute cross-validation + errors. 
```


cgnorthcutt (commented):
Getting there. A few main issues:

## Memoization prevents refitting

 - If you call `lnl.fit()` on a new dataset or fine-tune the model on new data, it will crash because the label error mask is being used from the previous data (and is a different length). If it doesn't crash because they are the same length, you'll get the wrong behavior.
 - You should throw a warning if this happens (and I'm not sure what to do if previous data and new data are the same length) and tell the user to set `memoize=False` in either `.fit()` or in the initialization.
 - need to add a memoize parameter (or name it something user friendly like `use_intermediate_results=False` (Default would be true).

## Remove get_label_issues()

If we end up going this route (still not clear due to issue above if we should make find_label_issues public), then probably should just delete `get_label_issues` as it no longer adds value.

## Avoid duplicated workflows.

Your third workflow seems redundant and I think we want users to call `cleanlab.filter.find_label_issues()` in your third workflow instead. This was part of the motivation for `get_label_issues()` just being called after `fit`

I'd actually recommend a different workflow entirely that's much simpler to compute (I think).

* just add a parameter `fit(final_training=False)`
* and keep `get_label_issues()` as is
* and if you call predict() and the final training never happened, then you do the final training (or you can just call fit again and final_training=True, it will do final fit)

I think this solves most of the issues above. The downside is its a little weird to call fit() and the model not be completely fitted (even if you told it not to). What do you think?

cgnorthcutt:
I think the way to implement this PR instead is:
* `self.find_label_issues()` always computes (no memoization) and saves the attribute _label_issues_mask
* `self.fit()` can take a parameter `label_issues_mask`. if its passed in, you skip all that computation and just do final fitting. This function totally ignores if self.label_issues_mask exists (no memoization given, but the user can pass in their own if they want)
* `self.get_label_issues()` is the accessor (keep this)

This enables the following workflow:

```python
clf = LearningWithNoisyLabels()
label_issues_mask = clf.find_label_issues(X, y)
# Now I don't want to waste time when I call fit, so I can do this if I want to save time
clf.fit(X, y, label_issues_mask=label_issues_mask)  # only does final fitting, skips cross validation
clf.predict(X_test)
```

This should support:
1.  grid search (which would otherwise break in the original PR if we support memoization since the previous label_errors_mask would be used when refitting the model on new hyperparams).
2. calling fit() multiple times on new data.

Consider adding unit tests for both of these two cases.

cleanlab/classification.py
```diff
@@ -293,8 +293,153 @@ class label.
         tuple
           (label_issues_mask, sample_weight)"""
 
+        clf_final_kwargs = {**clf_kwargs, **clf_final_kwargs}
```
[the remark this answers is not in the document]
jwmueller: I prefer the shorter name, otherwise need to also rename clf_kwargs

jwmueller:
updated functionality to not rely on memoization in `LearningWithNoisyLabels.find_label_issues()`

cleanlab/classification.py
```diff
@@ -277,6 +278,12 @@ class label.
           inverse_noise_matrix will be computed from pred_probs and labels.
           Assumes columns of inverse_noise_matrix sum to 1.
 
+        label_issues_mask : np.array<bool>, default = None
+            A boolean mask for the entire dataset such as those returned by:
+            `LearningWithNoisyLabels.find_label_issues()` or `filter.find_label_issues()`.
+            If specified, datapoints corresponding to False entries will be pruned from the data before
+            training the `clf` model.
+            Specifying this argument can save a lot of time if you have already estimated the issues in your data.
```
JohnsonKuan, line 286: nit: can indent 2 spaces

cleanlab/classification.py
```diff
@@ -293,8 +300,155 @@ class label.
...
+        from this LearningWithNoisyLabels instance, use `get_label_issues()` instead.
+
+        This is the method called to find label issues inside `LearningWithNoisyLabels.fit()`.
+        For descriptions of the arguments, see `LearningWithNoisyLabels.fit()` documentation.
+
+        Returns
+        -------
+        label_issues_mask : np.array<bool>
+            This method returns a boolean mask for the entire dataset where True represents
+            a label issue and False represents an example that is accurately labeled with high confidence.
```
JohnsonKuan, line 439: same as previous comment, can indent 2 spaces to match other docstrings

cleanlab/classification.py
```diff
@@ -404,105 +556,21 @@ class label.
...
-                self.clf.predict(X),
-                sample_weight=sample_weight,
-            )
+        return self.label_issues_mask
 
     def get_label_issues(self):
-        """Accessor. Returns `self.label_issues_mask` computed via `filter.find_label_issues()`"""
+        """Accessor. Returns `self.label_issues_mask` if previously already computed."""
 
         return self.label_issues_mask
```
JohnsonKuan, line 575: I was thinking it'd be nice to raise a warning and print a message if the user calls `get_label_issues()` when `self.label_issues_mask` is None. 

Something like: `"label_issues_mask has not been computed yet! To compute label_issues_mask, please run find_label_issues() or fit() first."`. This can be a separate PR though.

JohnsonKuan (commented):
LGTM. Just had a few minor comments on docstring indentation.

cleanlab/classification.py
cgnorthcutt: +1 Good idea. I'd add this in this PR. Its just one if statement and a warnings import. Seems useful to educate users.

cleanlab/classification.py
```diff
@@ -404,105 +556,21 @@ class label.
             self.confident_joint[self.pulearning][1 - self.pulearning] = 0
             self.confident_joint[1 - self.pulearning][1 - self.pulearning] = 1
 
-        # This is the actual work of this function.
         if self.verbose:
             print("Using predicted probabilities to identify label issues ...")
-        # Get the indices of the examples we wish to prune
+        # Get the indices of label issues
         self.label_issues_mask = find_label_issues(
```
cgnorthcutt, line 562: Change the import at the top to be `from cleanlab import filter` and then change this call everywhere to be `filter.find_label_issues()` to distinguish between the two.

cgnorthcutt: ^^ do this everywhere. Otherwise its very hard to distinguish between the two functions that are named identically. You could also consider changing the function name so that its not identical.

cleanlab/classification.py
```diff
@@ -404,105 +556,21 @@ class label.
...
-                self.clf.predict(X),
-                sample_weight=sample_weight,
-            )
+        return self.label_issues_mask
 
     def get_label_issues(self):
-        """Accessor. Returns `self.label_issues_mask` computed via `filter.find_label_issues()`"""
+        """Accessor. Returns `self.label_issues_mask` if previously already computed."""
 
         return self.label_issues_mask
```
cgnorthcutt, line 576: `_process_label_issues_kwargs` has several changes needed:
* it should not set global constants. I'd move MIN_EXAMPLES_PER_CLASS = 10 as a class global (why is it 10 btw? please clarify in a comment) and then just pass this global wherever its needed in a function call and remove this from here
* the docstring should be "Throws an error if you pass in a kwargs argument to filter.find_label_issues that is not supported by the LearningWithNoisyLabels.find_label_issues() function"
* the comment below the docstring should say "# Different defaults to use for LearningWithNoisyLabels.find_label_issues() vs filter.find_label_issues():"
* The confident_joint is being set here to `self.confident_joint` but if you follow the code in `LearningWithNoisyLabels.find_label_issues()` you'll see it gets set again afterwards. Instead the logic in `LearningWithNoisyLabels.find_label_issues` needs to be changed so that if the confident joint exists, you just compute `noise_matrix` and `inverse_noise_matrix` using `count.estimate_latent(labels, confident_joint)` (this would go just before the part where you compute the `pred_probs` if they don't exist.
* given all these issues i strongly recommend removing this function entirely and just set `return_indices_ranked_by=""` and `multi_label=False` directly when `filter.find_label_issues()` is called (we are going to add multi_label support anyway so that part will get removed soon too.)

cgnorthcutt (changes requested):
Several superficial changes. Also non-superficial changes with regards to `_process_label_issues_kwargs`

cleanlab/classification.py
cgnorthcutt: implemented.

cleanlab/classification.py
```diff
@@ -293,8 +300,155 @@ class label.
...
+            # Re-weight examples in the loss function for the final fitting
+            # labels.t. the "apparent" original number of examples in each class
+            # is preserved, even though the pruned sets may differ.
+            if self.verbose:
+                print(
+                    f"Assigning sample weights for final training based on estimated label quality ..."
+                )
+
+            self.sample_weight = np.ones(np.shape(labels_cleaned))
+            for k in range(self.K):
```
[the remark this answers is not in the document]
cgnorthcutt: fixed

cleanlab/classification.py
cgnorthcutt: fixed

cleanlab/classification.py
```diff
@@ -293,8 +300,155 @@ class label.
...
+        The resulting label issue mask is saved as an attribute of this LearningWithNoisyLabels instance.
+        Kwargs for `filter.find_label_issues` must have already been specified
+        in the initialization of LearningWithNoisyLabels, not here.
+
+        Unlike `filter.find_label_issues`, which requires `pred_probs`,
+        this method only requires a classifier and it can do the cross-validation training for you.
+        Both methods return the same boolean mask that identifies which datapoints have label issues.
+
+        Note: This method computes the label issues from scratch, to access previously-computed label issues
+        from this LearningWithNoisyLabels instance, use `get_label_issues()` instead.
```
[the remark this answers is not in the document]
cgnorthcutt: fixed.

cleanlab/classification.py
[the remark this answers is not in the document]
cgnorthcutt: changed.

cleanlab/classification.py
```diff
@@ -293,8 +300,155 @@ class label.
...
+                pred_probs=pred_probs,
+                thresholds=thresholds,
+                noise_matrix=noise_matrix,
+                inverse_noise_matrix=inverse_noise_matrix,
+                clf_kwargs=clf_kwargs,
+            )
+        elif self.verbose:
+            print("Using provided label_issues_mask instead of finding label issues.")
+
+        self.label_issues_mask = label_issues_mask
```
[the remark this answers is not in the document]
cgnorthcutt: done.

cleanlab/classification.py
```diff
@@ -404,105 +556,21 @@ class label.
             self.confident_joint[self.pulearning][1 - self.pulearning] = 0
             self.confident_joint[1 - self.pulearning][1 - self.pulearning] = 1
 
-        # This is the actual work of this function.
         if self.verbose:
             print("Using predicted probabilities to identify label issues ...")
-        # Get the indices of the examples we wish to prune
+        # Get the indices of label issues
```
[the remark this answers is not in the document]
cgnorthcutt: done

cleanlab/classification.py
```diff
@@ -404,105 +556,21 @@ class label.
...
-        ----------
-        X : :obj:`np.array` of shape (n, m)
-          The test data as a feature matrix."""
-
-        return self.clf.predict(*args, **kwargs)
-
-    def predict_proba(self, *args, **kwargs):
-        """Returns a vector of probabilities P(true_label=k)
-        for each example in X.
+            print(f"Identified {np.sum(self.label_issues_mask)} datapoints with label issues.")
```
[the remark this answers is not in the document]
cgnorthcutt: updated everywhere

cleanlab/classification.py
cgnorthcutt: done

cleanlab/classification.py
cgnorthcutt: fixed the second, third, and fourth bullets and marked the last bullet as a TODO.

cgnorthcutt:
#### Main fixes
* I resolved all issues and added support for confident joint to be used properly in `LearningWithNoisyLabels.find_label_issues()`.
* Added several tests of this functionality as well at the end of `tests.test_classification.py`. 
* All tests are passing.
* Fixed all other issues brought up in this PR.

#### Other
* Fixed a few other very minor bugs -- print statements we're in the wrong place in classification.
* The confident joint was randomly being output in a function. Removed.
* Removed unused imports.
* other miscellaneous minor fixes.

PR LGTM for squash and merge.

cgnorthcutt (approved):
Made the changes that I requested in this PR: https://github.com/cleanlab/cleanlab/pull/157/commits/9246ff3133904b39adc07ebcd6c79d4a1022fc29

PR #161, opened 2022-04-03
Remove fastText-related code

anishathalye:
The motivation behind this change is to not have any untested code in here that will just become out of sync with the rest of the repo and become useless over time. If we really want a fastText example somewhere, I'd suggest creating a `cleanlab-contrib` repo or similar where we can keep untested code like this that may be useful to some people.

Related to this: should we also remove `example_models/cifar_cnn.py`? Or add a test for it?


jwmueller:
Totally agree we don't want untested code in here. 
How about we move this and the other models to: `cleanlab/examples/` repo inside a `contrib/` or `experimental/` folder with disclaimer in this folder's README that this code is experimental and not maintained?

jwmueller (approved):
I vote we move all of the code in `example_models/` out of `cleanlab` repo and into: 
the `examples` repo (as well as the related stuff from the Readme.md).

@anishathalye Do you want to handle this?

Note: there is open PR to add `contrib/` folder to the `examples` repo:
https://github.com/cleanlab/examples/pulls

I'm not sure these files need to live in `examples/contrib/` though, they could just be in `examples/models/` for now (@cgnorthcutt mentioned that they all still work).

anishathalye:
Closing this in favor of another plan we're implementing (which will be two separate PRs)

- @jwmueller : fastText stays in this repo but moved to an `experimental` module where we don't make compatibility guarantees.
- @anishathalye : the other models (cifar_cnn and mnist_pytorch) get moved to https://github.com/cleanlab/examples, and we get rid of `example_models`

PR #162, opened 2022-04-03
Remove unnecessary version warning

anishathalye:
This was used to skip e.g. importing PyTorch on some tests. However, PyTorch supports all versions of Python that Cleanlab supports, so there's no need to issue this warning. Furthermore, having this kind of code in our codebase increases maintenance burden and contributes to user frustration. If we mark a version of some library as unsupported, but the library adds support for it in the future, there's no way users can use the two packages together without waiting for us to push a new version of cleanlab that marks that it's compatible and then switching to the new version, which may be painful for unrelated reasons. It's better to not have these kinds of checks; if users are using a version of Python that's unsupported by some other library that they want to use, that's not our problem, and they'll have issues installing or importing it and figure it out themselves.

Note: this patch doesn't remove the version warnings in the fastText-related code because #161 deletes that code.


jwmueller (approved):
LGTM but will defer to @cgnorthcutt to merge

cleanlab/internal/util.py
```diff
@@ -430,54 +430,3 @@ def print_joint_matrix(joint_matrix, round_places=2):
...
-        {}
-        cleanlab supports Python >= 3.6, but you are using Python version {}.
-        You'll need to use a version compatible with both.""".format(
-            warning_str, v
-        )
-        warnings.warn(warning)
-        return False
-
-
-class VersionWarning(object):
```
cgnorthcutt, line 464: Effectively we are saying that we will support all versions of python 3.6 and up and if you use pytorch or external libraries that don't support a version of python that cleanlab, supports, this will fail and you'll get a generic warning that you need to figure out. Overall, seems okay given that most libraries we interact with support similar versions than we do. It certainly makes the code a lot cleaner :)

cgnorthcutt (approved):
LGTM. We may have future dependencies requiring version warnings in the future, but I see no need to keep this code around while it is not at all in use or plans to be in use for the foreseeable future.

PR #163, opened 2022-04-03
Make tutorial file names consistent

anishathalye:
The choice of path and filename here is not inconsequential; it shows up in the URL for the docs, e.g. https://docs.cleanlab.ai/v2.0.0/tutorials/image.html


jwmueller:
This is already being addressed in Wei Jing's PR:
https://github.com/cleanlab/cleanlab/pull/154

Probably better to just go with that one than dealing with merge conflicts for name-change.

anishathalye:
Closed in favor of #154.

PR #169, opened 2022-04-05
Add CI check that .ipynb outputs are empty

### LINKED ISSUE #168: Add CI to ensure that Jupyter notebooks don't include output
Preserving the output cells in Jupyter notebooks that are checked in to this repo bloat the repo size, and keeping the output is not necessary because the notebooks are run as part of the CI for producing docs.

For example, in #165, the `audio.ipynb` is 112K (and now part of the repo history forever), while clearing the output cells reduces that size down to 24K. I suspect that Git's delta compression will also work better without the output cells there as we make changes to the notebook. Also, the diffs will be more readable.

We should add CI that enforces that Jupyter notebooks checked in to the repo don't have any output cells. In the meantime, let's be careful and do this manually. CC @weijinglok @JohnsonKuan @jwmueller

anishathalye:
Fixes #168

### DIFF
--- a/docs/source/tutorials/image.ipynb
+++ b/docs/source/tutorials/image.ipynb
@@ -358,7 +358,7 @@
    "hash": "ced20e3e49bb4fa4ce8ad38f8f2535b7fc4c39b2b89554502b5dbdad1ad67eda"
   },
   "kernelspec": {
-   "display_name": "Python 3.10.0 64-bit",
+   "display_name": "Python 3 (ipykernel)",
    "language": "python",
    "name": "python3"
   },
@@ -372,7 +372,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.10"
+   "version": "3.9.10"
   }
  },
  "nbformat": 4,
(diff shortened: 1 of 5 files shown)


docs/source/tutorials/image.ipynb
```diff
@@ -358,7 +358,7 @@
    "hash": "ced20e3e49bb4fa4ce8ad38f8f2535b7fc4c39b2b89554502b5dbdad1ad67eda"
   },
   "kernelspec": {
-   "display_name": "Python 3.10.0 64-bit",
+   "display_name": "Python 3 (ipykernel)",
```
jwmueller, line 361: just highlighting this in case it's unintentional

docs/source/tutorials/image.ipynb
```diff
@@ -372,7 +372,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.10"
+   "version": "3.9.10"
```
jwmueller, line 375: just highlighting this in case it's unintentional

jwmueller (approved):
LGTM, great call on this!

Issue #158, opened 2022-04-02
Incorrect handling of zero entries in true_labels_class_counts


anishathalye:
We don't correctly handle the case where there is a class where there isn't a single example where confident label = given label.

Potentially related to #156.

To reproduce:

```python
import numpy as np
from cleanlab.count import compute_confident_joint, estimate_latent

pred_probs = \
np.array(
[[0.27916167 , 0.14589103 , 0.29264585 , 0.28230144],
 [0.13429196 , 0.12536383 , 0.47943979 , 0.26090442],
 [0.41348584 , 0.13463275 , 0.25845595 , 0.19342546],
 [0.27753469 , 0.12295569 , 0.33125886 , 0.26825075],
 [0.11649856 , 0.11219034 , 0.51857122 , 0.25273988],
 [0.38010026 , 0.25572261 , 0.1330541  , 0.23112304],
 [0.31583755 , 0.1363069  , 0.29246806 , 0.2553875 ],
 [0.30240076 , 0.16925207 , 0.26499082 , 0.26335636],
 [0.44524505 , 0.27410085 , 0.08305069 , 0.19760341],
 [0.22903975 , 0.07783631 , 0.38035414 , 0.3127698 ],
 [0.2507156  , 0.120729   , 0.32551729 , 0.30303812],
 [0.43809229 , 0.14401381 , 0.208393   , 0.2095009 ],
 [0.20749181 , 0.11883556 , 0.38402152 , 0.28965111],
 [0.43840254 , 0.13538447 , 0.24518806 , 0.18102493],
 [0.28504779 , 0.1030975  , 0.34258602 , 0.26926868],
 [0.38425408 , 0.29168969 , 0.15181255 , 0.17224368],
 [0.19339907 , 0.10804265 , 0.37570368 , 0.3228546 ],
 [0.21509781 , 0.07190167 , 0.38914722 , 0.3238533 ],
 [0.27040334 , 0.1303784  , 0.3284232  , 0.27079507],
 [0.4059021  , 0.1671356  , 0.24889193 , 0.17807036]])
labels = np.array([3, 3, 1, 3, 0, 2, 2, 2, 0, 2, 2, 1, 0, 0, 0, 0, 2, 1, 3, 3])

cj = compute_confident_joint(labels, pred_probs, calibrate=False)
print(f'confident joint: {cj}\n')

py, noise_matrix, inv_noise_matrix = estimate_latent(cj, labels)
[...]

PR #178, opened 2022-04-06
Simple fix to Issue 158 (and potentially other issues)

cgnorthcutt:
Fixes https://github.com/cleanlab/cleanlab/issues/158

A simple fix to clip the diagonal confident joint so that it avoids columns entirely of zero and always requires one correctly labeled example in every class.

* This is the lowest-level place where the confident joint is computed. The fix (should) trickle to all downstream use cases.
  - In your review, consider if this is the case.
* this PR reduces the warnings in our tests by half


docs/source/tutorials/tabular.ipynb
```diff
@@ -391,7 +391,7 @@
    "source": [
     "# Hidden code cell to check that cleanlab has improved prediction accuracy\n",
     "if acc_og >= acc_cl:\n",
-    "    raise Exception(\"Cleanlab training failed to improve model accuracy.\")"
+    "    raise Exception(\"Cleanlab training failed to improve model accuracy.\")\n"
```
jwmueller, line 394: this \n is causing notebook linter to fail.

jwmueller (approved):
LGTM

cleanlab/count.py
```diff
@@ -410,6 +410,8 @@ def compute_confident_joint(
     true_labels_confident = true_label_guess[at_least_one_confident]
     labels_confident = labels[at_least_one_confident]
     confident_joint = confusion_matrix(true_labels_confident, labels_confident).T
+    # Guarantee at least one correctly labeled example is represented in every class
+    np.fill_diagonal(confident_joint, confident_joint.diagonal().clip(min=1))
```
jwmueller, line 414: Could consider printing warning if the results are likely  unreliable if there was a 0 before.  I do not know how to judge that.

docs/source/tutorials/tabular.ipynb
anishathalye: Why is there a change to this notebook at all?

anishathalye:
I rebased onto c4e8462.

anishathalye (approved):
Summarizing discussion offline: decided to skip warning in this situation.

LGTM.

docs/source/tutorials/tabular.ipynb
cgnorthcutt: fixed

cleanlab/count.py
cgnorthcutt: It shouldn't make a big difference. Will leave as is for now. we can re-evaluate if this seems to cause issues down the line.

PR #180, opened 2022-04-07
Update docs README

weijinglok:
Update docs README to include:

1. Warning that code changes have to be committed to a branch to be reflected in the docs build. 
2. Instructions to not execute Jupyter Notebooks during the docs build process.
3. Info to ignore `unknown config value 'smv_branch_whitelist' in override, ignoring` and `unknown config value 'smv_tag_whitelist' in override, ignoring`, after entering build command.


jwmueller (approved):
Thanks!

PR #182, opened 2022-04-07
Introducing the new Dataset Module for cleanlab 2.0

cgnorthcutt:
# Dataset module

This module introduces the new API for interacting and working with the entire datasets, with suggestions for overlapping classes that you may want to merge, low quality classes you may want to remove, overall dataset health score, and a fully health summary report (generate for your dataset in a single line of code).

* Code should probably be tested a bit more thoroughly before merge.
* Merge or rebase this commit. **Do not squash and merge.**

# need to add tests!
* please add tests as you review.

# Updates made to module during this PR:
* Merged the two label quality and label noise methods.
* report conditional matrix scores (instead of joint) for class methods
* All methods that previously returned multiple objects now work with and return pandas dataframes.
* Added support for string classes
* Added a health_summary function that prints everything about your dataset health in one line of code.
* completely rewrote the docstrings.
* all methods take in all input types.
* all methods have been renamed.
* Added testing for four real-world datasets.

# Future support

* in the future we should add support for all the functions from the raw data directly + a model (like `CleanLearning`), but that's for a later release.


cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+    via the approach published in (Northcutt, Jiang, Chuang (2021) JAIR).
+    paper: https://arxiv.org/abs/1911.00068
+
+    This method works by providing any one (and only one) of the following inputs:
+    1. confident_joint
+    2. joint
+    3. labels and pred_probs
+    but it is not necessary to provide a combination, e.g. the labels and pred_probs and joint.
+
+    For parameter info, see the docstring of `count.estimate_joint`
```
jwmueller, line 51: The docstring for `count.estimate_joint` does not seem to explain some of these args like `joint`, `asymmetric`. Furthermore the docstring of `count.estimate_joint` does not really explain its parameters either: 
https://docs.cleanlab.ai/master/cleanlab/count.html#cleanlab.count.estimate_joint
It just says:  See `cleanlab.count.calibrate_confident_joint` docstring.

jwmueller: Even if `count.estimate_joint` did have good explanations of its parameters, I would prefer to explain the parameters separately here.  The reason is that `count.estimate_joint` seems like a more advanced function that requires understanding concepts, whereas `find_similar_classes` is supposed to be a super accessible function easily used by anybody.

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+import numpy as np
+from count import estimate_joint, compute_confident_joint
+
+
+def find_similar_classes(
+    joint=None,
+    asymmetric=True,
+    *,
+    labels=None,
+    pred_probs=None,
```
jwmueller, line 33: I find this argument order a bit peculiar.
Consider having the arg order be like this match the other parts of package:
```
labels=None,
pred_probs=None,
*,
joint=None,
confident_joint=None,
asymmetric=True,
```

In particular, I'd think about it like this: 100% of cleanlab users will know what `labels` and `pred_probs` are. But only say 50% of users will know what `joint` and `confident_joint` are.

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+Supports dataset-level and class-level automated quality, including finding which classes should
+be merged with other classes in your dataset, which classes should be removed,
+and which classes have the best label accuracy, and which have the worst labels.
+"""
+
+import numpy as np
+from count import estimate_joint, compute_confident_joint
+
+
+def find_similar_classes(
```
jwmueller, line 28: Should we replace "similar" -> "commonly_confused" like we had in our cleanlab-studio report?

One reason is similar_classes just sounds like something I could get from a confusion matrix from model's predictions.

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+    multi_label=False,
+):
+    """Ranks all pairs of classes by their similarity and returns the ordered pairs as well as
+    a similarity score between 0 and 1. Use this function to determine which classes in a dataset
+    should be merged.
+
+    The method uses the joint distribution of noisy and true labels to compute ontological issues
+    via the approach published in (Northcutt, Jiang, Chuang (2021) JAIR).
+    paper: https://arxiv.org/abs/1911.00068
+
```
jwmueller, line 44: I think we want to clarify how this method differs from the traditional confusion matrix of a model's predictions.

Eg. while confusion matrix in traditional classification measures how similar a model thinks two classes are, this function measures how often the annotators confuse two classes.

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+    asymmetric : bool (default: True)
+      If `asymmetric==True`, includes both pairs (class1, class2) and (class2, class1). Use this for
+      finding "is a" relationships where for example "class1 is a class2".
+      If `asymmetric==False`, the pair (class1, class2) will onyl be returned once and order is
+      arbitrary (internally this is just summing score(class1, class2) + score(class2, class1)).
+
+    Returns
+    -------
+        A tuple of 2 ordered lists: (list_of_pairs_of_classes, likelihood_they_should_be_merged)"""
```
jwmueller, line 65: likelihood_they_should_be_merged is not well-defined.
I think it may be overly bold to claim "should_be_merged".

Two classes could be fundamentally confusing to annotators while still not mergeable (eg. "benign tumor" vs "malignant tumor" in medical images).

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+    ----------
+    See `count.estimate_joint` docstring.
+
+    asymmetric : bool (default: True)
+      If `asymmetric==True`, includes both pairs (class1, class2) and (class2, class1). Use this for
+      finding "is a" relationships where for example "class1 is a class2".
+      If `asymmetric==False`, the pair (class1, class2) will onyl be returned once and order is
+      arbitrary (internally this is just summing score(class1, class2) + score(class2, class1)).
+
+    Returns
```
jwmueller, line 63: Consider adding `verbose` arg, and also having this function print out the info in a pretty format.

jwmueller: Eg. I'd personally have this function organize this info in a pd.DataFrame something like this:
```
class1  class2  confusion_score
A          B           0.95
```
That can then easily be printed in nice format (and manipulated in various ways, which the current returned structure is harder to manipulate).

In future could consider plotting it as a heatmap, just like folks normally do with confusion matrix.

cleanlab/dataset.py
jwmueller: What is advantage of this return type vs say:
List[tuple] where each tuple = (class1, class2, confusion_score)

Or pd.DataFrame outlined in above comment.

jwmueller: the confusion_score (what you're calling likelihood_they_should_be_merged) should have some intuitive definition of what it's value represents right?

cleanlab/dataset.py
jwmueller: For the top few most-confused class-pairs, it'd also be nice to print more info like:

cleanlab found 124 examples with given label A that should actually be labeled as class B

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+    For parameter info, see the docstring of `count.estimate_joint`
+
+    Parameters
+    ----------
+    See `count.estimate_joint` docstring.
+
+    asymmetric : bool (default: True)
+      If `asymmetric==True`, includes both pairs (class1, class2) and (class2, class1). Use this for
+      finding "is a" relationships where for example "class1 is a class2".
+      If `asymmetric==False`, the pair (class1, class2) will onyl be returned once and order is
```
jwmueller, line 60: typo: onyl

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+      arbitrary (internally this is just summing score(class1, class2) + score(class2, class1)).
```
jwmueller, line 61: score(class1, class2) needs to be defined somewhere. Ideally it should have intuitive definition (like "proportion of examples with given label A that should likely be actually labeled as B instead")

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+        A tuple of 2 ordered lists: (list_of_pairs_of_classes, likelihood_they_should_be_merged)"""
+
+    def _find_largest_pairs(matrix):
+        """Helper function that returns the sorted (descending) (row, column) index pairs
+        and their values for an arbitrary 2-d numpy matrix."""
+
+        # Find indices of similar classes
+        idx = np.array(list(zip(*np.unravel_index(np.argsort(matrix.ravel()), matrix.shape)))[::-1])
+        # Bring max score between 0.1 and 1.0, otherwise score scales inversely with num_classes ^ 2
+        scaling_factor = 10 ** np.floor(-np.log10(np.max(matrix)))
```
jwmueller, line 74: Ideally the score has intuitive definition, which I'm not sure this rescaled one does?

jwmueller: why not just have the scores simply be this?

P(true_label = i | given_label = j)

That seems very interpretable to me

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+            f"parameter 'rank_by' should be 'given_label', 'true_label', or"
+            "'total_noise', but was {rank_by}"
+        )
+    class_indices, scores = np.argsort(scores), np.sort(scores)
+    if descending:
+        class_indices, scores = class_indices[::-1], scores[::-1]
+    return class_indices, scores
+
+
+def get_classes_ranked_by_label_quality(joint, descending=True):
```
jwmueller, line 132: could shorten to: `rank_classes_by_label_quality`

jwmueller: why not take in the similar args like `find_similar_classes()` does?

That function seems far more flexible. In particular, I think 100% of users will know about `labels` and `pred_probs` but many many not know what is `joint` (so I prefer to make this optional if user supplies labels & pred_probs)

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+        # Sum the upper and lower triangles and remove the lower triangle and the diagonal
+        # This provides values only in (the upper triangle) of the matrix.
+        sym_joint = np.triu(joint) + np.tril(joint).T
+        sym_joint = sym_joint - np.eye(len(sym_joint)) * sym_joint.diagonal()
+        idx, scores = _find_largest_pairs(sym_joint)
+        cutoff = (len(joint) ** 2 - len(joint)) / 2  # Omit lower triangle and diagonal elements
+        return idx[:cutoff], scores[:cutoff]
+
+
+def get_classes_ranked_by_label_noise(joint, descending=True, rank_by="given_label"):
```
jwmueller, line 101: All my comments below for `get_classes_ranked_by_label_quality` apply to this method as well

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+    class_indices, scores = np.argsort(scores), np.sort(scores)
+    if descending:
+        class_indices, scores = class_indices[::-1], scores[::-1]
+    return class_indices, scores
+
+
+def get_classes_ranked_by_label_quality(joint, descending=True):
+    """Returns the ordered class indices and associated scores quantifying the quality of the labels
+     in each class (larger number implies higher quality labels) for the joint of a dataset.
```
jwmueller, line 134: Clarify what is different about this method vs `get_classes_ranked_by_label_noise`

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+    Score values are unnormalized. What matters is relative ranking across the classes."""
+
+    scores = joint.diagonal()
+    class_indices, scores = np.argsort(scores), np.sort(scores)
+    if descending:
+        class_indices, scores = class_indices[::-1], scores[::-1]
+    return class_indices, scores
+
+
+def dataset_label_quality_score(joint):
```
jwmueller, line 145: could consider alternative name `overall_label_quality_score` or `global_label_quality_score`

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+    """Returns a single metric for the quality of the labels in a dataset. The score is between
```
jwmueller, line 146: "the labels" -> "all labels"

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+    0 and 1. A higher score implies higher quality labels, with 1 implying labels that have no
+    issues."""
```
jwmueller, line 148: data scientists will appreciate if you provide an intuitive definition of what the score means

cleanlab/dataset.py
jwmueller: again consider allowing many of the same args as `find_similar_classes`

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+
```
jwmueller, line 149: consider verbose arg and printing some interesting info like:
"Overall 3.6% of the labels in the dataset are estimated to have issues"

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
+
+
+"""Dataset Module
+Supports dataset-level and class-level automated quality, including finding which classes should
+be merged with other classes in your dataset, which classes should be removed,
+and which classes have the best label accuracy, and which have the worst labels.
```
jwmueller, line 21: "the best label accuracy, and which have the worst labels"
->
"tend to be annotated most/least correctly overall"

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+# cleanlab is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
+
+
+"""Dataset Module
```
jwmueller, line 18: consider demonstrating these functions at end of one of the quickstart tutorials, maybe `text.ipynb`

cleanlab/dataset.py
cgnorthcutt: I would think "confused" sounds more like it comes from a "confusion" matrix 🤣

cleanlab/dataset.py
cgnorthcutt: My gut also was to return a dataframe. I didn't because right now Cleanlab doesn't depend on pandas. Can you think of any downsides to introducing pandas as a new dependency? I kept it to just numpy and small fast packages to keep Cleanlab lean as long as possible. Thoughts @anishathalye and @jwmueller ?

cleanlab/dataset.py
cgnorthcutt: Agree. Any recommendations? It's based on the joint probability.

cleanlab/dataset.py
cgnorthcutt: I would think p(given|true) is more intuitive since typically you condition on the latent truth? How did you make that choice? Also why do you like the conditional over the joint? The joint is more expressive albeit the numbers are tiny.

I agree the scaling factor is somewhat arbitrary, it just makes it a good experience since the numbers will always be something readable, but definitely plan to improve it and happy to immediately. Or simply remove it.

cleanlab/dataset.py
jwmueller: Ya that is indeed annoying as well... 

Also sorry I didn't even meant confusion matrix in my criticism of "similar" but rather all that's needed is model's pred_probs to find similar classes (can just measure correlation in pred_probs).

Alternatives names for this function could include:
`find_conflated_classes`
`find_swapped_classes`
`label_error_types`
`mislabeling_rates`

cleanlab/dataset.py
jwmueller: I think it'll eventually have to support pandas in order to avoid having all sorts of arrays/lists all over tracking different data issues. Will probably add pandas dependency in 2.1 once multiple data issues are being flagged in a single run. And most data scientists will anyway already be using pandas (I've never personally had issue getting pandas to play nice with another package).  

For v2.0, I don't have strong preference, but it does seem like the return types are being somewhat sacrificed by not having pandas. And if we anyway want to add pandas dependency  in v2.1, then at that point would not want to be stuck with this suboptimal return type.

jwmueller: Another option is to lazy import pandas inside each function with an informative error msg if it fails, so pandas is only strictly needed for user to utilize these `dataset` functions.

cleanlab/dataset.py
jwmueller: Hm... as a user the information that I'd be most interested in is:
234 examples labeled as class A probably should actually be labeled as class B instead.

So I'd print that at least. 
If scores  = joint probability, you could explain it as:
"score(A,B) = The proportion of examples in the dataset that are labeled as class A but probably should actually be labeled as class B instead."

cleanlab/dataset.py
jwmueller: I thought conditioning on given is more natural because that is what user sees in their data.
So eg. I know that if I see example with given label A, it is untrustworthy.

I'm less sure what I could do with the information that many examples truly should be labeled B but are labeled something else (beyond realizing that my data is under-representing the occurrence of class B).

Re conditional vs joint, I'll leave it up to you. I think both are useful depending on the user's goal: joint is more useful if I want to maximally improve my dataset by fixing A->B mislabels that are overall common. conditional is more useful if I want to compare which given label is most/least trustworthy.

cleanlab/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+Supports dataset-level and class-level automated quality, including finding which classes should
+be merged with other classes in your dataset, which classes should be removed,
+and which classes have the best label accuracy, and which have the worst labels.
+"""
```
jwmueller, line 22: It could be nice if all the functions in this file support optional arg:
```class_names = None```
where user can provide mapping between string names for the classes and int representations.
Eg. `class_names` supported inputs could be dict and list:
```
{0: 'classA', 1: 'classB',...}
```
and 
```
['classA', 'classB']
```

Then if `class_names is not None`, all print statements / outputs are presented in terms of class names instead of ints.

This can also be for future PR.

File: cleanlab/dataset.py (hunk 3 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:28]: how about `find_classes_to_merge` ?

File: cleanlab/dataset.py (hunk 4 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:44]: done.

File: cleanlab/dataset.py (hunk 1 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:51]: added params to both

File: cleanlab/dataset.py (hunk 6 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:63]: made pandas a dependency for cleanlab. method now returns pd.DataFrame

File: cleanlab/dataset.py (hunk 9 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:74]: I like the idea of implementing the noise matrix version of the scoring (because the number will be normalized and bigger), but the problem is that when `symmetric==True`, it lose its meaning since this i computed by summing the P(x,y) + P(y, x) which doen't make sene if its p(x|y) + P(y|x). Added this as something to think about for 2.1 and moving on with using the joint.

Removed the scaling entirely though, so it will return small joint probability score.

File: cleanlab/dataset.py (hunk 15 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:148]: added

File: cleanlab/dataset.py (hunk 13 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:145]: Changed to `overall_label_health_score` and added args.

File: cleanlab/dataset.py (hunk 17 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:21]: changed

File: cleanlab/dataset.py (hunk 18 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:18]: I won't have time for this on my end, but leave this up for grabs! xD

File: cleanlab/dataset.py (hunk 19 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:22]: it actually even simpler! just needs to be a list with the class names in order (because we have guarantees on the format of `labels`).

File: cleanlab/dataset.py (hunk 16 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:149]: added

File: cleanlab/dataset.py (hunk 10 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:132]: good idea. both edits made.

**Comment by cgnorthcutt** [comment]
IMPORTANT - this merge should include the commit history (do not squash, either merge directly or rebase and merge).

File: cleanlab/dataset.py (hunk 19 above)
Comment (jwmueller) [reply cleanlab/dataset.py:22]: I was suggesting either list or dict input formats could be supported for this function (i.e. both types work).  When I use cleanlab, I personally prefer to track the mapping between class-names and ints as a dict in my own code before running cleanlab commands, because dict is much more interpretable when I print it.

Comment (cgnorthcutt) [reply cleanlab/dataset.py:22]: Went ahead and implemented this for 2.0 🎉

File: cleanlab/dataset.py (hunk 12 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:134]: ended up completely refactoring and removing this method entirely and wrapping its functionality into a single method `rank_classes_by_label_quality`

File: cleanlab/dataset.py (hunk 8 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:61]: i added very thorough docstrings for the Returns section that details all scores in depth

File: cleanlab/dataset.py (hunk 5 above)
Comment (cgnorthcutt) [reply cleanlab/dataset.py:65]: I implemented conditional probability scores for the class level functionality and added thorough explanations in all Returns sections

**Comment by cgnorthcutt** [comment]
## Module completely rewritten.

* Merged the two label quality and label noise methods.
* report conditional matrix scores (instead of joint) for class methods
* All methods that previously returned multiple objects now work with and return pandas dataframes.
* Added support for string classes
* Added a health_summary function that prints everything about your dataset health in one line of code.
* completely rewrote the docstrings.
* all methods take in all input types.
* all methods have been renamed.
@anishathalye @jwmueller

**Comment by cgnorthcutt** [comment]
Example output of all methods (below)
image

Example output of health summary in one line of code for CIFAR-100:
image

File: cleanlab/dataset.py [hunk 20]
```diff
@@ -0,0 +1,416 @@
...
+            multi_label=multi_label,
+        )
+    if num_examples is None:
+        num_examples = _get_num_examples(labels=labels)
+    joint_trace = joint.trace()
+    if verbose:
+        num_issues = (num_examples * (1 - joint_trace)).round().astype(int)
+        print(
+            f" * Overall, about {1 - joint_trace:.0%} ({num_issues:,} of the {num_examples:,}) "
+            f"labels in your dataset have issues.\n"
```
cgnorthcutt, line 301: Changed to `labels in your dataset **may potentially have** issues`

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+# cleanlab is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
+
+
+"""Dataset Module
```
jwmueller, line 18: [Nit] None of the other modules descriptions start like this: 
Dataset Module

Cf: https://docs.cleanlab.ai/master/cleanlab/rank.html

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+Supports dataset-level and class-level automated quality, including finding which classes should
```
jwmueller, line 19: could be clarified: "Supports dataset-level and class-level automated quality, including ..."

Example alternative: "Provides dataset-level and class-level overviews of the label issues in your data, including ..."

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+    df_class_label_quality = rank_classes_by_label_quality(
+        labels=labels,
+        pred_probs=pred_probs,
+        class_names=class_names,
+        num_examples=num_examples,
+        joint=joint,
+        confident_joint=confident_joint,
+        multi_label=multi_label,
+    )
+    print("Overall Class Quality and Noise across your dataset (below)")
```
jwmueller, line 363: maybe instead of "(below)" could say something more descriptive like:
"(sorted below with lowest quality labels near the top)"

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+    print("-" * 60, "\n", flush=True)
+    print(df_class_label_quality)
```
jwmueller, line 365: should we just print `df.head()` instead? The topmost ones seem most important to review

jwmueller: this applies to all dataframe prints throughout this module

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+        "*" * 56
+        + "\n"
+        + "Generating a Cleanlab Dataset Health Summary\n"
+        + f" for your dataset with {num_examples:,} examples and {len(joint):,} classes.\n"
+        + "Note, Cleanlab is not a medical doctor... yet.\n"
+        + "*" * 56
+        + "\n",
+    )
+
+    df_class_label_quality = rank_classes_by_label_quality(
```
jwmueller, line 354: Some of the DF column names could be more clear if using terms from our Studio report-template instead.

Eg:

"Overall Label Issues" -> "Num. Label Issue"

jwmueller: Could add informative column from our Studio report template:
"% of Examples with Label Issue" 
which is supposed to = 100 * (num-examples with this given label and flagged as having issue) / (num-examples with this given label)

jwmueller: Also could add informative column from our Studio report template:
"Number of examples given this label"
which helps user trade-off between prioritizing classes that are very noisy but rare overall vs classes that are somewhat noisy but very common overall.

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+            "Inverse Label Issues": (true_label_noise * num_examples).round().astype(int),
+            "Overall Label Noise Score": given_conditional_noise,  # p(y!=k | s=k)
+            "Inverse Label Noise Score": true_conditional_noise,  # p(s!=k | y=k)
+            # Below could equivalently be computed as: joint.diagonal() / joint.sum(axis=1)
+            "Overall Label Quality Score": 1 - given_conditional_noise,  # p(y=k | s=k)
+        }
+    )
+    if class_names is not None:
+        df.insert(loc=0, column="Class Name", value=class_names)
+    return df.sort_values(by="Overall Label Noise Score", ascending=False)
```
jwmueller, line 105: [Nit] future development will be less bug-prone if you define global variable:

OVERALL_LABEL_NOISE_SCORE = "Overall Label Noise Score"
and have all code refer to OVERALL_LABEL_NOISE_SCORE instead throughout.  This would ideally be done for all strings referenced in multiple places of the code.

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+        "Overall Label Issues", "Inverse Label Noise Score", "Overall Label Quality Score" and a
+        description of each below. Noise scores are between 0 and 1, where 0 implies no label issues
+        in the class. The "Overall Label Quality Score" is also between 0 and 1 where 1 implies
+        perfect quality.
+        Columns:
+        * "Class Index"
+            The index of the class in 0, 1, ..., num_classes - 1.
+        * Overall Label Issues
+            count(given_label = k, true_label != k)
+            Estimated number of label issues in the class (usually the most accurate method).
```
jwmueller, line 63: unclear what this means: "(usually the most accurate method)"

jwmueller: Consider rename: "Overall Label Issues" -> "Number of Label Issues" or "Num. Label Issues"

jwmueller: "number of label issues in the class" could be clarified more:

"number of label issues among the examples whose given label is this class"

jwmueller: [Nit] I believe the way you have formatted docstring may not render well in the docs. May be better to format it like this:

* Overall Label Issues
            Estimated number of label issues in the class (usually the most accurate method), i.e.:  count(given_label = k, true_label != k)
            
Alternatively if you want count(...) to be on separate line, might need to insert additional line-break.

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+        * Inverse Label Issues
```
jwmueller, line 64: Consider rename of "Inverse Label Issues"  to something more descriptive like: "Number of Missed Instances"

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+            count(given_label != k, true_label = k)
+            Estimated number of times this class occurs in other classes in the dataset.
```
jwmueller, line 66: "Estimated number of times this class occurs in other classes in the dataset." could be more clear as something like:
"Estimated number of examples for which this class is more appropriate label than their given label."

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+        * Overall Label Noise Score
+            prob(true_label != k | given_label = k)
+            Estimated proportion of label issues in the class.
+        * Inverse Label Noise Score
+            prob(given_label != k | true_label = k)
+            Estimated proportion of times this class label occurs in other classes in the dataset.
+        * Overall Label Quality Score
+            p(given_label = k | true_label = k)
+            How accurate the labels are in each class as a conditional prob p(given=k | true=k).
```
jwmueller, line 75: Do we really need all 3 of these?  I found it a little overwhelming to see so many scores whose definition I didn't know as a user.

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+        By default, the DataFrame is ordered by "Overall Label Noise Score" descending.
```
jwmueller, line 76: Shouldn't the whole cleanlab package stick with the convention:
 lower scores -> more serious problem that warrants attention
 like we did for our current (per-example) label-quality scores?
I think we need to stick with one convention throughout, otherwise user has to remember what higher/lower means for every score and we'll soon have tons of scores.

cleanlab/dataset.py
jwmueller: At least "Overall Label Quality Score" and "Overall Label Noise Score" convey the same information AFAICT. Would just keep "Overall Label Quality Score" to stick with same convention as rest of the package.

jwmueller: "Inverse Label Noise Score" is derivable from "Inverse Label Issues" and the other columns. So want to ask ourself, is precomputing this score for the user worth potentially overwhelming them a little?  I felt a little  overwhelmed by seeing so many scores.

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+        )
+    if num_examples is None:
+        num_examples = _get_num_examples(labels=labels)
+    joint_trace = joint.trace()
+    if verbose:
+        num_issues = (num_examples * (1 - joint_trace)).round().astype(int)
+        print(
+            f" * Overall, about {1 - joint_trace:.0%} ({num_issues:,} of the {num_examples:,}) "
+            f"labels in your dataset have issues.\n"
+            f" ** The overall label health score for this dataset is: {joint_trace:.2f}."
```
jwmueller, line 302: can be clearer if you add: 
XXX (scores range between 0-1, higher is better)

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+):
+    """Returns the classes that are often confused by machine learning model or data labelers.
+    Consider merging the top pairs of classes returned by this method each into a single class.
+    If the dataset is labeled by human annotators, consider clearly defining the
+    difference between the classes prior to have annotators label the data.
+
+    This method provides two scores in the pandas Data Frame that is returned:
+    * "Num Overlapping Examples" - The number of examples where the two classes overlap
+    * "Joint Probability" - "Num Overlapping Examples" / total number of examples in the dataset.
+    interchangeable and returns a dataframe with the classe and the joint probability score
```
jwmueller, line 127: typos in text and formatting issue

cleanlab/dataset.py
jwmueller: Finally anytime the nebulous term "Score" is used, we probably want to stick with convention throughout that lower score -> more serious problem that warrants user's attention.  It is major QOL improvement for user if they don't have to remember which scores are higher = better vs lower = better.

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+    multi_label=False,
+):
+    """Returns the classes that are often confused by machine learning model or data labelers.
+    Consider merging the top pairs of classes returned by this method each into a single class.
+    If the dataset is labeled by human annotators, consider clearly defining the
+    difference between the classes prior to have annotators label the data.
+
+    This method provides two scores in the pandas Data Frame that is returned:
+    * "Num Overlapping Examples" - The number of examples where the two classes overlap
+    * "Joint Probability" - "Num Overlapping Examples" / total number of examples in the dataset.
```
jwmueller, line 126: I like the score "joint probability" because it's very interpretable. However one issue is this violates the convention of our package that lower scores => more serious problem. 

If we want to be consistent with this convention (which I advise given we plan to have tons of data-quality scores in the package), then alternate option is to call this "Overlap Score" and in the docstring define it as 1 - joint probability, i.e. = 1 - (Number of overlapping examples) / (Total number of examples in the dataset)

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+        errors considered is based on the number of labels, not the number of examples. So, the
+        calibrated confident_joint will sum to the number of total labels.
+        The multi-label setting supports classification tasks where an example has 1 or more labels.
+        Example of a multi-labeled `labels` input: [[0,1], [1], [0,2], [0,1,2], [0], [1], ...]
+
+    Returns
+    -------
+    pd.DataFrame
+        A Pandas DataFrame with columns "Class Index A", "Class Index B",
+        "Num Overlapping Examples", "Joint Probability" and a description of each below.
```
jwmueller, line 200: Add: "Each row corresponds to a pair of classes."

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+    Parameters
+    ----------
+    For parameter info, see the docstring of `dataset.find_overlapping_classes`
+
+    Returns
+    -------
+    pd.DataFrame
+        A Pandas DataFrame with cols "Class Index", "Overall Label Issues", "Inverse Label Issues",
+        "Overall Label Issues", "Inverse Label Noise Score", "Overall Label Quality Score" and a
+        description of each below. Noise scores are between 0 and 1, where 0 implies no label issues
```
jwmueller, line 55: Add after first sentence: "Each row corresponds to one class label."

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+    labels=None,
+    pred_probs=None,
+    *,
+    num_examples=None,
+    joint=None,
+    confident_joint=None,
+    multi_label=False,
+    verbose=True,
+):
+    """Returns a single score/metric between 0 and 1 for the quality of all labels in a dataset.
```
jwmueller, line 270: [Nit] "quality" -> "overall quality"

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+    Intuitively, the score is the average label consistency across all classes in the
```
jwmueller, line 271: "correctness of the given labels" instead of "label consistency"?

cleanlab/dataset.py
jwmueller: [Nit] "may potentially" can be shortened to "potentially"

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+    pred_probs=None,
+    *,
+    asymmetric=False,
+    class_names=None,
+    num_examples=None,
+    joint=None,
+    confident_joint=None,
+    multi_label=False,
+):
+    """Prints a healthy summary of your datasets including results for powerful statistics like:
```
jwmueller, line 318: "healthy summary of your datasets including results for powerful statistics like"
->
"summary of your dataset's overall label health, including:"

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+    * the classes with the most and least label issues
+    * classes that overlap and could potentially be merged
+    * overall data label quality health score statistics for your dataset
```
jwmueller, line 321: [Nit] could delete final bullet since it's implied by "overall health summary"

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+
+    Parameters
+    ----------
+    For parameter info, see the docstring of `dataset.find_overlapping_classes`
```
jwmueller, line 325: I'd imagine `health_summary()` is the method we envision most users invoking from this module (and just getting results from find_overlapping_classes from the internal call inside health_summary).

So consider moving the docstrings for parameters to this method, and then having the Docstring for find_overlapping_classes say: See docstring of `health_score` instead.

cleanlab/dataset.py
jwmueller: [Nit] There are a couple occurrences of "datasets" in this file. Would replace them with "dataset"

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+        num_examples=num_examples,
+        joint=joint,
+        confident_joint=confident_joint,
+        multi_label=multi_label,
+    )
+    print("Overall Class Quality and Noise across your dataset (below)")
+    print("-" * 60, "\n", flush=True)
+    print(df_class_label_quality)
+
+    df_class_label_quality = find_overlapping_classes(
```
jwmueller, line 367: [Nit] confusing variable name choice, would name: `df_overlapping_classes` instead, especially if we want to have option to return both of these.

cleanlab/dataset.py
```diff
@@ -0,0 +1,416 @@
...
+        + "\n"
+        + "-" * 83
+        + "\n",
+        flush=True,
+    )
+    print(df_class_label_quality)
+    print()
+    health_score = overall_label_health_score(labels=labels, pred_probs=pred_probs, verbose=True)
+    print("\nGenerated with <3 from Cleanlab.")
+    return health_score
```
jwmueller, line 388: `health_summary()` seems to already compute the results of every other function in this module, but they're being tossed. Any way to return these to user as well?

This speaks to the need for future `Dataset` class to store intermediate results. 

For now, this method could have boolean flag argument: `return_value :  List[str]` with default value: `["health_score", "class_label_quality", "class_overlap"]` (and another option for this list could be `"joint"`). Then 
based on `return_value`, this method returns a dict whose keys = the values in `return_value` and values = the corresponding scores / dataframes / arrays.

jwmueller (changes requested):
New output format looks great!  I mostly suggest some minor API refactors, and major changes in naming conventions to improve usability/clarity.

JohnsonKuan:
> Example output of all methods for Google Quickdraw dataset (below): Screen Shot 2022-04-10 at 7 31 09 AM
> 
> Example output of health summary in one line of code for CIFAR-100: image

Output looks awesome.

It may be helpful to explain in the docstring why the estimated joint prob of `(A, B)` is not the same as `(B, A)` like in your quickdraw example where joint prob of `(birthday cake, cake)` differed from `(cake, birthday cake)`. I know why they differ (after initial read of the confident learning paper) but I'm guessing most new users will probably not know.

cleanlab/dataset.py
jwmueller: Label Quality Score can be very clearly defined as follows:

How accurate the labels tend to be overall for a given class, i.e. the estimated probability that the true label should be `k` averaged across all examples whose given label = `k`.

cleanlab/dataset.py
anishathalye: Removed "Score", so now it's just "Overall Label Noise"

cleanlab/dataset.py
anishathalye: Great idea, will do later

tests/test_dataset.py
```diff
@@ -0,0 +1,426 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
+
+import requests
+import pytest
+import io
+import numpy as np
+from cleanlab import dataset
+
```
jwmueller, line 22: May want test for edge-case with only 2 classes as well?  I could envision potential bugs / formatting issues in that edge-case

cleanlab/dataset.py
```diff
@@ -0,0 +1,411 @@
...
+        + "\n",
+        flush=True,
+    )
+    print(df_overlapping_classes)
+    print()
+    health_score = overall_label_health_score(labels=labels, pred_probs=pred_probs, verbose=True)
+    print("\nGenerated with <3 from Cleanlab.")
+    return {
+        "health_score": health_score,
+        "joint": estimate_joint,
```
jwmueller, line 380: BUG:  This "joint" key was a function for me in my test run:

<function estimate_joint at 0x1236dc4c0>

Does not seem right

cleanlab/dataset.py
```diff
@@ -0,0 +1,411 @@
...
+    num_overlapping = (df["Joint Probability"] * num_examples).round().astype(int)
+    df.insert(loc=2, column="Num Overlapping Examples", value=num_overlapping)
+    if class_names is not None:
+        df.insert(
+            loc=0, column="Class Name A", value=df["Class Index A"].apply(lambda x: class_names[x])
+        )
+        df.insert(
+            loc=1, column="Class Name B", value=df["Class Index B"].apply(lambda x: class_names[x])
+        )
+    return df.sort_values(by="Joint Probability", ascending=False)
```
jwmueller, line 246: May want to be careful about the dataframe's index (ie. reset it).
Currently looks arbitrary in my run:

```
Class Overlap. In some cases, you may want to merge classes in the top rows (below)
-----------------------------------------------------------------------------------

  Class Name A Class Name B  Class Index A  Class Index B  Num Overlapping Examples  Joint Probability
7            C            D              2              3                         4               0.04
9            D            E              3              4                         3               0.03
3            A            E              0              4                         1               0.01
6            B            E              1              4                         1               0.01
0            A            B              0              1                         0               0.00
1            A            C              0              2                         0               0.00
2            A            D              0              3                         0               0.00
4            B            C              1              2                         0               0.00
5            B            D              1              3                         0               0.00
8            C            E              2              4                         0               0.00
```

jwmueller: The index for classes_by_label_quality returned DataFrame seems to make sense though AFAICT, it's simply the class index.

tests/test_dataset.py
cgnorthcutt: yup included in current tests (imdb is 2 classes)

cleanlab/dataset.py
cgnorthcutt: i think you fixed this. thanks

cgnorthcutt:
Addressed all comments and added thorough testing on real datasets with 2, 10, and more classes. Verified that methods return correct values in some cases.

PR #184, opened 2022-04-07
Move rest of example_models/ -> experimental/

jwmueller:
@cgnorthcutt mentioned these may have some value to certain users.
Especially thought it was odd to separate the cifar_cnn.py from coteaching.py as these are tied together. These seem fine in experimental/ since it has a clear disclaimer.


anishathalye (approved):
LGTM. I deleted the test that wasn't being run.

PR #185, opened 2022-04-07
Add more ergonomic method to skip notebooks

anishathalye:
This way, developers don't have to edit a file and remember to change it back before submitting a pull request: that seems error-prone.


jwmueller (approved):
Love it!

PR #186, opened 2022-04-07
Fix typo in test

anishathalye:
This test had a typo in it ("pred_neq_given" rather than "predicted_neq_given"), but the way the test was written, it didn't
fail, because GridSearchCV raises a warning if some fits fail. This patch fixes the typo and adds the assertion that no warnings are raised during the grid search.

### DIFF
--- a/tests/test_classification.py
+++ b/tests/test_classification.py
@@ -15,6 +15,7 @@
 # along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
 
 from copy import deepcopy
+import warnings
 from sklearn.linear_model import LogisticRegression
 from sklearn.base import BaseEstimator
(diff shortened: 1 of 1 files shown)


jwmueller (approved):
Great catch!

JohnsonKuan:
Good catch. When was `predicted_neq_given` introduced as an option for `filter_by`? I remember the original test for GridSearchCV did not include this option.

anishathalye:
It was introduced in the 8f9f3f53801c1100a25b458255e134877844e9d0 mega PR. The bad test option was added to the test in d3eb08e75a9a0319bf1f9e13c9ddde86d6cd50f1.

JohnsonKuan:
Oh yes, I remember now. Thanks!

PR #187, opened 2022-04-07
Add link checking for all Markdown files

anishathalye:
Example failing run: https://github.com/anishathalye/cleanlab/runs/5870066596?check_suite_focus=true
Example passing run: https://github.com/anishathalye/cleanlab/actions/runs/2109524114


jwmueller (approved):
Awesome to have!

PR #188, opened 2022-04-07
Add link checking for compiled docs

anishathalye:
Example failing run: https://github.com/anishathalye/cleanlab/runs/5870279217?check_suite_focus=true (reST parsed this link incorrectly)
Example passing run: https://github.com/anishathalye/cleanlab/runs/5870656641?check_suite_focus=true


jwmueller (approved):
Super useful!

PR #189, opened 2022-04-07
Docs cleanup

anishathalye:
This PR revises all public documentation (quickstart, tutorials, and docstrings). It also documents our documentation conventions in `DEVELOPMENT.md`.

This PR **does not** revise the documentation in `cleanlab.internal` or `cleanlab.experimental`. We can do that later.

**If you're pressing the merge button, please create a merge commit (or rebase and merge) to preserve the history here. Or just approve and let me merge myself.**


docs/source/index.rst
```diff
@@ -66,9 +66,9 @@ When the ``.fit()`` method is called, it automatically removes any examples iden
    from sklearn.linear_model import LogisticRegression
    from cleanlab.classification import CleanLearning
 
-   clf = LogisticRegression() # any classifier implementing the sklearn API
+   clf = LogisticRegression() # Here we've used sklearn's Logistic Regression model, but this can be any classifier that implements sklearn's API.
    cl = CleanLearning(clf=clf)
-   cl.fit(X=X, labels=labels)
+   cl.fit(X=X, labels=y)
```
cgnorthcutt, line 71: I may be out of context here, but I had changed `labels` to `y` universally, even in examples, everywhere. Any reason to make it `y` here?

anishathalye: Good catch. This is actually wrong in `master` right now, and I fixed it in one commit in this PR, and then reverted it back to master accidentally. Fixed.

jwmueller:
BTW in case you're able to handle in this PR (I don't want to cause merge conflict):  

the docstring for `cleanlab.filter.find_label_issues` doesn't even say what the method does :) 

The first line of its docstring should be something like:

Identifies potential label errors in the dataset via Confident Learning

anishathalye:
Ready for review. Again, this version history is useful, so if you're pressing the merge button, please create a merge commit (or rebase and merge) to preserve the history here. Or just approve and let me merge myself.

jwmueller (approved):
LGTM overall. 
I think you should just merge it, and we can continually revise any minor issues we spot in the docs. Docs improvement will be an ongoing continual process IMO and the docs will never be 100% immaculate, which is ok.

PR #190, opened 2022-04-08
Minor improvements to count docstring

cgnorthcutt:
## changes to docstrings
* Add params to joint.
* link calibrate to joint.
* make order of params match signature
* add type on docstring return


anishathalye (approved):
LGTM, will do cleanup in my mega docs PR.

PR #195, opened 2022-04-08
Remove import of internal package

anishathalye:
This import made `cleanlab.internal.util` available as `cleanlab.util`, which we don't want.


jwmueller (approved):
Good catch!

PR #197, opened 2022-04-09
Remove deprecated functions

anishathalye:
No reason to have any deprecated functions in a backwards-incompatible release like 2.0.


jwmueller (approved):
Good call

PR #199, opened 2022-04-12
Returns DataFrame type from CleanLearning functions

jwmueller:
CleanLearning.fit() can take in intermediate computation from either:
CleanLearning.find_label_issues() -- now a DataFrame
or filter.find_label_issues() -- a 1D np.array which is Boolean mask or integer indices if return_indices_ranked_by was specified.

This PR also fixes bug in entropy() and in get_confidence_weighted_entropy_for_each_label() related to potential 0s in logarithms.

Note: I have not updated unit tests (so currently tests will fail). I'll update the unit tests after getting feedback on the APIs/code.

Many possible workflows (updated):

```
from sklearn.ensemble import RandomForestClassifier

clf = RandomForestClassifier()
pred_probs = get_cross_validated_pred_probs(data, clf)

cl = CleanLearning(clf=RandomForestClassifier())

df = cl.find_label_issues(data, labels)
```

```
>>> df
    is_label_issue  label_quality  given_label  predicted_label
0             True       0.393853            2                1
1             True       0.322861            4                3
2             True       0.345246            4                3
3            False       0.850860            0                0
4            False       0.855554            0                0
..             ...            ...          ...              ...
95           False       0.500262            4                4
96           False       0.787859            0                0
97           False       0.706194            4                4
98           False       0.825338            4                4
99           False       0.970152            4                4
```

```
_ = cl.fit(data, labels)  # returns self 
```

```
>>> cl.get_label_issues()
    is_label_issue  label_quality  given_label  predicted_label  sample_weight
0             True       0.287438            2                1            0.0
1             True       0.468316            4                3            0.0
2             True       0.374178            4                3            0.0
3            False       0.949294            0                0            1.0
4            False       0.808242            0                0            1.0
..             ...            ...          ...              ...            ...
95            True       0.424141            4                3            0.0
96           False       0.869043            0                0            1.0
97           False       0.699087            4                4            1.0
98           False       0.895582            4                4            1.0
99           False       0.970152            4                4            1.0
```

```
issue_mask = find_label_issues(labels, pred_probs)
_ = cl.fit(data, labels, label_issues=issue_mask)
```

```
>>> cl.get_label_issues()
    is_label_issue  sample_weight
0             True            0.0
1             True            0.0
2             True            0.0
3            False            1.0
4            False            1.0
..             ...            ...
95           False            1.0
96           False            1.0
97           False            1.0
98           False            1.0
99           False            1.0
```

```
issue_inds = find_label_issues(y, pred_probs, return_indices_ranked_by='normalized_margin')
_ = cl.fit(data, labels,  label_issues=issue_inds)
```

```
>>> cl.get_label_issues()
    is_label_issue  sample_weight
0             True            0.0
1             True            0.0
2             True            0.0
3            False            1.0
4            False            1.0
..             ...            ...
95           False            1.0
96           False            1.0
97           False            1.0
98           False            1.0
99           False            1.0
```

```
>>> cl.save_space()
Deleted non-sklearn attributes such as label_issues_df to save space.

>>> cl.get_label_issues()
UserWarning: The label issues have not yet been computed. Run `self.find_label_issues()` or `self.fit()` first.
```

### DIFF
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -363,8 +363,7 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
           `self.label_issues_df`: a pd.DataFrame accessible via
           :py:meth:`get_label_issues
           <cleanlab.classification.CleanLearning.get_label_issues>`
-          of similar format as the one returned by:
-          :py:meth:`CleanLearning.find_label_issues
+          of similar format as the one returned by: :py:meth:`CleanLearning.find_label_issues
             <cleanlab.classification.CleanLearning.find_label_issues>`.
           See documentation of :py:meth:`CleanLearning.find_label_issues
             <cleanlab.classification.CleanLearning.find_label_issues>`
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -340,8 +340,7 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
           pruned from the data before training the final `clf` model.
 
           Caution: If you provide `label_issues` without having previously called
-          :py:meth:`self.find_label_issues
-          <cleanlab.classification.CleanLearning.find_label_issues>`,
+          :py:meth:`self.find_label_issues<cleanlab.classification.CleanLearning.find_label_issues>`,
           e.g. as a np.array, then some functionality like training with sample weights may be disabled.
 
         clf_kwargs : dict, optional
@@ -363,10 +362,8 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
           `self.label_issues_df`: a pd.DataFrame accessible via
           :py:meth:`get_label_issues
           <cleanlab.classification.CleanLearning.get_label_issues>`
-          of similar format as the one returned by: :py:meth:`CleanLearning.find_label_issues
-            <cleanlab.classification.CleanLearning.find_label_issues>`.
-          See documentation of :py:meth:`CleanLearning.find_label_issues
-            <cleanlab.classification.CleanLearning.find_label_issues>`
+          of similar format as the one returned by: :py:meth:`CleanLearning.find_label_issues<cleanlab.classification.CleanLearning.find_label_issues>`.
+          See documentation of :py:meth:`CleanLearning.find_label_issues<cleanlab.classification.CleanLearning.find_label_issues>`
           for column descriptions.
           After calling ``self.fit()``, `self.label_issues_df` may also contain an extra column:
 
@@ -590,24 +587,12 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
             Each row represents an example from our dataset and
             `label_issues_df` may contain the following columns:
 
-            * *is_label_issue*: boolean mask for the entire dataset where ``True`` represents a
-            label issue and ``False`` represents an example that is accurately
-            labeled with high confidence.
-            This column is equivalent to `label_issues_mask` output from
-            :py:func:`filter.find_label_issues<cleanlab.filter.find_label_issues>`.
-            * *label_quality*: Numeric score that measures the quality of each label
-            (how likely it is to be correct, with lower scores indicating potentially erroneous labels).
-            * *given_label*: Integer indices corresponding to the class label originally given for this example
-            (same as `labels` input). Included here for ease of comparison against `clf` predictions,
-             only present if "predicted_label" column is present.
-            * *predicted_label*: Integer indices corresponding to the class predicted by trained `clf` model
-            Only present if ``pred_probs`` were provided as input or computed during label-issue-finding.
-            * *sample_weight*: Numeric values used to weight examples during
-            the final training of `clf` in :py:meth:`CleanLearning.fit()<cleanlab.classification.CleanLearning.fit>`.
-            This column not be present after `self.find_label_issues()` but may be added
-            after call to :py:meth:`CleanLearning.fit()<cleanlab.classification.CleanLearning.fit>`.
-            For more precise definition of sample weights, see documentation of
-            :py:meth:`CleanLearning.fit()<cleanlab.classification.CleanLearning.fit>`.
+            * *is_label_issue*: boolean mask for the entire dataset where ``True`` represents a label issue and ``False`` represents an example that is accurately labeled with high confidence. This column is equivalent to `label_issues_mask` output from :py:func:`filter.find_label_issues<cleanlab.filter.find_label_issues>`.
+            * *label_quality*: Numeric score that measures the quality of each label (how likely it is to be correct, with lower scores indicating potentially erroneous labels).
+            * *given_label*: Integer indices corresponding to the class label originally given for this example (same as `labels` input). Included here for ease of comparison against `clf` predictions, only present if "predicted_label" column is present.
+            * *predicted_label*: Integer indices corresponding to the class predicted by trained `clf` model. Only present if ``pred_probs`` were provided as input or computed during label-issue-finding.
+            * *sample_weight*: Numeric values used to weight examples during the final training of `clf` in :py:meth:`CleanLearning.fit()<cleanlab.classification.CleanLearning.fit>`. This column not be present after `self.find_label_issues()` but may be added
+            after call to :py:meth:`CleanLearning.fit()<cleanlab.classification.CleanLearning.fit>`. For more precise definition of sample weights, see documentation of :py:meth:`CleanLearning.fit()<cleanlab.classification.CleanLearning.fit>`
         """
 
         if self.label_issues_df is not None and self.verbose and not save_space:
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -128,6 +128,7 @@ from cleanlab.count import (
     estimate_py_and_noise_matrices_from_probabilities,
     estimate_cv_predicted_probabilities,
     estimate_latent,
+    compute_confident_joint,
 )
 from cleanlab.internal.latent_algebra import (
     compute_py_inv_noise_matrix,
@@ -691,6 +692,13 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
                     thresholds=thresholds,
                     converge_latent_estimates=self.converge_latent_estimates,
                 )
+        # If needed, compute the confident_joint (e.g. occurs if noise_matrix was given)
+        if self.confident_joint is None:
+            self.confident_joint = compute_confident_joint(
+                labels=labels,
+                pred_probs=pred_probs,
+                thresholds=thresholds,
+            )
         # If needed, compute P(label=k|x), denoted pred_probs (the predicted probabilities)
         if pred_probs is None:
             if self.verbose:
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -326,7 +326,9 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
           Specifies the label issues for each example in dataset.
           If pd.DataFrame, must be formatted as the one returned by:
           :py:meth:`CleanLearning.find_label_issues
-          <cleanlab.classification.CleanLearning.find_label_issues>` or by this function.
+          <cleanlab.classification.CleanLearning.find_label_issues>` or
+          :py:meth:`CleanLearning.get_label_issues
+          <cleanlab.classification.CleanLearning.get_label_issues>`.
           If np.array, must contain either boolean label_issues_mask as output by:
           default :py:func:`filter.find_label_issues <cleanlab.filter.find_label_issues>`,
           or integer indices as output by
@@ -355,25 +357,39 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
         -------
         self
             Fitted estimator that has all the same methods as any sklearn estimator.
+
+
         After calling `fit()` this estimator also stores a few extra useful attributes, in particular
+        `self.label_issues_df`: a pd.DataFrame accessible via `self.get_label_issues()`
+        of similar format as the one returned by:
+        :py:meth:`CleanLearning.find_label_issues
+          <cleanlab.classification.CleanLearning.find_label_issues>`
+
         `self.label_issues_df`: a pd.DataFrame in which each row represents an example from our dataset.
         `self.label_issues_df` may contain the following columns:
 
-          - "is_label_issue": boolean where ``True`` represents a
-            label issue and ``False`` represents an example that is accurately
-            labeled with high confidence.
-            This column is equivalent to `label_issues_mask` output from `filter.find_label_issues()`.
-
-          - "label_quality": numeric score that measures the quality of each label
-            (how likely it is to be correct, with lower scores indicating potentially erroneous labels).
-
-          - "predicted_class": integer indices corresponding the class predicted by trained `clf` model
-            (only present if `pred_probs` were provided as input or computed during label-issue-finding).
-
-          - "sample_weight": numeric values corresponding to weights to assign each example
-            based on how trustworthy its label is.
-            These sample weights were used in the final model training in `CleanLearning.fit()`.
-            sample_weight column will only be present if sample weights were actually used.
+          * *is_label_issue*: boolean mask for the entire dataset where ``True`` represents a
+          label issue and ``False`` represents an example that is accurately
+          labeled with high confidence.
+          This column is equivalent to `label_issues_mask` output from `filter.find_label_issues()`.
+          * *label_quality*: Numeric score that measures the quality of each label
+          (how likely it is to be correct, with lower scores indicating potentially erroneous labels).
+          * *given_label*: Integer indices corresponding to the class label originally given for this example
+          (same as `labels` input). Included here for ease of comparison against `clf` predictions,
+           only present if "predicted_label" column is present.
+          * *predicted_label*: Integer indices corresponding to the class predicted by trained `clf` model
+          Only present if `pred_probs` were provided as input or computed during label-issue-finding.
+          * *sample_weight*: Numeric values that were used to weight examples during
+          the final training of `clf` in `CleanLearning.fit()`.
+          sample_weight column will only be present if sample weights were actually used.
+          These weights are assigned to each example based on the class it belongs to,
+          i.e. there are only num_classes unique sample_weight values.
+          The sample weight for an example belonging to class k is computed as 1 / p(given_label = k | true_label = k).
+          This sample_weight normalizes the loss to effectively trick `clf` into learning with the distribution
+          of the true labels by accounting for the noisy data pruned out prior to training on cleaned data.
+          In other words, examples with label issues were removed, so this weights the data proportionally
+          so that the classifier trains as if it had all the true labels,
+          not just the subset of cleaned data left after pruning out the label issues.
         """
 
         clf_final_kwargs = {**clf_kwargs, **clf_final_kwargs}
@@ -437,7 +453,9 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
             sample_weight_expanded[x_mask] = sample_weight
             # Store the sample weight for every example in the original, unfiltered dataset
             self.label_issues_df["sample_weight"] = sample_weight_expanded
-            self.sample_weight = self.label_issues_df["sample_weight"]
+            self.sample_weight = self.label_issues_df[
+                "sample_weight"
+            ]  # pointer to here to avoid duplication
             if self.verbose:
                 print("Fitting final model on the clean data ...")
 
@@ -559,7 +577,7 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
         :py:meth:`CleanLearning.fit()
         <cleanlab.classification.CleanLearning.fit>`
         and they share mostly the same parameters.
-        
+
         Parameters
         ----------
         save_space : bool, optional
@@ -719,10 +737,12 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
         predicted_labels = pred_probs.argmax(axis=1)
         label_issues_df["given_label"] = compress_int_array(labels, self.num_classes)
         label_issues_df["predicted_label"] = compress_int_array(predicted_labels, self.num_classes)
-        
+
         if not save_space:
             self.label_issues_df = label_issues_df
-            self.label_issues_mask = label_issues_mask
+            self.label_issues_mask = label_issues_df[
+                "is_label_issue"
+            ]  # pointer to here to avoid duplication
         elif self.verbose:
             print(  # pragma: no cover
                 "Not storing label_issues as attributes since save_space was specified."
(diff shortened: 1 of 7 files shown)


jwmueller:
Note: sample_weight column corresponds to what was being returned from CleanLearning.fit() before (except now padded with 0s in pruned examples). It contains floats and is not binary vector, just happens to look like that in this particular example.

cleanlab/classification.py
```diff
@@ -313,16 +321,17 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
 
-        label_issues_mask : np.array, optional
-          A boolean mask for the entire dataset such as those returned by:
+        label_issues : pd.DataFrame or np.array, optional
+          Specifies the label issues for each example in dataset.
+          If DataFrame, must be formatted as the one returned by:
           :py:meth:`CleanLearning.find_label_issues
-          <cleanlab.classification.CleanLearning.find_label_issues>` or
+          <cleanlab.classification.CleanLearning.find_label_issues>` or by this function.
+          If np.array, must contain boolean label_issues_mask as output by:
```
anishathalye, line 329: While we're at it, why not also support an array of integers, in case the user got ranked label issues back from `find_label_issues`?

cleanlab/classification.py
```diff
@@ -355,15 +383,41 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+                raise ValueError("DataFrame label_issues must contain column: 'is_label_issue'")
+            if len(label_issues) != len(labels):
+                raise ValueError("label_issues and labels must have same length")
+            self.label_issues_df = label_issues
+        elif isinstance(label_issues, np.ndarray):
+            if label_issues.shape != labels.shape:
+                raise ValueError(
+                    "If it is numpy.array, label_issues must have same shape as labels"
+                )
+            if not ((label_issues == 0) | (label_issues == 1)).all():
```
anishathalye, line 397: Why not check `label_issues.dtype` instead?

cleanlab/classification.py
```diff
@@ -379,17 +433,20 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
                 )
 
-            self.sample_weight = np.ones(np.shape(labels_cleaned))
+            sample_weight = np.ones(np.shape(labels_cleaned))
             for k in range(self.num_classes):
                 sample_weight_k = 1.0 / max(self.noise_matrix[k][k], 1e-3)  # clip sample weights
-                self.sample_weight[labels_cleaned == k] = sample_weight_k
+                sample_weight[labels_cleaned == k] = sample_weight_k
 
+            sample_weight_expanded = np.zeros(len(labels))  # pad pruned examples with zero weight
```
cgnorthcutt, line 438: i would move this logic into an accessor for `get_label_issues_df` (see comment below). You could just return self.clf here as is the normal convention and you can build up a dataframe as its own accessor method, you can still store predictions, given labels, label quality scores, and boolean mask in a `self.label_issues_df` (if you don't mind bloating the size of the instance), but i would move all other logic and columns into the accessor method and just add columns for `self.sample_weight`, etc.

cleanlab/classification.py
```diff
@@ -490,18 +547,21 @@ def find_label_issues(
...
-          label issue and ``False`` represents an example that is accurately
-          labeled with high confidence.
+        label_issues_df : pd.DataFrame
+          DataFrame with same format as the one returned by :py:meth:`CleanLearning.fit()
+        <cleanlab.classification.CleanLearning.fit>`.
+          See there for documentation regarding column definitions.
         """
 
-        if self.label_issues_mask is not None and self.verbose:
+        try_import_pandas()
```
cgnorthcutt: delete

cleanlab/classification.py
```diff
@@ -623,26 +683,33 @@ def find_label_issues(
...
         if self.verbose:
-            print(f"Identified {np.sum(self.label_issues_mask)} examples with label issues.")
+            print(f"Identified {np.sum(label_issues_mask)} examples with label issues.")
 
-        return self.label_issues_mask
+        return self.label_issues_df
 
     def get_label_issues(self):
-        """Accessor. Returns `label_issues_mask` if previously already computed."""
+        """Accessor. Returns `label_issues_df` if previously already computed."""
```
cgnorthcutt, line 699: Add Returns pd.DataFrame (short description) to the docstring.

just use this as the accessor i mention above (`get_label_issues_df`)... the current name `get_label_issues` is fine

cgnorthcutt (changes requested):
Great idea to return the dataframe for `CleanLearning.find_label_issues`

Main feedback:
* Maybe just return self.clf in fit(). That's standard for sklearn compatible classifiers.
* move the logic you added for building up the `self.label_issues_df` into the accessor `get_label_errors` (it doesn't belong in .fit()`
* I would keep self.sample_weight and self.label_errors_mask and related instance variables around. Dataframes can use pointers (so no space duplication by storing both) and these are independent of the label_issues_df, so they should be accessible on their own (easier to find them too)
* I'd add given labels as a column to the dataframe which does add space, but i think it makes it more useful when juxtaposed with the prediction column.

cleanlab/classification.py
jwmueller: updated to:

```
if not label_issues.dtype is np.dtype('bool'):
                raise ValueError(
                    "If it is numpy.array, label_issues.dtype must be 'bool'."
                )
```

cleanlab/internal/util.py
```diff
@@ -439,3 +439,12 @@ def print_joint_matrix(joint_matrix, round_places=2):
         short_title="p(labels,y)",
         round_places=round_places,
     )
+
+
+def try_import_pandas():
```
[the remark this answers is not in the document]
jwmueller: I can just update the Dataset module to also use this lazy import? 
That way v2.0 filter/rank can still be run without pandas if somebody really hates it?

cleanlab/classification.py
```diff
@@ -131,6 +131,8 @@ def score(self, X, y, sample_weight=None):
     compute_py_inv_noise_matrix,
     compute_noise_matrix_from_inverse,
 )
+from cleanlab.internal.util import try_import_pandas
```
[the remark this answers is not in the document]
jwmueller: I can just update the Dataset module to also use this lazy import?
That way v2.0 filter/rank can still be run without pandas if somebody really hates it?

cleanlab/classification.py
```diff
@@ -337,15 +346,34 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+
+          "label_quality": Numeric score that measures the quality of each label
+          (how likely it is to be correct, with lower scores indicating potentially erroneous labels).
+
+          "predicted_class": Integer indices corresponding the class predicted by trained `clf` model
+          (only present if `pred_probs` were provided as input or computed during label-issue-finding).
+
+          "sample_weight": Numeric values corresponding to weights to assign each example
+          based on how trustworthy its label is.
+          These sample weights were used in the final model training in `CleanLearning.fit()`.
```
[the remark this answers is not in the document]
jwmueller: I think it's a 'were'. The final model training has already happened by the time sample-weight is shown here.
The intention of providing it is if the user wants to train other variants of `clf` with the same sample-weights that were used to produce the `clf` produced via CleanLearning.fit()

cleanlab/classification.py
```diff
@@ -337,15 +346,34 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+          labeled with high confidence.
+          This column is equivalent to `label_issues_mask` output from `filter.find_label_issues()`.
+
+          "label_quality": Numeric score that measures the quality of each label
+          (how likely it is to be correct, with lower scores indicating potentially erroneous labels).
+
+          "predicted_class": Integer indices corresponding the class predicted by trained `clf` model
+          (only present if `pred_probs` were provided as input or computed during label-issue-finding).
+
+          "sample_weight": Numeric values corresponding to weights to assign each example
```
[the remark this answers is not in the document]
jwmueller: added

cleanlab/classification.py
```diff
@@ -337,15 +346,34 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+
+          "is_label_issue": boolean mask for the entire dataset where ``True`` represents a
+          label issue and ``False`` represents an example that is accurately
+          labeled with high confidence.
+          This column is equivalent to `label_issues_mask` output from `filter.find_label_issues()`.
+
+          "label_quality": Numeric score that measures the quality of each label
+          (how likely it is to be correct, with lower scores indicating potentially erroneous labels).
+
+          "predicted_class": Integer indices corresponding the class predicted by trained `clf` model
```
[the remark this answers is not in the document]
jwmueller: done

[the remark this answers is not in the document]
jwmueller: FYI given_label will only be present when predicted_label is present as a column (written in docstring). I don't see value as adding given_label when predicted_label is not there.

cleanlab/classification.py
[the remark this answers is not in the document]
jwmueller: updated "adjusts the loss proportionally" to "weights the data proportionally" since many classifiers don't explicitly  optimize a loss

cleanlab/classification.py
```diff
@@ -355,15 +383,41 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
                 clf_kwargs=clf_kwargs,
             )
         elif self.verbose:
-            print("Using provided label_issues_mask instead of finding label issues.")
+            print("Using provided label_issues instead of finding label issues.")
+
+        # label_issues always overwrites self.label_issues_df. Ensure it is properly formatted:
+        if isinstance(label_issues, pd.DataFrame):
+            if "is_label_issue" not in label_issues.columns:
+                raise ValueError("DataFrame label_issues must contain column: 'is_label_issue'")
```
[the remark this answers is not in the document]
jwmueller: added

cleanlab/classification.py
```diff
@@ -379,17 +433,20 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
-            self.sample_weight = np.ones(np.shape(labels_cleaned))
+            sample_weight = np.ones(np.shape(labels_cleaned))
             for k in range(self.num_classes):
                 sample_weight_k = 1.0 / max(self.noise_matrix[k][k], 1e-3)  # clip sample weights
-                self.sample_weight[labels_cleaned == k] = sample_weight_k
+                sample_weight[labels_cleaned == k] = sample_weight_k
 
+            sample_weight_expanded = np.zeros(len(labels))  # pad pruned examples with zero weight
+            sample_weight_expanded[x_mask] = sample_weight
+            self.label_issues_df["sample_weight"] = sample_weight_expanded
```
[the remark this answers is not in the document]
jwmueller: get_label_issues() already returns this DF currently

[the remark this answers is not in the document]
jwmueller: added instance variables back, I think user can just access them directly via `clf.sample_weight` and `clf.label_issues_mask` if they want them

cleanlab/classification.py
jwmueller: added

cleanlab/classification.py
```diff
@@ -398,7 +455,7 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
             # This is less accurate, but best we can do if no sample_weight.
             self.clf.fit(x_cleaned, labels_cleaned, **self.clf_final_kwargs)
 
-        return self.clf
```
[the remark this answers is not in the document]
jwmueller: Are you sure we shouldn't return `self` instead of `self.clf`?

Cf:  https://github.com/scikit-learn/scikit-learn/blob/baf828ca126bcb2c0ad813226963621cafe38adb/sklearn/ensemble/_forest.py#L493

and your LogReg link also returns `self`

cleanlab/classification.py
```diff
@@ -337,15 +346,34 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
 
         Returns
         -------
-        tuple
-          (`label_issues_mask`, `sample_weight`)
+        label_issues_df : pd.DataFrame
+          DataFrame in which each row represents an example from our dataset. May contain the following columns:
+
+          "is_label_issue": boolean mask for the entire dataset where ``True`` represents a
```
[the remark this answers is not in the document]
jwmueller: nice! done

cleanlab/classification.py
jwmueller: updated. also added new `self.save_space()` method to delete all the non-sklearn attributes to streamline deployment of the fitted Estimator.

cleanlab/classification.py
jwmueller: added support for that, with check to make sure they're all unique

jwmueller:
Addressed comments except lazy import. See new workflow/outputs above.

cleanlab/classification.py
```diff
@@ -355,10 +386,52 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
                 clf_kwargs=clf_kwargs,
             )
         elif self.verbose:
-            print("Using provided label_issues_mask instead of finding label issues.")
+            print("Using provided label_issues instead of finding label issues.")
 
-        # Always overwrites `self.label_issues_mask` and ensures same length as `X` and `labels`.
-        self.label_issues_mask = label_issues_mask
+        # label_issues always overwrites self.label_issues_df. Ensure it is properly formatted:
+        if isinstance(label_issues, pd.DataFrame):
```
cgnorthcutt, line 392: Maybe move all of the parameter checking logic to helper function if this code can only be used here or to utils if this code can be re-used across the package.

setup.py
```diff
@@ -89,6 +89,7 @@ def run(self):
         "scikit-learn>=0.18",
         "scipy>=1.1.0",
         "tqdm>=4.53.0",
+        "pandas>=1.0.0",
```
cgnorthcutt, line 92: this is already added here: https://github.com/cleanlab/cleanlab/blob/8470c72667629e09b3f14850448b9f947b28b68a/setup.py#L92

cleanlab/classification.py
```diff
@@ -623,26 +705,62 @@ def find_label_issues(
...
             pred_probs,
             **self.find_label_issues_kwargs,
         )
+        label_quality_scores = get_label_quality_scores(
+            labels, pred_probs, **self.label_quality_scores_kwargs
+        )
+        self.label_issues_df = pd.DataFrame(
+            {"is_label_issue": self.label_issues_mask, "label_quality": label_quality_scores}
+        )
+        if pred_probs is not None:
```
cgnorthcutt, line 719: is this supposed to be `if self.label_issues_df` is not None`?

cgnorthcutt: if pred_probs was None here, wouldn't the previous lines fail?

cgnorthcutt: how do you know that self.label_issues_df has been created yet if the workflow is
```python
# Find label issues in a dataset in one line of code.
label_issues_df = CleanLearning(model).find_label_issues(labels, pred_probs)
```

cleanlab/internal/util.py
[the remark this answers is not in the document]
jwmueller: removed lazy import

cleanlab/classification.py
```diff
@@ -379,17 +452,21 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+            sample_weight = np.ones(np.shape(labels_cleaned))
             for k in range(self.num_classes):
                 sample_weight_k = 1.0 / max(self.noise_matrix[k][k], 1e-3)  # clip sample weights
-                self.sample_weight[labels_cleaned == k] = sample_weight_k
+                sample_weight[labels_cleaned == k] = sample_weight_k
 
+            sample_weight_expanded = np.zeros(len(labels))  # pad pruned examples with zero weight
+            sample_weight_expanded[x_mask] = sample_weight
+            self.label_issues_df["sample_weight"] = sample_weight_expanded
+            self.sample_weight = self.label_issues_df["sample_weight"]
```
cgnorthcutt, line 463: So, you're changing the functionality here right? Seem okay to store the expanded version instead, but maybe add a comment here above "Store the sample weight for every example in the original, unfiltered dataset"

cleanlab/classification.py
cgnorthcutt: nice. i like it!

cleanlab/classification.py
```diff
@@ -251,7 +259,7 @@ def fit(
         thresholds=None,
         noise_matrix=None,
         inverse_noise_matrix=None,
-        label_issues_mask=None,
+        label_issues=None,
```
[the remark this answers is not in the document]
cgnorthcutt: got it. i'd comment to the right then `  # can be pandas.dataframe, np.array<bool>, or np.array<int>` and also update in docstring.

cleanlab/classification.py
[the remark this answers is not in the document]
cgnorthcutt: great catch

cleanlab/classification.py
```diff
@@ -490,18 +572,18 @@ def find_label_issues(
...
         Returns
         -------
-        label_issues_mask : np.array
-          A boolean mask for the entire dataset where ``True`` represents a
-          label issue and ``False`` represents an example that is accurately
-          labeled with high confidence.
+        label_issues_df : pd.DataFrame
+          DataFrame with same format as the one returned by :py:meth:`CleanLearning.fit()
+        <cleanlab.classification.CleanLearning.fit>`.
+          See there for documentation regarding column definitions.
```
cgnorthcutt, line 578: I think you sould move the return docstring for label_issues_df  here since THIS is where the df actually gets returned and just refer to it here in fit()?

cleanlab/classification.py
```diff
@@ -379,17 +452,21 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+            sample_weight_expanded[x_mask] = sample_weight
+            self.label_issues_df["sample_weight"] = sample_weight_expanded
+            self.sample_weight = self.label_issues_df["sample_weight"]
             if self.verbose:
                 print("Fitting final model on the clean data ...")
             self.clf.fit(
                 x_cleaned,
                 labels_cleaned,
-                sample_weight=self.sample_weight,
+                sample_weight=sample_weight,
```
cgnorthcutt, line 469: a comment in the code explaining that `sample_weight` is length of the pruned data after cleaning out label issues and `self.sample_weight` is length of the entire dataset, could be helpful for developers (they're named the same but they are different... pretty tricky stuff with out a comment)

cleanlab/classification.py
```diff
@@ -623,26 +705,62 @@ def find_label_issues(
...
         )
+        label_quality_scores = get_label_quality_scores(
+            labels, pred_probs, **self.label_quality_scores_kwargs
+        )
+        self.label_issues_df = pd.DataFrame(
+            {"is_label_issue": self.label_issues_mask, "label_quality": label_quality_scores}
+        )
+        if pred_probs is not None:
+            self.label_issues_df["given_label"] = labels
+            self.label_issues_df["predicted_label"] = pred_probs.argmax(axis=1)
```
cgnorthcutt, line 721: optional ---  if you don't want to store these is to just return a local version that has these two columns in it, without changing `self.label_issues_df`.

You don't lose any speed by doing this -- this function is not memoized at all and you recompute everything anyway when the function is called.

cgnorthcutt (changes requested):

cleanlab/classification.py
jwmueller: yes storing expanded weights is more useful IMO. added the comment.

cleanlab/classification.py
jwmueller: added helper class method _process_label_issues_arg

cleanlab/classification.py
jwmueller: added comments

cleanlab/classification.py
jwmueller: dont fully grok this comment, clarify more?

jwmueller:
Note all lines which do not pass codecov are defensive `raise ValueError` statements. Added a `raise ValueError` line to .coveragerc under `exclude_lines` which should stop codecov complaining about such lines but I don't think it takes effect until this PR is merged in.

cgnorthcutt (approved):

cleanlab/classification.py
```diff
@@ -557,6 +559,12 @@ def find_label_issues(
         <cleanlab.classification.CleanLearning.fit>`. See there for
```
cgnorthcutt, line 559: Move this into the Parameters block, just below the `save_space`

cgnorthcutt (changes requested):
super quick change

cleanlab/classification.py
cgnorthcutt: Again, I really think you should provide a spec for the df thats returned since this is the method to create that df and return it and fit doesnt even return it

cleanlab/classification.py
```diff
@@ -689,23 +697,38 @@ def find_label_issues(
...
-        self.label_issues_df["predicted_label"] = pred_probs.argmax(axis=1)
         if self.verbose:
-            print(f"Identified {np.sum(self.label_issues_mask)} examples with label issues.")
-
-        return self.label_issues_df
+            print(f"Identified {np.sum(label_issues_mask)} examples with label issues.")
+
+        predicted_labels = pred_probs.argmax(axis=1)
+        compressed_type = None
+        if self.num_classes < np.iinfo(np.dtype("int16")).max:
```
cgnorthcutt, line 716: This logic should go in utils. its fairly universal and we should use it to reduce space of ints everywhere.

cgnorthcutt (changes requested):
two other small changes

cleanlab/classification.py
jwmueller: where else should it be used?

cleanlab/classification.py
jwmueller: can you clarify "this"

jwmueller: addressed it partially, feel free to implement a better solution directly if you have one.

cleanlab/classification.py
jwmueller: moved to utils function. This function is now also being applied to `labels` as I benchmarked memory and realized the labels are being duplicated in order to create the dataframe, so may as well compress them.

jwmueller:
fixed docstrings formatting

cleanlab/classification.py
```diff
@@ -488,20 +559,26 @@ def find_label_issues(
...
-        if self.label_issues_mask is not None and self.verbose:
+        if self.label_issues_df is not None and self.verbose and not save_space:
             print(
-                "Overwriting previously identified label issues stored at self.label_issues_mask. "
+                "Overwriting previously identified label issues stored at self.label_issues_df. "
                 "`self.get_label_issues()` will now return the newly identified label issues. "
-                "If you already ran `self.find_label_issues()` and don't want to recompute, you"
-                "should pass the `label_issues_mask` in as a parameter to this function."
+                "If you already ran `self.find_label_issues()` and don't want to recompute, you "
+                "should pass the `label_issues_df` in as a parameter to this function."
```
cgnorthcutt, line 606: Bug? this method doesn't have `label_issues_df` as a parameter? How can you ask the user to pass it in?

cgnorthcutt (commented):
Potentially a bug or potentially an old print statement that is out of date. see comment.

cleanlab/classification.py
jwmueller: good catch, I moved part of the print statement elsewhere in latest commit.

cgnorthcutt (approved):
LGTM

PR #204, opened 2022-04-14
set verbose default false. fix order of printing

cgnorthcutt:
Very Minor changes.


anishathalye:
While we're at it, should we also set `dataset.py:267` to have `verbose=False` by default?

cgnorthcutt:
> While we're at it, should we also set `dataset.py:267` to have `verbose=False` by default?

I vote no because its only one line and its actually a very useful line (and also the point of the function is to get a summary, so it makes more sense to allow printing for a summary function versus a classifier training function.

However, if we add a lot more print statements, eventually we should set to False. Okay as is for now imo.

PR #212, opened 2022-04-15
bug fixes and jupyter notebook support added

cgnorthcutt:
This PR adds three things to ``dataset.health_summary``
* Support for displaying pandas dataframe in a jupyter notebook instead of printing it out, but if the user runs `health_summary` in a terminal with no jupyter notebook, it should still print to console.
* Fixed a bug that would cause health_summary to fail if `pred_probs` is not provided.
* Fixed the printing of the header at the top.

### DIFF
--- a/cleanlab/internal/util.py
+++ b/cleanlab/internal/util.py
@@ -458,13 +458,6 @@ def smart_display_dataframe(df):  # pragma: no cover
     try:
         from IPython.display import display
 
-        # no exception, let's determine if we are running inside a jupyter notebook
-        shell = get_ipython().__class__.__name__
-        if shell == "ZMQInteractiveShell":
-            display(df)  # Jupyter notebook or qtconsole
-        elif shell == "TerminalInteractiveShell":
-            print(df)  # terminal, not notebook
-        else:
-            print(df)  # something else, but not notebook
-    except:  # pragma: no cover
+        display(df)
+    except:
         print(df)
--- a/cleanlab/internal/util.py
+++ b/cleanlab/internal/util.py
@@ -453,7 +453,7 @@ def compress_int_array(int_array, num_possible_values):
     return int_array
 
 
-def smart_display_dataframe(df):
+def smart_display_dataframe(df):  # pragma: no cover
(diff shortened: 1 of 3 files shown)


cgnorthcutt:
Output of `dataset.health_summary` now looks like this if run in a jupyter notebook.

image

cleanlab/internal/util.py
```diff
@@ -451,3 +451,20 @@ def compress_int_array(int_array, num_possible_values):
...
+        from IPython.display import display
+
+        # no exception, let's determine if we are running inside a jupyter notebook
+        shell = get_ipython().__class__.__name__
+        if shell == "ZMQInteractiveShell":
+            display(df)  # Jupyter notebook or qtconsole
+        elif shell == "TerminalInteractiveShell":
+            print(df)  # terminal, not notebook
+        else:
+            print(df)  # something else, but not notebook
```
jwmueller, line 468: An alternative could be:  just `display(df)` in L460 and delete these extra checks. They anyway just goto L470, which would happen anyway if `try` clause fails.

I ask because not sure if `ZMQInteractiveShell` is overly limiting. Maybe `display(df)` works in other envs as well?  

Should keep as is if there's some specific situation where `display(df)` will not cause `try` to fail, but is not desirable to execute?

jwmueller: FYI from my terminal console (shell  = TerminalInteractiveShell), both display(df) and print(df) work fine. They just give same output.

jwmueller: Eg. in colab, I don't think `shell == "ZMQInteractiveShell"` but display(df) still would be preferred over print(df)

jwmueller (approved):
LGTM, just suggested possible minor improvement of the display(df) part

cleanlab/internal/util.py
cgnorthcutt: Agree, but similarly do not know the answer to your last question. I can make this change, but it makes exception handling more challenging for developers in the future if we wish to expand this functionality.

cgnorthcutt: to your second and third comment, we still need error checking because in my console, i don't have IPython installed so it throws an error if i try to display

PR #215, opened 2022-04-15
Configure display setting of Dataframe in 2.0 tutorial

JohnsonKuan:
Configure display setting of Dataframe in 2.0 tutorial to address issues with background color of header in dark mode for docs.

Background color of rows will be addressed in another PR after 2.0 tagging.

```
from IPython.core.display import HTML
HTML("""
<style>
    .dataframe {
        background: #D7D7D7;
    }
    
    th {
        color:black;
    }
</style>
""")
```

Preview in dark mode of Jupyter Lab:
image

Preview in non-dark mode:
image

### DIFF
--- a/docs/source/tutorials/indepth_overview.ipynb
+++ b/docs/source/tutorials/indepth_overview.ipynb
@@ -80,6 +80,24 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%HTML\n",
+    "<style>\n",
+    "    .dataframe {\n",
+    "        background: #F0F0F0;\n",
+    "    }\n",
+    "    \n",
+    "    th {\n",
+    "        color:black;\n",
+    "    }\n",
+    "</style>"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
    "metadata": {
     "id": "avXlHJcXjruP"
    },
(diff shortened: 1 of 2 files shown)


jwmueller:
Won't this be visible in the docs build and in the colab notebook?

docs/source/tutorials/indepth_overview.ipynb
```diff
@@ -77,6 +77,24 @@
     "%config InlineBackend.print_figure_kwargs={\"facecolor\": \"w\"}"
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%HTML\n",
```
jwmueller, line 86: Won't this be visible in the docs build and in the colab notebook?

cgnorthcutt (approved):
LGTM. Double check jonas's comment. After merge, double check it worked (takes a few minutes to load) and double check these settings and weird code blocks are not visible to user.

docs/source/tutorials/indepth_overview.ipynb
jwmueller: At least need to replace `metadata` above with:

```"metadata": {
    "nbsphinx": "hidden"
   },
```

That will at least hide in the docs website.

JohnsonKuan: Yes I believe so, the code will be visible. I'm not sure how to hide it though.

jwmueller: Actually if you want to add this stuff to this notebook it should go into the cell that is already hidden (the one starting with `# Package installation.`)

JohnsonKuan: Sounds good, agree. I moved it to the cell that is already hidden.

PR #219, opened 2022-04-16
Deploy docs when new release is tagged

weijinglok:
This PR fixes the issue where docs are not deployed when a new release is tagged. 

Previously, it only deploys if there is a trigger from the `master` branch. I've added an OR condition to also deploy if it's a trigger from a new release tag.


jwmueller (approved):
LGTM, wonder why mockup did not catch this issue?

PR #228, opened 2022-04-18
Remove unneeded alt text

anishathalye:
The anchor already says "Run in Google Colab", so if the logo fails to load, it will say "Google Colab Logo Run in Google Colab"; it's better to have the alt text just be empty.


weijinglok (approved):
LGTM!

PR #229, opened 2022-04-18
Generalize text tutorial to multiclass datasets

calebchiam:
The current training model assumes that the dataset has binary labels. This PR tweaks the model setup to generalize to datasets where there are >2 classes. As part of this change, the ROC-AUC score has been replaced with Log Loss.

### DIFF
--- a/docs/source/tutorials/text.ipynb
+++ b/docs/source/tutorials/text.ipynb
@@ -128,7 +128,8 @@
    "execution_count": null,
    "outputs": [],
    "source": [
-    "num_classes = len(set(full_labels))"
+    "num_classes = len(set(full_labels))\n",
+    "print(f\"Classes: {set(full_labels)}\")"
    ],
    "metadata": {}
   },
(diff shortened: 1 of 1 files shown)


docs/source/tutorials/text.ipynb
```diff
@@ -123,6 +123,20 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "outputs": [],
+   "source": [
+    "num_classes = len(set(full_labels))"
+   ],
+   "metadata": {
+    "collapsed": false,
+    "pycharm": {
+     "name": "#%%\n"
```
weijinglok, line 136: Might consider removing these as it seems specific to PyCharm IDE. might flicker back and forth if any contributor just collapses/hides this code cell.

weijinglok (commented):
Overall LGTM! just a tiny comment to consider.

docs/source/tutorials/text.ipynb
calebchiam: Good catch -- I didn't even notice Pycharm added that in. Removed.

jwmueller (approved):
Good call generalizing this multiclass!
I just made 2 very minor edits

PR #235, opened 2022-04-20
Make small fixes

anishathalye:
Note: the docs system and Jupyter render Markdown cells slightly differently. E.g. this renders as a paragraph followed by a bulleted list in Jupyter, but in the rendered docs, it will render as a paragraph that contains the `* bulleted list` as part of the _text_ in the paragraph.

```
An example paragraph.
* bulleted list that doesn't have an empty newline before it
```


jwmueller (approved):
LGTM

PR #239, opened 2022-04-20
update release instructions for pypi and conda.

cgnorthcutt:
* Removing --universal tag since we no longer support py2
* replace -r with -m since its now markdown.


anishathalye (approved):
btw, `-m` is not a Markdown version of `-r`; `-m` means verify metadata (which seems good to include).

PR #241, opened 2022-04-21
Fix edgecase divide-by-0 in entropy-score

jwmueller:
Note: still some redundant compute/clipping steps in get_confidence_weighted_entropy_for_each_label() due to transformations not being analytically calculated out. Current division of the function into entropy() and then confidence_weighted_entropy() is the cause of this as some steps in entropy-calculation are canceled out by confidence_weighted_entropy-rescaling transformation. Likely better to just have confidence_weighted_entropy() be its own function that does not call subroutine beyond self_confidence().

### DIFF
--- a/cleanlab/rank.py
+++ b/cleanlab/rank.py
@@ -440,14 +440,15 @@ def get_confidence_weighted_entropy_for_each_label(
       being correctly labeled.
     """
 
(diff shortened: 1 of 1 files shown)


JohnsonKuan (approved):
LGTM!

PR #247, opened 2022-04-30
Developed class that allow the use of cleanlab with tensorflow and huggingface models

MattiaSangermano:
This is the way the class can be used:

```python
import tensorflow as tf
import tensorflow_datasets as tfds
import numpy as np
from cleanlab.experimental.huggingface_keras_classifier import HuggingfaceKerasClassifier
from transformers import AutoTokenizer, TFAutoModel
import os
from sklearn.metrics import accuracy_score


os.environ['TOKENIZERS_PARALLELISM']='false'

def build_model(model_name:str, max_len:int, n_classes:int):
    # 1 - Input ids and attention mask as input to NN
    input_ids = tf.keras.layers.Input(
        shape=(max_len,), dtype='int32', name='input_ids')
    attention_mask = tf.keras.layers.Input(
        shape=(max_len,), dtype='int32', name='attention_mask')

    # 2 - Get bert Main layer and add it to the NN, passing input ids and attention mask as inputs
    bert_layer = TFAutoModel.from_pretrained(model_name)
    
    layer = bert_layer(input_ids=input_ids, attention_mask=attention_mask)[1]
    # embeddings = tf.keras.layers.Embedding(30_522 + 1, 128)(input_ids)
    # flatten = tf.keras.layers.Flatten()(embeddings)
    output_layer = tf.keras.layers.Dense(
      n_classes, activation='sigmoid')(layer)

    # model instance
    model = tf.keras.Model(inputs=[input_ids, attention_mask], outputs=output_layer)
    model.summary()
    return model

MODEL_NAME = "bert-base-uncased"
train = tfds.load('imdb_reviews', split='train', shuffle_files=True)
val = tfds.load('imdb_reviews', split='test', shuffle_files=True)
train_df = tfds.as_dataframe(train)[0:1000]
train_df['text'] = train_df['text'].apply(lambda x: x.decode('utf-8'))
val_df = tfds.as_dataframe(val)[0:1000]
val_df['text'] = val_df['text'].apply(lambda x: x.decode('utf-8'))
tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME)

train_input = tokenizer(
  train_df['text'].to_list(),
  padding='max_length',
  truncation=True,
  max_length=20,
  return_tensors='tf')

val_input = tokenizer(
  val_df['text'].to_list(),
  padding='max_length',
  truncation=True,
  max_length=20,
  return_tensors='tf',
  return_token_type_ids=False)

early_stopping = tf.keras.callbacks.EarlyStopping(
        monitor='val_accuracy', mode='max', verbose=1, patience=3, restore_best_weights=True)

val_dataset = tf.data.Dataset.from_tensor_slices(
                (
                    dict(val_input),
                    val_df['label'],
                )
            ).batch(32)

model = HuggingfaceKerasClassifier(
    # --- model function parameters ---
    model=build_model,
    model_name=MODEL_NAME,
    max_len=20,
    n_classes=1,
    # --- HuggingfaceKerasClassifier Parameters ---
    seq_len=20,
    train_input=dict(train_input),
    #   --- TF training Parameters ---
    optimizer=tf.keras.optimizers.Adam(2e-5),
    loss=tf.keras.losses.BinaryCrossentropy(),
    metrics=['accuracy'],
    epochs=10,
    batch_size=64,
    shuffle=True,
    callbacks=[early_stopping],
    verbose=True
)

print('Model Created.')
lnl = CleanLearning(clf=model, cv_n_folds=3)

print('Training CleanLearning model...')
training_ids = np.array(list(range(len(train_input['input_ids']))))

# Note that the validation dataset provided in the clf_kwargs is not related 
# to the validation set used internally by cleanlab. In this case, the validation 
# set is required by the early stopping callback used during the trainings.
lnl.fit(training_ids, np.array(train_df['label']), clf_kwargs={'validation_data': val_dataset})
print('Model trained.')

predictions = lnl.predict(test_input = val_input)
print('Accuracy on val data: ', accuracy_score(val_df['label'], predictions))

```

Once the pull request is accepted, I will put an example inside the [examples](https://github.com/cleanlab/examples) repo.

### DIFF
--- a/cleanlab/experimental/tf_nlp.py
+++ b/cleanlab/experimental/tf_nlp.py
@@ -44,9 +44,7 @@ class MultiInputKerasClassifier(KerasClassifier):
     def _get_tf_input(self, ids, train_input):
         indexed_input_ids = np.array(tf.gather(train_input["input_ids"], indices=ids))
 
-        indexed_attention_mask = np.array(
-            tf.gather(train_input["attention_mask"], indices=ids)
-        )
+        indexed_attention_mask = np.array(tf.gather(train_input["attention_mask"], indices=ids))
 
         return np.hstack([indexed_input_ids, indexed_attention_mask])
 
--- a/cleanlab/experimental/tf_nlp.py
+++ b/cleanlab/experimental/tf_nlp.py
@@ -1,31 +1,66 @@
 """
 A Scikeras classifier (adapted for huggingface models) which
-can be used for finding label issues in any nlp dataset.
+can be used for finding label issues in any text datasets.
 """
 
+from typing import Dict
 from scikeras.wrappers import KerasClassifier
 from sklearn.preprocessing import FunctionTransformer
 import tensorflow as tf
 import numpy as np
 
 
-class MultiInputKerasClassifier(KerasClassifier):
-    def __init__(self, train_input, seq_len, valid_input=None, y_val=None, **kwargs):
+class HuggingfaceKerasClassifier(KerasClassifier):
+    def __init__(self, train_input: Dict, seq_len: int, **kwargs):
         """
         Basic scikeras/scikeras does not directly provide the option of having
-        a multi-input model. However, you can work around this problem by
+        a multi-input model, i.e the input must be in the form (num_samples, feature_size). However, you can work around this problem by
         adding the `feature_encoder` property to the class which extends
         Scikitlearn's BaseEstimator.
+
+        Example of use:
+
+        * ``model = HuggingfaceKerasClassifier(``
+        * ``    # --- model function parameters ---``
+        * ``    model=model_fn,``
+        * ``    n_classes=2,``
+        * ``    # --- HuggingfaceKerasClassifier Parameters ---``
+        * ``    train_input=dict(train_input),``
+        * ``    seq_len=20,``
+        * ``    #   --- TF training Parameters ---``
+        * ``    optimizer=tf.keras.optimizers.Adam(2e-5),``
+        * ``    loss=tf.keras.losses.BinaryCrossentropy(),``
+        * ``    metrics=['accuracy'],``
+        * ``    epochs=10,``
+        * ``    batch_size=64,``
+        * ``    shuffle=True,``
+        * ``    callbacks=[early_stopping],``
+        * ``    verbose=True``
+        * ``)``
+        * `` ``
+        * ``lnl = CleanLearning(clf=model, cv_n_folds=3)``
+        * ``lnl.fit(training_ids, train_labels, clf_kwargs={'validation_data': val_dataset})``
+        * ``lnl.score(test_ids, test_labels)``
 
         References:
         - https://towardsdatascience.com/scikeras-tutorial-a-multi-input-multi-output-wrapper-for-capsnet-hyperparameter-tuning-with-keras-3127690f7f28
         - https://www.adriangb.com/scikeras/stable/notebooks/DataTransformers.html#4.-Multiple-inputs
 
+        Parameters
+        ----------
+        train_input : dictionary or pandas DataFrame,
+        Tokenized input data. Must contains the following keys/columns (For a more detailed explanation refers to `Huggingface documentation <https://huggingface.co/docs/transformers/main_classes/tokenizer#transformers.PreTrainedTokenizer.__call__>`):
+        - `input_ids`: a list or tensorflow tensor of token ids to be fed to the model. Shape = (num_sample, sequence_len).
+        - `attention_mask` a list or tensorflow tensor of indices specifying which tokens should be attended to by the model. Shape = (num_sample, sequence_len).
+
+        sequence_len : int,
+        Tokenized sentences length.
+
+        kwargs : optional,
+        Optional arguments useful to fit a `Tensorflow model`. Refers to Tensorflow documentation for a detailed list of attributes `<https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit>`
         """
         super().__init__(**kwargs)
         self.train_input = train_input
-        self.valid_input = valid_input
-        self.y_val = y_val
         self.seq_len = seq_len
 
     def split_input(self, X):
@@ -48,27 +83,47 @@ class MultiInputKerasClassifier(KerasClassifier):
 
         return np.hstack([indexed_input_ids, indexed_attention_mask])
 
-    def fit(self, ids, y, sample_weight=None):
+    def fit(self, ids, y, sample_weight=None, **kwargs):
+        """
+        Constructs and fit a new model using the given data.
+        Parameters
+        ----------
+        ids : array-like of shape (n_samples,)
+            Ids of training samples to be used to train the model.
+        y : array-like of shape (n_samples,)
+            True labels.
+        sample_weight : array-like of shape (n_samples,), default=None
+            Array of weights that are assigned to individual samples.
+            If not provided, then each sample is given unit weight.
+        **kwargs : Dict[str, Any]
+            Extra arguments to route to Tensorflow `Model.fit`.
+        """
         X = self._get_tf_input(ids, self.train_input)
-        val_dataset = None
+        return super().fit(X, y, sample_weight=sample_weight, **kwargs)
 
-        if self.valid_input:
-            val_dataset = tf.data.Dataset.from_tensor_slices(
-                (
-                    {
-                        "input_ids": self.valid_input["input_ids"],
-                        "attention_mask": self.valid_input["attention_mask"],
-                    },
-                    self.y_val,
-                )
-            ).batch(32)
+    def predict_proba(self, ids, **kwargs):
+        """
+        Returns class probability estimates for the given data.
+        Parameters
+        ----------
+        ids : array-like of shape (n_samples,)
+            Ids of training samples to be used to train the model.
+        **kwargs : Dict[str, Any]
+            Extra arguments to route to Tensorflow ``Model.predict``.
+        """
 
-        return super().fit(X, y, validation_data=val_dataset)
-
-    def predict_proba(self, ids):
         X = self._get_tf_input(ids, self.train_input)
-        return super().predict_proba(X)
+        return super().predict_proba(X, **kwargs)
 
-    def predict(self, ids):
+    def predict(self, ids, **kwargs):
+        """
+        Returns predictions for the given data.
+        Parameters
+        ----------
+        ids : array-like of shape (n_samples,)
+            Ids of training samples to be used to train the model.
+        **kwargs : Dict[str, Any]
+            Extra arguments to route to Tensorflow `Model.predict`.
+        """
         X = self._get_tf_input(ids, self.train_input)
-        return super().predict(X)
+        return super().predict(X, **kwargs)
--- a/cleanlab/experimental/tf_nlp.py
+++ b/cleanlab/experimental/tf_nlp.py
@@ -10,15 +10,8 @@ import numpy as np
 
 
 class MultiInputKerasClassifier(KerasClassifier):
-    def __init__(
-        self,
-        train_input,
-        seq_len,
-        valid_input=None,
-        y_val=None,
-        **kwargs
-    ):
-        '''
+    def __init__(self, train_input, seq_len, valid_input=None, y_val=None, **kwargs):
+        """
         Basic scikeras/scikeras does not directly provide the option of having
         a multi-input model. However, you can work around this problem by
         adding the `feature_encoder` property to the class which extends
@@ -28,7 +21,7 @@ class MultiInputKerasClassifier(KerasClassifier):
         - https://towardsdatascience.com/scikeras-tutorial-a-multi-input-multi-output-wrapper-for-capsnet-hyperparameter-tuning-with-keras-3127690f7f28
         - https://www.adriangb.com/scikeras/stable/notebooks/DataTransformers.html#4.-Multiple-inputs
 
-        '''
+        """
         super().__init__(**kwargs)
         self.train_input = train_input
         self.valid_input = valid_input
@@ -37,9 +30,9 @@ class MultiInputKerasClassifier(KerasClassifier):
 
     def split_input(self, X):
         splitted_X = [
-            X[:, :self.seq_len],  # input_ids
-            X[:, self.seq_len:]   # attention_mask
-            ]
+            X[:, : self.seq_len],  # input_ids
+            X[:, self.seq_len :],  # attention_mask
+        ]
         return splitted_X
 
     @property
@@ -49,16 +42,10 @@ class MultiInputKerasClassifier(KerasClassifier):
         )
 
     def _get_tf_input(self, ids, train_input):
-        indexed_input_ids = np.array(
-            tf.gather(
-                train_input['input_ids'],
-                indices=ids)
-        )
+        indexed_input_ids = np.array(tf.gather(train_input["input_ids"], indices=ids))
 
         indexed_attention_mask = np.array(
-            tf.gather(
-                train_input['attention_mask'],
-                indices=ids)
+            tf.gather(train_input["attention_mask"], indices=ids)
         )
 
         return np.hstack([indexed_input_ids, indexed_attention_mask])
@@ -68,10 +55,15 @@ class MultiInputKerasClassifier(KerasClassifier):
         val_dataset = None
 
         if self.valid_input:
-            val_dataset = tf.data.Dataset.from_tensor_slices(({
-                'input_ids': self.valid_input['input_ids'],
-                'attention_mask': self.valid_input['attention_mask'],
-                }, self.y_val)).batch(32)
+            val_dataset = tf.data.Dataset.from_tensor_slices(
+                (
+                    {
+                        "input_ids": self.valid_input["input_ids"],
+                        "attention_mask": self.valid_input["attention_mask"],
+                    },
+                    self.y_val,
+                )
+            ).batch(32)
 
         return super().fit(X, y, validation_data=val_dataset)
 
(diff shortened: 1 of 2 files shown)


jwmueller:
Awesome contribution @MattiaSangermano!  If you're interested in contributing more or authoring some blogs/tutorials, feel free to chat with our team in our Slack community channel: 
https://join.slack.com/t/cleanlab-community/shared_invite/zt-17lszn4hv-gg2FhZPXYfljq_l01uo92g

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,74 @@
+"""
+A Scikeras classifier (adapted for huggingface models) which
+can be used for finding label issues in any nlp dataset.
```
jwmueller, line 3: may be clearer to replace nlp -> text

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,74 @@
...
+from sklearn.preprocessing import FunctionTransformer
+import tensorflow as tf
+import numpy as np
+
+
+class MultiInputKerasClassifier(KerasClassifier):
+    def __init__(self, train_input, seq_len, valid_input=None, y_val=None, **kwargs):
+        """
+        Basic scikeras/scikeras does not directly provide the option of having
+        a multi-input model. However, you can work around this problem by
```
jwmueller, line 16: Would be helpful to more precisely define (or give example of) "multi-input".

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,74 @@
...
+    @property
+    def feature_encoder(self):
+        return FunctionTransformer(
+            func=self.split_input,
+        )
+
+    def _get_tf_input(self, ids, train_input):
+        indexed_input_ids = np.array(tf.gather(train_input["input_ids"], indices=ids))
+
+        indexed_attention_mask = np.array(tf.gather(train_input["attention_mask"], indices=ids))
```
jwmueller, line 47: This class only seems to work with very specifically formatted `train_input`, so needs to be documented in detail what exactly this format is and what it represents.

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,74 @@
...
+
+        return np.hstack([indexed_input_ids, indexed_attention_mask])
+
+    def fit(self, ids, y, sample_weight=None):
```
jwmueller, line 51: Documentation needed for args: ids, y, sample_weight

jwmueller: Why did you decide to pass some data into the __init__() of this class and some data into fit()?  It seems nicer to pass all the data into fit() instead of __init__()

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,74 @@
...
+        if self.valid_input:
+            val_dataset = tf.data.Dataset.from_tensor_slices(
+                (
+                    {
+                        "input_ids": self.valid_input["input_ids"],
+                        "attention_mask": self.valid_input["attention_mask"],
+                    },
+                    self.y_val,
+                )
+            ).batch(32)
```
jwmueller, line 64: Make  batch = 32 an optional argument instead of hardcoding it?

jwmueller: And why specify particular validation batch-size but not one for training?

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,74 @@
...
+can be used for finding label issues in any nlp dataset.
+"""
+
+from scikeras.wrappers import KerasClassifier
+from sklearn.preprocessing import FunctionTransformer
+import tensorflow as tf
+import numpy as np
+
+
+class MultiInputKerasClassifier(KerasClassifier):
```
jwmueller, line 12: Can you propose a more precise name for this class?  It doesn't appear to me to be a generic MultiInputKerasClassifier. Eg. the inputs are expected to be text and of a very particular format with `train_input` having keys with specific names like ""input_ids", "attention_mask",

cleanlab/experimental/tf_nlp.py
MattiaSangermano: Yes, I would have liked to do it that way, but when I call the `CleanLearning` fit I must pass a 2-D array as X. In this case the input is a 3-D like array. So what is happening now is that in the `__init__` the training dataset is loaded and as X of the `fit` I pass the ids corresponding to the data I want to use during the training. This is something that I don't like, but it is the only way that I found to make cleanlab work with multi-input data. If you have any suggestions to avoid this workaround I would appreciate it.

jwmueller: Note that CleanLearning does support non-2D array data as demonstrated in this unit test:
https://github.com/cleanlab/cleanlab/blob/master/tests/test_classification.py#L567

But yes I see your point that it unfortunately does not support non-strictly array data and have opened an issue for this: 
https://github.com/cleanlab/cleanlab/issues/252

Thanks for pointing it out!  Let me know if you'd like to help with this :)

jwmueller: Although the current way you had to get around this issue is not pretty, I also do not see an alternative work-around until we've added support for non-strictly array data in CleanLearning. If you want to address the other feedback on this PR, we should hopefully have this sorted out reasonably quickly.

MattiaSangermano: Happy to hear that you are working towards the integration of different input types in CleanLab, it would be really useful!

In the coming day I will fix the remaining suggestions. After that, we may discuss whether to use the non-strictly arrays within this or in a different PR.

jwmueller: Sound good :) 
By then we should have fixed the issue hopefully and then you can avoid the current hack.

cleanlab/experimental/tf_nlp.py
MattiaSangermano: I just noticed that when we will use non-strictly arrays most probably `train_input` won't have a particular format. So, if you agree then I would keep this part temporarily on hold.

jwmueller: Right understood that `train_input` can have whatever format you like here.
I'm just suggesting that you can add a little documentation to describe what this particular `train_input` is supposed to look like for your specific class.  For example, `train_input` clearly has to have fields like: "input_ids", "attention_mask", etc, so for each of these fields:  can you just add a 1 sentence description of the field or provide a link to another website where the field is defined?

Thanks! Having some (even minimal) documentation will be very helpful to our users!

MattiaSangermano:
@jwmueller Please, take a look at the new version. As you can see, in addition to the documentation I removed the hardcoded parts as much as possible. Now in order to use the validation set or any additional arguments to the fit method is sufficient to pass them using the `clf_kwargs` inside `CleanLearning.fit` method. Please let me know what you think.

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
...
+import numpy as np
+
+
+class HuggingfaceKerasClassifier(KerasClassifier):
+    def __init__(self, train_input: Dict, seq_len: int, **kwargs):
+        """
+        Basic scikeras/scikeras does not directly provide the option of having
+        a multi-input model, i.e the input must be in the form (num_samples, feature_size). However, you can work around this problem by
+        adding the `feature_encoder` property to the class which extends
+        Scikitlearn's BaseEstimator.
```
jwmueller, line 19: Before the sentence starting with "However, ...", let's add one more sentence to explain what is "multi-input" about this model. Something like:

"In our case, the multiple inputs are: (1) token ids (input_ids); (2) indices specifying what tokens to attend to (attention_mask)".

jwmueller: Also this sentence makes it sound like the user has to do this, but you've already nicely done it for them!  So I'd replace:
"However, you can work around this problem by adding the `feature_encoder` property to the class which extends Scikitlearn's BaseEstimator."
->
"Here we work around this by the `feature_encoder` property to our class which extends Scikit-learn's BaseEstimator"

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
+"""
+A Scikeras classifier (adapted for huggingface models) which
```
jwmueller, line 2: "adapted for huggingface models"
->
"adapted for fine-tuning HuggingFace pretrained Transformer models"

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
+"""
```
jwmueller, line 1: Can we name this file to match the class name?
I suggest filename: `huggingface_keras_classifier.py`

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
...
+            If not provided, then each sample is given unit weight.
+        **kwargs : Dict[str, Any]
+            Extra arguments to route to Tensorflow `Model.fit`.
+        """
+        X = self._get_tf_input(ids, self.train_input)
+        return super().fit(X, y, sample_weight=sample_weight, **kwargs)
+
+    def predict_proba(self, ids, **kwargs):
+        """
+        Returns class probability estimates for the given data.
```
jwmueller, line 106: "for the given data"
-> 
"for particular training examples"

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
...
+        **kwargs : Dict[str, Any]
+            Extra arguments to route to Tensorflow ``Model.predict``.
+        """
+
+        X = self._get_tf_input(ids, self.train_input)
+        return super().predict_proba(X, **kwargs)
+
+    def predict(self, ids, **kwargs):
+        """
+        Returns predictions for the given data.
```
jwmueller, line 120: "for the given data"
-> 
"for particular training examples"

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
...
+        X = self._get_tf_input(ids, self.train_input)
+        return super().fit(X, y, sample_weight=sample_weight, **kwargs)
+
+    def predict_proba(self, ids, **kwargs):
+        """
+        Returns class probability estimates for the given data.
+        Parameters
+        ----------
+        ids : array-like of shape (n_samples,)
+            Ids of training samples to be used to train the model.
```
jwmueller, line 110: "Ids of training samples to be used to train the model."
->
"Indices of training examples for which to produce predictions."

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
...
+        X = self._get_tf_input(ids, self.train_input)
+        return super().predict_proba(X, **kwargs)
+
+    def predict(self, ids, **kwargs):
+        """
+        Returns predictions for the given data.
+        Parameters
+        ----------
+        ids : array-like of shape (n_samples,)
+            Ids of training samples to be used to train the model.
```
jwmueller, line 124: "Ids of training samples to be used to train the model."
->
"Indices of training examples for which to produce predictions."

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
...
+        sample_weight : array-like of shape (n_samples,), default=None
+            Array of weights that are assigned to individual samples.
+            If not provided, then each sample is given unit weight.
+        **kwargs : Dict[str, Any]
+            Extra arguments to route to Tensorflow `Model.fit`.
+        """
+        X = self._get_tf_input(ids, self.train_input)
+        return super().fit(X, y, sample_weight=sample_weight, **kwargs)
+
+    def predict_proba(self, ids, **kwargs):
```
jwmueller, line 104: After user has run `Cleanlearning.fit()`, how are they supposed to use this classifier to produce predictions on *new* test data?  Your example did not demonstrate how that would work.

jwmueller: I think `predict()` and `predict_proba()` should look more like this (I've only shown `predict` below but `predict_proba` should look similar). What do you think?

```
def predict_proba(self, ids=None, *, test_input=None, **kwargs):
    if ids is not None:
        if test_input is not None:
            raise ValueError("One of ids or test_input must be None")
        X = self._get_tf_input(ids, self.train_input)
    elif test_input is not None:
        X = self._get_tf_input(np.arange(len(test_input), test_input)  # probably need to replace len(...) with proper way to get length of the test_input tokenizer 
    else:
        raise ValueError("Both ids and test_input cannot be None")
    return super().predict_proba(X, **kwargs)
```

Then presumably your example code could look like this to appear more complete:  
```
cl = Cleanlearning.fit(...)`:
test_preds = cl.predict(val_input)
# report test accuracy of test_preds vs test_df['label']
```

jwmueller:
> @jwmueller Please, take a look at the new version. As you can see, in addition to the documentation I removed the hardcoded parts as much as possible. Now in order to use the validation set or any additional arguments to the fit method is sufficient to pass them using the `clf_kwargs` inside `CleanLearning.fit` method. Please let me know what you think.

Awesome work, I think it's almost there!  Just need to make sure your `predict` and `predict_proba` functions can operate on new data too, not just only work with `CleanLearning.fit`.  Once you've done that, please update your example in the comment as well to reflect the new code, and I'll test it out and merge if it works smoothly.  I think the last 3 lines of the example should look like something this for completeness:

```
cl = Cleanlearning.fit(...)`:
test_preds = cl.predict(test_input=test_input)
# report test accuracy of test_preds vs test_df['label']
```

I'd also consistently refer to the test data as "test" instead of "val" (so above I suggest replacing your current `val_input` variable with one named `test_input`).

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
+"""
+A Scikeras classifier (adapted for huggingface models) which
+can be used for finding label issues in any text datasets.
+"""
```
jwmueller, line 4: Please add one final sentence which comments on which package versions you used for each dependency in this file, ie. what version of scikeras, tensorflow, etc.

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
...
+        * ``    epochs=10,``
+        * ``    batch_size=64,``
+        * ``    shuffle=True,``
+        * ``    callbacks=[early_stopping],``
+        * ``    verbose=True``
+        * ``)``
+        * `` ``
+        * ``lnl = CleanLearning(clf=model, cv_n_folds=3)``
+        * ``lnl.fit(training_ids, train_labels, clf_kwargs={'validation_data': val_dataset})``
+        * ``lnl.score(test_ids, test_labels)``
```
jwmueller, line 43: I don't think `lnl.score()` is actually computing the correct thing right now, due to the problem I pointed out below for how `predict()` is implemented?

cleanlab/experimental/tf_nlp.py
```diff
@@ -0,0 +1,129 @@
...
+
+class HuggingfaceKerasClassifier(KerasClassifier):
+    def __init__(self, train_input: Dict, seq_len: int, **kwargs):
+        """
+        Basic scikeras/scikeras does not directly provide the option of having
+        a multi-input model, i.e the input must be in the form (num_samples, feature_size). However, you can work around this problem by
+        adding the `feature_encoder` property to the class which extends
+        Scikitlearn's BaseEstimator.
+
+        Example of use:
```
jwmueller, line 21: This is overall an awesome contribution!  If this code happens to identify interesting label issues in the IMDB dataset (you can try `lnl.find_label_issues()` to quickly find out), then might be interested in submitting your example and the label issues you found in IMDB as a guest post for the HuggingFace blog?  https://huggingface.co/blog

I think they might be interested in featuring that!  Feel free to chat with me in our [Slack channel](https://join.slack.com/t/cleanlab-community/shared_invite/zt-17lszn4hv-gg2FhZPXYfljq_l01uo92g) if you're interested in discussing details.

cleanlab/experimental/tf_nlp.py
MattiaSangermano: Yes, we can manage new data in this way, even if the best thing would be to take the original data as input instead of the indexes. In that way, the code would be better and there would be no need to have the `test_input` extra field. For the time being, we can modify the code as you suggested and fix this issue in future.

cleanlab/experimental/tf_nlp.py
MattiaSangermano: Honestly, I don't know why there is that line, I think comes from my personal code because it really doesn't make sense in our case. I will replace it with an example of the predict function if you agree.

cleanlab/experimental/tf_nlp.py
jwmueller: yes sounds good thanks for making the update!

jwmueller:
Code looks good, I'm ready to approve the PR!
Could you just update the code example in your PR description comment to reflect the new code (along with adding the `test_preds = predict(test_data)` and `accuracy(test_preds, test_labels)` to make the example more complete?

I'll merge this PR once I've verified the update code example runs properly.

jwmueller (approved):
Verified this runs well, thanks for the amazing contribution!  You can make corresponding PR to our examples now :)

jwmueller:
Two small changes in your example script I would consider for the examples/ PR are:

1) Add the verbose arg here:
```
lnl = CleanLearning(clf=model, cv_n_folds=3, verbose=True)
```

2) Explicitly define variable:
```
num_epochs = 10  # small value used here to limit runtime, may prefer larger value in real applications
```

Issue #33, opened 2020-07-03
May I set the sample_weight parameter manually？


liumengna123:
Hi, It's a pretty lib. When I try to train model with cleanlab, I find it is not allowed to set sample_weight manually. It is inteligent to set sample weight according to the ratio of positive samples and negitave samples. However, I need to tun the paramater for better performance. 
Thank you for your contribution about this lib.

cgnorthcutt:
Sure thing. Are you familiar with how to submit a pull request? Please submit a pull request and I'll check it. Please add testing for the pull request.

cgnorthcutt:
@liumengna123 bump.

jwmueller:
This has been addressed in: https://github.com/cleanlab/cleanlab/pull/248
So you can now pass in custom datapoint weights via `sample_weight` argument of `CleanLearning.fit()` if you install the master branch version of cleanlab.

Feel free to re-open this issue if you still have questions!

PR #248, opened 2022-05-02
FeatureReq #33: Added custom sample_weight

rushic24:
Ref: https://github.com/cleanlab/cleanlab/issues/33

You can now add a custom sample weight like below

```
from sklearn.datasets import load_iris
import numpy as np

X, labels = load_iris(return_X_y=True)
clf = LogisticRegression(random_state=0)

cl = CleanLearning(clf=clf)
cl.fit(X=X, labels=labels, sample_weight=np.random.randn(len(labels)))
```

### DIFF
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -481,9 +481,24 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
                     "sample_weight"
                 ]  # pointer to here to avoid duplication
                 self.clf_final_kwargs["sample_weight"] = self.sample_weight
+                if self.verbose:
+                    print("Fitting final model on the clean data ...")
+            else:
+                if (
+                    "sample_weight" in inspect.getfullargspec(self.clf.fit).args
+                    and "sample_weight" not in self.clf_final_kwargs
+                    and self.noise_matrix is None
+                ):
+                    print(
+                        "Cannot utilize sample weights for final training. To utilize must either specify noise_matrix "
+                        "or have previously called self.find_label_issues() instead of filter.find_label_issues()"
+                    )
 
-            if self.verbose:
-                print("Fitting final model on the clean data ...")
+                if self.verbose:
+                    if "sample_weight" in self.clf_final_kwargs:
+                        print("Fitting final model on the clean data with custom sample_weight...")
+                    else:
+                        print("Fitting final model on the clean data ...")
 
         elif sample_weight is not None and "sample_weight" not in self.clf_final_kwargs:
             clf_final_kwargs["sample_weight"] = sample_weight[x_mask]
@@ -491,16 +506,6 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
                 print("Fitting final model on the clean data with custom sample_weight...")
 
         else:
-            if (
-                "sample_weight" in inspect.getfullargspec(self.clf.fit).args
-                and "sample_weight" not in self.clf_final_kwargs
-                and self.noise_matrix is None
-            ):
-                print(
-                    "Cannot utilize sample weights for final training. To utilize must either specify noise_matrix "
-                    "or have previously called self.find_label_issues() instead of filter.find_label_issues()"
-                )
-
             if self.verbose:
                 if "sample_weight" in self.clf_final_kwargs:
                     print("Fitting final model on the clean data with custom sample_weight...")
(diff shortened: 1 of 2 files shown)


jwmueller:
Hi @rushic24 many thanks for the contribution!  
Note you'll have to sign the CLA by clicking the link above before we can merge your PR, thanks!

cleanlab/classification.py
```diff
@@ -263,6 +263,7 @@ def fit(
         noise_matrix=None,
         inverse_noise_matrix=None,
         label_issues=None,
+        sample_weight=None,
```
jwmueller, line 266: I'm not sure we want to add `sample_weight` as an extra required argument of `fit()`. This makes this class no longer applicable to users' own bespoke models whose `fit()` function does not take in this argument.

Instead note that you can already pass in kwargs into the `clf_kwargs` argument. So training with `sample_weight` is already supported for `clf.fit()` during cross-validation. It is currently just not supported during the final `clf.fit()` call on the cleaned subset of data, but you could easily support it (see my tips below on how to do this).

cleanlab/classification.py
```diff
@@ -428,13 +432,15 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+        if sample_weight is not None:
+            sample_weight = sample_weight[x_mask]
         if self.verbose:
             print(f"Pruning {np.sum(self.label_issues_mask)} examples with label issues ...")
             print(f"Remaining clean data has {len(labels_cleaned)} examples.")
 
         # Check if sample_weight in args of clf.fit()
         if (
-            "sample_weight" in inspect.getfullargspec(self.clf.fit).args
+            sample_weight is None
```
jwmueller, line 443: The main change I would like to see is:  We should not assume `clf.fit()` takes in `sample_weight` as an argument. With your current code, CleanLearning is no longer applicable to users' own bespoke models whose `fit()` function does not take in this argument.

So we should not pass `sample_weight` into `self.clf.fit()` except when it is not None. The easiest way to do that would be to add `sample_weight` to `clf.kwargs` if it is not None.  If we go this route, here are the changes that need to happen:

1) remove `sample_weight` from the `clf.fit()` call.

2) add the following code-block near the start of `CleanLearning.fit()`:

```
if "sample_weight" in clf_kwargs:
    raise ValueError("sample_weight should be provided directly to fit() rather than as clf_kwargs")  # prevent there being multiple ways to pass in sample_weight

if sample_weight is not None:
    if "sample_weight" not in inspect.getfullargspec(self.clf.fit).args:
        raise ValueError("sample_weight must be a supported fit() argument for your model in order to be specified here")
   else:
        clf_kwargs["sample_weight"] = sample_weight
```

cleanlab/classification.py
```diff
@@ -462,6 +468,16 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
             if self.verbose:
                 print("Fitting final model on the clean data ...")
 
+            self.clf.fit(
+                x_cleaned,
+                labels_cleaned,
+                sample_weight=sample_weight,
```
jwmueller, line 474: remove sample_weight from clf.fit() call

cleanlab/classification.py
jwmueller: 3) Also cannot delete the check:
```
if "sample_weight" in inspect.getfullargspec(self.clf.fit).args
```
as this will be necessary to determine if cleanlab's built-in sample-weighting can be used or not.

So code for the final clf.fit() on cleaned data should look like:

```
if sample_weight is None:
    if "sample_weight" in inspect.getfullargspec(self.clf.fit).args and "sample_weight" not in self.clf_final_kwargs:
        # use cleanlab's built-in sample-weighting in final clf.fit()...
    else:
        call final clf.fit() without passing sample_weight 
elif sample_weight is not None and "sample_weight" not in self.clf_final_kwargs:
    sample_weight = sample_weight[x_mask]  # I would move this line from above to here
    clf_final_kwargs["sample_weight"] = sample_weight
    # call clf.fit() passing in only clf_final_kwargs, not sample_weight

jwmueller (approved):
Code is looking good, thanks for implementing the requested changes! 
 
I've added various fixes and refactored some of the code to make it more modular. Added unit tests for this as well. I'll merge this once you confirm it looks good!

If you still want to propose any additional changes, please let me know first before implementing them.

rushic24:
OMG, those seem to be a lot! looks good, thanks

PR #249, opened 2022-05-04
Updated project urls in setup.py

calebchiam:
The PyPI page has a section for Project links, such as Homepage, Documentation, Source Code, and Bug Tracker. (See [pandas](https://pypi.org/project/pandas/) as an example.)

This PR updates `setup.py` to include said project links.


jwmueller (approved):
Good stuff!

PR #255, opened 2022-05-10
Allow users to pass custom weights for ensemble label quality scoring

JohnsonKuan:
The purpose of this PR is to allow users to use custom weights for aggregating scores in the `get_label_quality_ensemble_scores` method.

To use custom weights, the user will need to set `weight_ensemble_members_by="custom"` and pass a numpy array of weights into `custom_weights`.

This PR also adds tests for scenarios using `custom_weights`.

### DIFF
--- a/cleanlab/rank.py
+++ b/cleanlab/rank.py
@@ -218,6 +218,7 @@ def get_label_quality_ensemble_scores(
     method: str = "self_confidence",
     adjust_pred_probs: bool = False,
     weight_ensemble_members_by: str = "accuracy",
+    custom_weights: np.array = None,
     verbose: bool = True,
 ) -> np.array:
     """Returns label quality scores based on predictions from an ensemble of models.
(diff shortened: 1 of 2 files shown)


cleanlab/rank.py
```diff
@@ -286,6 +291,12 @@ def get_label_quality_ensemble_scores(
             """
         )
 
+    # Raise a warning if user passed custom_weights array but did not choose weight_ensemble_members_by="custom"
+    if custom_weights is not None and weight_ensemble_members_by != "custom":
+        warnings.warn(
```
jwmueller, line 296: ValueError instead of warning? It seems misleading to ignore custom weights if they were purposefully passed in.

cleanlab/rank.py
```diff
@@ -252,11 +253,15 @@ def get_label_quality_ensemble_scores(
...
-    weight_ensemble_members_by : {"uniform", "accuracy"}, default="accuracy"
+    weight_ensemble_members_by : {"uniform", "accuracy", "custom"}, default="accuracy"
       Weighting scheme used to aggregate scores from each model:
 
       - "uniform": take the simple average of scores
       - "accuracy": take weighted average of scores, weighted by model accuracy
+      - "custom": take weighted average of scores using custom weights that the user passes to the custom_weights parameter.
+
+    custom_weights : np.array, default=None
+      Weights used to aggregate scores from each model if weight_ensemble_members_by="custom".
```
jwmueller, line 264: specify length of this array must match the number of models, i.e. `len(pred_probs_list)`.
Would be good to have explicit check that this actually matches too and raise ValueError if not

cleanlab/rank.py
JohnsonKuan: Thanks, good call. I made this change.

Initially, I chose warning because I thought the user could have accidentally passed a `custom_weights` array when choosing `"accuracy"` or `"uniform"` for `weight_ensemble_members_by`. But now that I think about it, this is unlikely and it makes more sense to raise a ValueError.

cleanlab/rank.py
JohnsonKuan: Thanks, done! I also added an assertion check.

jwmueller (approved):
LGTM

PR #257, opened 2022-05-12
Fix link checking

anishathalye:
For some reason, GitHub is returning a 403 to the GitHub actions runner when it tries to access these URLs. Maybe docs.github.com doesn't allow access from GitHub actions.


jwmueller (approved):
thanks!

Issue #85, opened 2021-10-08
Compute Confident Joint number of classes K


CompareSan:
Hi,

In the compute_confident_joint class there is the possibility of passing the number of classes K. But no matter what value of K I set, the library returns the dimension of the confident joint with K=len(np.unique(s)).  

If I want to use this library dynamically like in an iteration of a self-learning or active-learning algorithm in order to see what are the labels with high probability of being noisy, I should be able to set K as the number of unique classes I know I have in my classification problem and not as the number of unique classes the Clean Lab sees. At every iteration I might not see all the classes and this library in that case seems impossible to be used.

jcklie:
I encountered the same issue. Even when computing the calibration_join beforehand which allows setting `K`, the get_noise_indices uses unique label count. These methods should all respect `K`.

vtsouval:
@jcklie @filippoBUO It seems that `K` is not used inside 'compute_confident_joint'. I have written a modified version to overcome this issue.

```
def compute_confident_joint(labels, pred_probs, *, num_classes=None, thresholds=None, calibrate=True, multi_label=False, return_indices_of_off_diagonals=False,):

	from cleanlab.count import _compute_confident_joint_multi_label, calibrate_confident_joint, confusion_matrix
	import numpy as  np

	if multi_label:
		return _compute_confident_joint_multi_label(
			labels=labels,
			pred_probs=pred_probs,
			thresholds=thresholds,
			calibrate=calibrate,
			return_indices_of_off_diagonals=return_indices_of_off_diagonals,
		)

	# labels needs to be a numpy array
	labels = np.asarray(labels)

	# Find the number of unique classes
	if num_classes is None: num_classes = len(np.unique(labels))

	# Estimate the probability thresholds for confident counting
	if thresholds is None:
		# P(we predict the given noisy label is k | given noisy label is k)
		thresholds = [np.mean(pred_probs[:, k][labels == k]) for k in range(num_classes)]
	thresholds = np.asarray(thresholds)

	# Compute confident joint (vectorized for speed).

	# pred_probs_bool is a bool matrix where each row represents a training example as a boolean vector of
	# size num_classes, with True if the example confidently belongs to that class and False if not.
	pred_probs_bool = pred_probs >= thresholds - 1e-6
	num_confident_bins = pred_probs_bool.sum(axis=1)
	at_least_one_confident = num_confident_bins > 0
	more_than_one_confident = num_confident_bins > 1
[...]

PR #259, opened 2022-05-17
Set num classes based on num columns of pred_probs

jwmueller:
Instead of setting the number of classes based on `len(unique(labels))`. This addresses issue:  https://github.com/cleanlab/cleanlab/issues/85

Thanks @[vtsouval](https://github.com/vtsouval) for pointing out where some of the fixes had to be made and providing a helpful candidate implementation!

### DIFF
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -145,7 +145,7 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False):
     """
     num_classes = confident_joint.shape[0]
     if multi_label:
-        label_counts = value_counts([x for lst in labels for x in lst])  # TODO:FIXME
+        label_counts = value_counts([x for lst in labels for x in lst])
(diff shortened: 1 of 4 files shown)


cgnorthcutt:
Can you explain I'm the description what is the value added by this change? Is the goal to support pred probs and labels with certain classes missing? Can you provide an example input that didn't work before but should work now with this commit?

jwmueller:
The new unit tests added demonstrate examples that did not work before and now do with this commit. See also the linked issue for others' use-cases: https://github.com/cleanlab/cleanlab/issues/85

jwmueller:
When this PR is revisited and cleanlab supports missing classes that are present in `pred_probs` but not in `labels`, remember to correspondingly update:
* multi-label functionality
* multi-annotator functionality

which currently may contain some suboptimal code to get around this issue.

PR #260, opened 2022-05-18
Fix line index of CleanLearning(), some text of links, etc.

Yulv-git:
Fix line index of CleanLearning(), some text of links, etc.


anishathalye (approved):
Thank you for that great attention to detail :)

PR #262, opened 2022-05-19
Upgrade to latest version of coverage

anishathalye:
The issue with coverage==6.3 has been fixed: https://github.com/nedbat/coveragepy/issues/1310#issuecomment-1129894701.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 # Python dependencies for development
-coverage<6.3
+coverage != 6.3, != 6.3.*
 pre-commit
 pytest
 pytest-cov


anishathalye:
coverage==6.4 has been released, so we can merge this now.

anishathalye:
Oops, turns out that coverage 6.x dropped support for Python 3.6, so we need to think about how to handle that. We probably want to continue supporting 3.6, so maybe we stick with the old version of coverage for now?

anishathalye:
In working shape now, should be good to merge.

jwmueller (approved):
Thanks for bumping the version!

Issue #263, opened 2022-05-20
get_confident_thresholds fails with numpy error on multi-label problem, what are the limitations of multi-label ?


suneeta-mall:
Thanks a lot for your work on clean lab and for pulling all the theories together to build nice general framework for error management.  

I noticed that in your paper [here](https://arxiv.org/abs/2103.14749) & [here](https://arxiv.org/abs/1911.00068) you don't talk a whole a lot about multi-label support but your implementation (cleanlab) supports multi-label.  It's great that the support is there. However, fundamentally I am not sure how you represent multi-label as a pair-wise joint distribution and still consider off-diagonal elements in the joint matrix as noise?  Could you please provide some detail and thinking behind multi-level support?

For records, I have used version ` '2.0.0'` of cleanlab.
More than a theoretical question, I have run cleanlab over a usecase I have with multi-label and I can get the label noise and self-confidence metrics working, like following:
```
In [38]:     confident_joint_matrix = cleanlab.count.compute_confident_joint(
    ...:        truth
    ...:         predictions,
    ...:         multi_label=True,
    ...:     )
```
The result I get here is as expected the following (I have 6 class output, lets say my classes are  `['airplane', 'airport', 'buildings', 'cars', 'runway', 'trees']`). What I am unclear on is how do interpret this? I can have airplane and airport both a valid class so essentially means off-diagonal are no longer noise.
```
In [39]: confident_joint_matrix
Out[39]:
array([[119,  26,  36,  18,  29,   8],
       [ 28,  51,  60,   6,  59,  29],
       [ 24,  53,  63,   8,  64,  31],
       [ 12,   2,   6,   6,   0,   0],
       [  6,  50,  61,   0,  72,  34],
       [  2,  26,  31,   0,  34,  24]])
```

Also, another issue I ran into when I looked at estimating the threshold,  details of this is as following:
```
/core/_methods.py:189: RuntimeWarning: invalid value encountered in true_divide
  ret = ret.dtype.type(ret / rcount)
Out[40]: array([nan, nan, nan, nan, nan, nan], dtype=float32)
```
Maybe this error is more helpful:
```
   1091 def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array:
   1092     """Returns expected (average) "self-confidence" for each class.
   1093 
   1094     The confident class threshold for a class j is the expected (average) "self-confidence" for class j.
   (...)
   1114       An array of shape ``(K,)``.
   1115     """
[...]

PR #264, opened 2022-05-21
Fix bug. get thresholds broken for multi_label

cgnorthcutt:
Addresses the error found by @suneeta-mall in https://github.com/cleanlab/cleanlab/issues/263. Thanks Suneeta!

The bug was introduced  when the`get_confident_thresholds` method  was added to count.py (multi_label support was dropped the computation used does not work for multi_label -- this should be fixed by this PR).

In the future, more testing of multi_label labels should be added to our tests to catch errors like [this](https://github.com/cleanlab/cleanlab/issues/263).


cleanlab/count.py
```diff
@@ -1108,12 +1111,31 @@ def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array
...
     confident_thresholds : np.array
       An array of shape ``(K,)``.
     """
-    confident_thresholds = np.array(
-        [np.mean(pred_probs[:, k][labels == k]) for k in range(pred_probs.shape[1])]
-    )
+
+    if multi_label:
+        # Compute unique number of classes K by flattening labels (list of lists)
+        K = len(np.unique([i for lst in labels for i in lst]))
```
jwmueller, line 1132: why not define `K = pred_probs.shape[1]` like is being done below when multi_label=false?

jwmueller: otherwise I think anywhere num-classes are being computed based on num-unique-values, the docstring should explicitly state that all classes must be present in `labels` to avoid possible user confusion like we have seen in some Issues.

cleanlab/count.py
```diff
@@ -1108,12 +1111,31 @@ def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array
...
+        # Compute thresholds = p(label=k | k in set of given labels)
+        k_in_l = np.array([[k in lst for lst in labels] for k in range(K)])
+        # The avg probability of class given that the label is represented.
+        confident_thresholds = [np.mean(pred_probs[:, k][k_in_l[k]]) for k in range(K)]
+    else:
+        confident_thresholds = np.array(
+            [np.mean(pred_probs[:, k][labels == k]) for k in range(pred_probs.shape[1])]
+        )
     return confident_thresholds
```
jwmueller, line 1141: seems like `confident_thresholds` is not always a np.array. I'd remove the `np.array(...)` wrapper from L1138 and instead put it at the return in L1141:

```
return np.array(confident_thresholds)
```

cleanlab/count.py
```diff
@@ -425,7 +424,7 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
     # Estimate the probability thresholds for confident counting
     if thresholds is None:
         # P(we predict the given noisy label is k | given noisy label is k)
-        thresholds = [np.mean(pred_probs[:, k][labels == k]) for k in range(num_classes)]
+        thresholds = get_confident_thresholds(labels, pred_probs, multi_label=multi_label)
     thresholds = np.asarray(thresholds)
```
jwmueller, line 428: `thresholds = np.asarray(thresholds)` seems unnecessary if the `get_confident_thresholds()` properly returns a np.array like it says it does in its docstring.

cleanlab/count.py
```diff
@@ -263,16 +263,15 @@ def _compute_confident_joint_multi_label(
 
     # Compute unique number of classes K by flattening labels (list of lists)
     K = len(np.unique([i for lst in labels for i in lst]))
```
jwmueller, line 265: If we're computing num_classes based on unique values, the docstring should explicitly state that all classes must be present in labels to avoid possible user confusion like we have seen in some Issues.

cleanlab/count.py
jwmueller: overall I think it's important to be consistent more than anything else. So num_classes should be either be defined based on `pred_probs.shape[1]` or num-unique-values throughout the whole package. 

The former is more computationally efficient, but has the disadvantage that existing code would need to be changed (eg. in thresholds computation) to account for potentially missing classes not present in `labels`.

cleanlab/internal/label_quality_utils.py
```diff
@@ -35,19 +39,33 @@ def _subtract_confident_thresholds(labels: np.array, pred_probs: np.array) -> np
     pred_probs : np.array (shape (N, K))
       Predicted-probabilities in the same format expected by the `cleanlab.count.get_confident_thresholds()` method.
 
+    multi_label : bool, optional
+      multi_label in the same format expected by the `cleanlab.count.get_confident_thresholds()` method.
+
+    multi_label : bool, optional
+      If ``True``, labels should be an iterable (e.g. list) of iterables, containing a
+      list of labels for each example, instead of just a single label.
```
jwmueller, line 47: multi_label arg is documented twice

jwmueller (changes requested):
just have some minor changes/questions.

cleanlab/count.py
```diff
@@ -263,16 +263,15 @@ def _compute_confident_joint_multi_label(
...
     if thresholds is None:
         # the avg probability of class given that the label is represented.
-        thresholds = [np.mean(pred_probs[:, k][k_in_l[k]]) for k in range(K)]
+        thresholds = get_confident_thresholds(labels, pred_probs, multi_label=True)
     # Create mask for every example if for each class, prob >= threshold
     pred_probs_bool = pred_probs >= thresholds
 
     # Compute confident joint
     # (no need to avoid collisions for multi-label, double counting is okay!)
+    k_in_l = np.array([[k in lst for lst in labels] for k in range(K)])
```
jwmueller, line 274: A nitpick is this code may be 2x as slow as the previous version for dataset with tons of labels, because you are computing this `k_in_l` object twice now (whereas previously it was only computed once), the first time being inside `get_confident_thresholds`

cleanlab/count.py
jwmueller: Perf optimization if we use num-unique-values to set num-classes:  then certain helper functions could take in `num_classes` as an optional argument to avoid having to necessarily recompute: `unique(labels)` which may be costly for massive dataset.

cleanlab/count.py
suneeta-mall: I second this comment ^^. This is so true in multi-label and you are likely to run into this case give you explicitly don't ask for one hot encoded label for multi-label settings. 
I more I think about it, the more I feel, that it will be better to have a different abstraction for multi-label than trying to wedge it in via flag. At the same time, I appreciate that this support is experimental so all good! 

Left some thoughts here https://github.com/cleanlab/cleanlab/issues/263#issuecomment-1133885728 but I won't say my thoughts are well thought through just yet.

cleanlab/count.py
cgnorthcutt: done

cleanlab/count.py
cgnorthcutt: Leave as is. User could have passed in thresholds... as a list. Note this occurs after the if statement.

cleanlab/count.py
cgnorthcutt: clarified this in labels param description. Also the following tutorials PR helped address this: https://github.com/cleanlab/cleanlab/pull/279

cleanlab/internal/label_quality_utils.py
cgnorthcutt: fixed. thx @jwmueller

cleanlab/count.py
cgnorthcutt: I'm going to allow this redundancy for now. The two options to fix are:
1. compute `k_in_l` beforehand and pass it into `get_confident_thresholds`, but its a weird thing to pass in and the code is much easier to read if its computed inside of `get_confident_thresholds`
2. return `k_in_l` also when calling `get_confident_thresholds` with `multi_label=true` but that's pretty weird behavior. I can optionally include it in the return tuple with another parameter, but this seems like a somewhat silly amount of extra code to save very little compute.

the computation here should be relatively negligible so going to leave as is for the purpose of easier-to-read simpler code.

cgnorthcutt: Easy for me to implement either solution if you think its preferable. I vote to leave as is for code simplicity, but can go either way. Agree its 2x speed, but if we get to the place where this is the bottleneck than I'll be extremely happy :)

cleanlab/count.py
```diff
@@ -1108,12 +1107,31 @@ def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array
...
     -------
     confident_thresholds : np.array
       An array of shape ``(K,)``.
     """
-    confident_thresholds = np.array(
-        [np.mean(pred_probs[:, k][labels == k]) for k in range(pred_probs.shape[1])]
-    )
+
+    if multi_label:
+        # [0, 1, 2, ... num_classes - 1] Assumes all classses are represented in labels
```
jwmueller, line 1127: move this sentence to the docstring for `labels` arg?

cleanlab/count.py
```diff
@@ -1108,12 +1107,31 @@ def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array
...
+        unique_classes = range(pred_probs.shape[1])
+        # Compute thresholds = p(label=k | k in set of given labels)
+        k_in_l = np.array([[k in lst for lst in labels] for k in unique_classes])
+        # The avg probability of class given that the label is represented.
+        confident_thresholds = [np.mean(pred_probs[:, k][k_in_l[k]]) for k in unique_classes]
+    else:
+        confident_thresholds = np.array(
+            [np.mean(pred_probs[:, k][labels == k]) for k in range(pred_probs.shape[1])]
```
jwmueller, line 1135: replace `range(pred_probs.shape[1])` with `unique_classes`?

cleanlab/count.py
```diff
@@ -1108,12 +1107,31 @@ def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array
...
     confident_thresholds : np.array
       An array of shape ``(K,)``.
     """
-    confident_thresholds = np.array(
-        [np.mean(pred_probs[:, k][labels == k]) for k in range(pred_probs.shape[1])]
-    )
+
+    if multi_label:
+        # [0, 1, 2, ... num_classes - 1] Assumes all classses are represented in labels
+        unique_classes = range(pred_probs.shape[1])
```
jwmueller, line 1128: move this before the if statement?

cleanlab/count.py
jwmueller: Isn't the return type of this a list for the multilabel setting?  

Current docstring for the function says it returns np.array always

cleanlab/count.py
```diff
@@ -1098,6 +1095,8 @@ def get_confident_thresholds(labels: np.array, pred_probs: np.array) -> np.array
     labels : np.array
       An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
       Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
+      All unique labels should be present, such that:
```
jwmueller, line 1098: wouldn't "all unique labels": be present by definition?
Maybe reword to: 

"all possible classes in 0, 1,...K-1" must be present in labels"

jwmueller (approved):
LGTM, just had minor comments.

We should consider also adding the clarification that "All possible classes in 0,1,...,K-1 must be present in labels" to the docstrings of higher-level functions like find_label_issues(). 

Can do this in separate follow-up PR though.

cleanlab/benchmarking/noise_generation.py
```diff
@@ -116,7 +116,7 @@ def generate_noisy_labels(true_labels, noise_matrix):
     Parameters
     ----------
     true_labels : np.array
-      An array of shape ``(N,)`` representing perfect labels, without any
+      An array of shape ``(N, )`` representing perfect labels, without any
```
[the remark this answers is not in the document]
cgnorthcutt: this was convention at other places i've worked because it reduces edit distance if you need to add a second dimension. It's also clearer to have separation between two punctuation with different meaning.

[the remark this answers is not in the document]
cgnorthcutt: will change back to numpy convention to move things forward in agreement

PR #267, opened 2022-05-25
Add Negative Log Loss Weighting Scheme for Ensemble Label Quality Score

JohnsonKuan:
This PR implements a weighting scheme for ensemble label quality scoring which was initially proposed by @jwmueller .

This weighting scheme is enabled by setting `weight_ensemble_members_by="log_loss_search"` and optionally allows the user to pass a list of hyper-parameter values into `log_loss_search_T_values` for the search procedure.

This weighting scheme weights the scores from each model based on `exp(t * -log_loss)` where `t` is a hyper-parameter and log_loss is the log_loss between the given label and the model's prediction.

For each `t` value in `log_loss_search_T_values`, we weight each model based on `exp(t * -log_loss)`. We then take the weighted average of predictions from all models using these weights and compute the log loss between the given label and this weighted average ensemble prediction. The `t` with the lowest log loss is selected.

This PR also adds this weighting scheme to the unit tests.

### DIFF
--- a/cleanlab/rank.py
+++ b/cleanlab/rank.py
@@ -39,6 +39,7 @@ from cleanlab.internal.label_quality_utils import (
     _subtract_confident_thresholds,
     get_normalized_entropy,
 )
+from sklearn.metrics import log_loss
 
 
 def order_label_issues(
@@ -253,11 +254,12 @@ def get_label_quality_ensemble_scores(
     adjust_pred_probs : bool, optional
       `adjust_pred_probs` in the same format expected by the :py:func:`get_label_quality_scores <cleanlab.rank.get_label_quality_scores>` function.
 
-    weight_ensemble_members_by : {"uniform", "accuracy", "custom"}, default="accuracy"
+    weight_ensemble_members_by : {"uniform", "accuracy", "log_loss_search", "custom"}, default="accuracy"
       Weighting scheme used to aggregate scores from each model:
 
       - "uniform": take the simple average of scores
       - "accuracy": take weighted average of scores, weighted by model accuracy
+      - "log_loss_search": take weighted average of scores, weighted by negative log loss
       - "custom": take weighted average of scores using custom weights that the user passes to the custom_weights parameter.
 
     custom_weights : np.array, default=None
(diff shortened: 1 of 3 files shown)


cleanlab/rank.py
```diff
@@ -253,17 +255,21 @@ def get_label_quality_ensemble_scores(
...
       - "accuracy": take weighted average of scores, weighted by model accuracy
+      - "log_loss_search": take weighted average of scores, weighted by exp(t * -log_loss) where t is selected from log_loss_search_T_values parameter.
       - "custom": take weighted average of scores using custom weights that the user passes to the custom_weights parameter.
 
     custom_weights : np.array, default=None
       Weights used to aggregate scores from each model if weight_ensemble_members_by="custom".
       Length of this array must match the number of models: len(pred_probs_list).
 
+    log_loss_search_T_values : List, default=[1e-4, 1e-3, 1e-2, 1e-1, 1e0, 1e1, 1e2, 2e2]
+      List of t values used if weight_ensemble_members_by="log_loss_search".
```
jwmueller, line 271: "List of t values used" -> "List of t values considered"

Add sentence: 
We will choose the value of t that leads to weights which produce the best log-loss when used to form a weighted average of pred_probs from the models.

cleanlab/rank.py
```diff
@@ -332,14 +372,20 @@ def get_label_quality_ensemble_scores(
...
             for i, acc in enumerate(accuracy_list):
                 print(f"  Model {i} accuracy : {acc}")
                 print(f"  Model {i} weights  : {weights[i]}")
 
         # Aggregate scores with weighted average
         label_quality_scores = (scores_ensemble * weights).sum(axis=1)
 
+    elif weight_ensemble_members_by == "log_loss_search":
+        weights = neg_log_loss_weights  # Weight by exp(t * -log_loss) where t is found by searching set of T
+
```
jwmueller, line 385: also print the weights here if verbose?

More generally can probably move lines 376-381 to make them apply to both weight_ensemble_members_by = 'accuracy' and 'log_loss_search' without copy/pasting code.

cleanlab/rank.py
```diff
@@ -253,17 +255,21 @@ def get_label_quality_ensemble_scores(
     adjust_pred_probs : bool, optional
       `adjust_pred_probs` in the same format expected by the :py:func:`get_label_quality_scores <cleanlab.rank.get_label_quality_scores>` function.
 
-    weight_ensemble_members_by : {"uniform", "accuracy", "custom"}, default="accuracy"
+    weight_ensemble_members_by : {"uniform", "accuracy", "log_loss_search", "custom"}, default="accuracy"
       Weighting scheme used to aggregate scores from each model:
 
       - "uniform": take the simple average of scores
       - "accuracy": take weighted average of scores, weighted by model accuracy
+      - "log_loss_search": take weighted average of scores, weighted by exp(t * -log_loss) where t is selected from log_loss_search_T_values parameter.
```
jwmueller, line 263: and log_loss is the log-loss between a model's pred_probs and the given labels.

JohnsonKuan:
For our reference, the codecov test is failing and I believe it is due to the recently reported bug below:

https://github.com/codecov/codecov-action/issues/745

cleanlab/rank.py
JohnsonKuan: Added this sentence. Thanks.

cleanlab/rank.py
JohnsonKuan: Sounds good. Added commit to print weights for `"log_loss_search"` if verbose.

Agree, can refactor in the future to remove some redundancies.

cleanlab/rank.py
JohnsonKuan: Made this change and added the sentence. Thanks.

cleanlab/rank.py
```diff
@@ -332,10 +373,20 @@ def get_label_quality_ensemble_scores(
...
-                print(f"  Model {i} weights  : {weights[i]}")
+                print(f"  Model {i} weight   : {weights[i]}")
+
+        # Aggregate scores with weighted average
+        label_quality_scores = (scores_ensemble * weights).sum(axis=1)
+
+    elif weight_ensemble_members_by == "log_loss_search":
+        weights = neg_log_loss_weights  # Weight by exp(t * -log_loss) where t is found by searching set of T
+        if verbose:
+            print("Ensemble members will be weighted by exp(t * -log_loss)")
```
jwmueller, line 387: "weighted by exp(t * -log_loss)"
->
"weighted by log-loss between their predicted probabilities and given labels"

(The former output is intimidating for novice user who doesn't know (or need to know) what is happening internally)

JohnsonKuan: Good call. Made this change. Thanks.

cleanlab/rank.py
```diff
@@ -300,6 +306,40 @@ def get_label_quality_ensemble_scores(
             """
         )
 
+    # This weighting scheme performs search of t in log_loss_search_T_values for "best" log loss
+    if weight_ensemble_members_by == "log_loss_search":
+
```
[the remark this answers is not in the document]
JohnsonKuan: Good point on this edge case. I prefer clipping as well.

I added commit to use clipped pred_probs when calculating log loss used for the weights.

Thanks

cleanlab/rank.py
```diff
@@ -300,6 +307,40 @@ def get_label_quality_ensemble_scores(
...
+            neg_log_loss_list = []
+
+            # pred_probs for each model
+            for pred_probs in pred_probs_list:
+                neg_log_loss = np.exp(t * (-log_loss(labels, pred_probs)))
+                neg_log_loss_list.append(neg_log_loss)
+
+            # weights using negative log loss
+            neg_log_loss_weights_temp = np.array(neg_log_loss_list) / sum(neg_log_loss_list)
```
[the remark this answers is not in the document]
jwmueller: clipping seems to address this

jwmueller (approved):
LGTM!

PR #268, opened 2022-05-27
Add KNN distance OOD scoring function and unit tests

JohnsonKuan:
This PR introduces an out-of-distribution (OOD) scoring function using KNN distances.

Example of workflow:

```python
from cleanlab.rank import get_knn_distance_ood_scores

# Fit nearest neighbors on feature matrix X_train
nbrs = NearestNeighbors(n_neighbors=10, metric="cosine").fit(X_train)

# Get OOD scores for feature matrix X_test
ood_scores = get_knn_distance_ood_scores(features=X_test, nbrs=nbrs, k=3) # k can be lower than n_neighbors
```

The OOD scoring function takes as input an instantiated sklearn NearestNeighbors class object `nbrs` that's been fitted on a dataset. However, it's important to note that we can extend this to other KNN implementations (e.g. Annoy) as long as the `nbrs` can return an array of nearest neighbor distances.

Example below of workflow using [Annoy](https://github.com/spotify/annoy):

```python
from annoy import AnnoyIndex

class ApproxNearestNeighbors:

    def __init__(self, n_neighbors: int = 5, metric: str = "angular", n_trees: int = 10):
        self.ann_index = None
        self.n_neighbors = n_neighbors
        self.metric = metric
        self.n_trees = n_trees
        
    def fit(self, features: np.array):
        dim = features.shape[1]
        self.ann_index = AnnoyIndex(dim, self.metric)
        for i, x in enumerate(features):
            self.ann_index.add_item(i, x)
        self.ann_index.build(self.n_trees)
        return self
    
    def kneighbors(self, features: np.array):
    
        distances = []
        indices = []
        for x in features:
            idx, dist = self.ann_index.get_nns_by_vector(x, self.n_neighbors, include_distances=True)    
            distances.append(dist)
            indices.append(idx)
        distances = np.array(distances)
        indices = np.array(indices)
        
        return distances, indices

nbrs = ApproxNearestNeighbors(n_neighbors=5, metric="angular").fit(X_train)
ood_scores = get_knn_distance_ood_scores(features=X_test, nbrs=nbrs, k=3)
```

This PR also includes 2 unit tests.

### DIFF
--- a/cleanlab/rank.py
+++ b/cleanlab/rank.py
@@ -544,7 +544,9 @@ def get_confidence_weighted_entropy_for_each_label(
     return label_quality_scores
 
 
-def get_knn_distance_ood_scores(features: np.array, nbrs: NearestNeighbors, k: int) -> np.array:
+def get_knn_distance_ood_scores(
+    features: np.array, nbrs: NearestNeighbors, k: int = None
+) -> np.array:
     """Returns the KNN distance out-of-distribution (OOD) score for each datapoint.
 
     This is a function to compute OOD scores where higher scores indicate the datapoint is more likely to be OOD.
@@ -560,9 +562,10 @@ def get_knn_distance_ood_scores(features: np.array, nbrs: NearestNeighbors, k: i
       Note that the distance metric and n_neighbors is specified when instantiating this class.
       See: https://scikit-learn.org/stable/modules/neighbors.html
 
-    k : int
+    k : int, default=None
       Number of neighbors to use when calculating average distance to neighbors.
       This value k needs to be less than or equal to max_k which is the n_neighbors used when fitting instantiated NearestNeighbors class object.
+      If k=None, then by default k=min(10, max_k) is used where max_k is extracted from the given nbrs.
 
     Returns
     -------
@@ -572,6 +575,10 @@ def get_knn_distance_ood_scores(features: np.array, nbrs: NearestNeighbors, k: i
 
     # number of neighbors specified when fitting instantiated NearestNeighbors class object
     max_k = nbrs.n_neighbors
+
+    # if k is not provided, then use default
+    if k is None:
+        k = min(10, max_k)
 
     assert (
         k <= max_k
--- a/cleanlab/rank.py
+++ b/cleanlab/rank.py
@@ -545,7 +545,7 @@ def get_confidence_weighted_entropy_for_each_label(
 
 
 def get_knn_distance_ood_scores(
-    query_features: np.array, nbrs: NearestNeighbors, k: int
+    features: np.array, nbrs: NearestNeighbors, k: int = 10
 ) -> np.array:
     """Returns the KNN distance out-of-distribution (OOD) score for each datapoint.
 
@@ -553,8 +553,8 @@ def get_knn_distance_ood_scores(
 
     Parameters
     ----------
-    query_features : np.array
-      Feature matrix of shape ``(N, ...)``, where N is the number of datapoints.
+    features : np.array
+      Feature matrix of shape (N, M), where N is the number of datapoints and M is the number of features.
       This is the "query set" of features for each datapoint which are used for nearest neighbor search.
 
     nbrs : sklearn.neighbors.NearestNeighbors
@@ -562,7 +562,7 @@ def get_knn_distance_ood_scores(
       Note that the distance metric and n_neighbors is specified when instantiating this class.
       See: https://scikit-learn.org/stable/modules/neighbors.html
 
-    k : int
+    k : int, default=10
       Number of neighbors to use when calculating average distance to neighbors.
       This value k needs to be less than or equal to max_k which is the n_neighbors used when fitting instantiated NearestNeighbors class object.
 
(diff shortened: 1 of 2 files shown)


cleanlab/rank.py
```diff
@@ -541,3 +542,49 @@ def get_confidence_weighted_entropy_for_each_label(
...
+    query_features: np.array, nbrs: NearestNeighbors, k: int
+) -> np.array:
+    """Returns the KNN distance out-of-distribution (OOD) score for each datapoint.
+
+    This is a function to compute OOD scores where higher scores indicate the datapoint is more likely to be OOD.
+
+    Parameters
+    ----------
+    query_features : np.array
+      Feature matrix of shape ``(N, ...)``, where N is the number of datapoints.
```
jwmueller, line 557: This would work for higher than 2D data?

cleanlab/rank.py
```diff
@@ -541,3 +542,49 @@ def get_confidence_weighted_entropy_for_each_label(
...
+def get_knn_distance_ood_scores(
+    query_features: np.array, nbrs: NearestNeighbors, k: int
+) -> np.array:
+    """Returns the KNN distance out-of-distribution (OOD) score for each datapoint.
+
+    This is a function to compute OOD scores where higher scores indicate the datapoint is more likely to be OOD.
+
+    Parameters
+    ----------
+    query_features : np.array
```
jwmueller, line 556: Can we just call this arg `features` instead or `query_features`?

cleanlab/rank.py
```diff
@@ -541,3 +542,49 @@ def get_confidence_weighted_entropy_for_each_label(
...
+      Feature matrix of shape ``(N, ...)``, where N is the number of datapoints.
+      This is the "query set" of features for each datapoint which are used for nearest neighbor search.
+
+    nbrs : sklearn.neighbors.NearestNeighbors
+      Instantiated NearestNeighbors class object that's been fitted on a dataset in the same feature space.
+      Note that the distance metric and n_neighbors is specified when instantiating this class.
+      See: https://scikit-learn.org/stable/modules/neighbors.html
```
jwmueller, line 563: Is it possible to write the code such that annoy nbrs object could also be plugged in here? or would that be nontrivial because their APIs are too different?

cleanlab/rank.py
```diff
@@ -541,3 +542,49 @@ def get_confidence_weighted_entropy_for_each_label(
...
+
+    k : int
```
jwmueller, line 565: default to 10?

tests/test_rank.py
```diff
@@ -388,3 +389,48 @@ def test_unsupported_method_for_adjust_pred_probs():
...
+    X_test = data["X_test"]
+
+    # Create OOD datapoint
+    X_ood = np.array([[999999999.0, 999999999.0]])
+
+    # Add OOD datapoint to X_test
+    X_test_with_ood = np.vstack([X_test, X_ood])
+
+    # Fit nearest neighbors on X_train
+    nbrs = NearestNeighbors(n_neighbors=5, algorithm="ball_tree").fit(X_train)
```
jwmueller, line 406: why ball_tree instead of auto?

cleanlab/rank.py
```diff
@@ -541,3 +542,49 @@ def get_confidence_weighted_entropy_for_each_label(
...
+
+    This is a function to compute OOD scores where higher scores indicate the datapoint is more likely to be OOD.
+
+    Parameters
+    ----------
+    query_features : np.array
+      Feature matrix of shape ``(N, ...)``, where N is the number of datapoints.
+      This is the "query set" of features for each datapoint which are used for nearest neighbor search.
+
+    nbrs : sklearn.neighbors.NearestNeighbors
```
jwmueller, line 560: Is it possible to allow `nbrs` to be None?  (not necessarily by default)

If it is None, then we could construct the `nbrs` from the passed in `features` via sklearn.nearestneighbors inside of this function.

In particular if I just want to find OOD points in my training data, it appears I currently have to compute the neighbors twice, once when constructing `nbrs` and once again inside this function when I pass in the same data as `features`.  It would be nice if `nbrs` is None, then we only need to compute neighbors once for the `features` to construct `nbrs` and simultaneously compute their OOD score.

cleanlab/rank.py
JohnsonKuan: Good catch. I meant to put shape of (N, M).

I updated docstring with below:
`Feature matrix of shape (N, M), where N is the number of datapoints and M is the number of features.`

Thanks

cleanlab/rank.py
JohnsonKuan: Yeah, I'm cool with that. I changed it to `features`. Thanks.

tests/test_rank.py
JohnsonKuan: No good reason, I just took that from one of the sklearn docs. I'll remove for brevity so the default will be "auto".

cleanlab/rank.py
JohnsonKuan: That would be nontrivial since their APIs are too different. 

However, if we can wrap Annoy in a class that behaves like sklearn NearestNeighbors, then we can pass it as an arg.

I added an example in the PR description of the workflow with Annoy.

cleanlab/rank.py
JohnsonKuan: That's a good idea. I updated the code to use this. Thanks.

cleanlab/rank.py
JohnsonKuan: Yeah it's possible and makes sense if we want to construct `nbrs` using the `features` we pass into `get_knn_distance_ood_scores`.

In this case where we want to find OOD points in the training data: when computing the neighbor distances, we'll need to ignore the first neighbor since the nearest neighbor of each training point here is the point itself, with a distance of zero. We can perhaps add a `ignore_first_neighbor` boolean parameter if we want to ignore first neighbor when averaging distances.

jwmueller (approved):
Thanks @JohnsonKuan !!
I'll add support for `nbrs = None` option in a future PR.

PR #270, opened 2022-06-06
Dataset docstring clarifications

jwmueller:
Edits to documentation to address points of confusion raised by multiple users.

### DIFF
--- a/cleanlab/dataset.py
+++ b/cleanlab/dataset.py
@@ -64,11 +64,11 @@ def rank_classes_by_label_quality(
         perfect quality. Columns:
 
         * *Class Index*: The index of the class in 0, 1, ..., K-1.
-        * *Label Issues*: ``count(given_label = k, true_label != k)``, estimated number of label issues in the class (usually the most accurate method).
+        * *Label Issues*: ``count(given_label = k, true_label != k)``, estimated number of examples in the dataset that are labeled as class k but should have a different label.
         * *Inverse Label Issues*: ``count(given_label != k, true_label = k)``, estimated number of examples in the dataset that should actually be labeled as class k but have been given another label.
-        * *Label Noise*: ``prob(true_label != k | given_label = k)``, estimated proportion of label issues in the class. This is computed by taking the number of examples with "Label Issues" in the given class and dividing it by the total number of examples in that class.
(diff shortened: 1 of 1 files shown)


cleanlab/dataset.py
```diff
@@ -113,7 +113,7 @@ def find_overlapping_classes(
     confident_joint=None,
     multi_label=False,
 ):
-    """Returns the classes that are often confused by machine learning model or data labelers.
+    """Returns the classes that are often confused by the data labelers.
```
cgnorthcutt, line 116: @jwmueller in our language, we should not say `by the data labelers` unless we include other options. I am the one who added ML model here previously. The reason is because not all labels come from labelers. Some labels come from other ML models. other labels come for free (facebook up/down votes or any human-involved data system where human interactions are the labels -- many datasets are of this type)

cleanlab/dataset.py
```diff
@@ -144,10 +144,9 @@ def find_overlapping_classes(
     label). cleanlab takes these differences into account for you automatically via the joint
     distribution. If you do not want this behavior, simply set ``asymmetric=False``.
 
-    This method measures how often the annotators confuse two classes.
-    This method differs from just using a similarity matrix or confusion matrix. Instead, it works
-    even if the model that generated `pred_probs` in more confident in some classes than others
-    and has heterogeneity in average confidence across classes.
+    This method estimates how often the annotators confuse two classes.
+    This differs from just using a similarity matrix or confusion matrix, as these summarize characteristics of the predictive model rather than the data labelers (i.e. annotators).
```
cgnorthcutt, line 148: 100 character length limit.

cgnorthcutt (commented):
Please adhere to 100 character length limit in doc strings for readability.

https://github.com/cleanlab/cleanlab/blob/master/pyproject.toml#L2

cleanlab/dataset.py
jwmueller: To me, "confused by machine learning model" sounds exactly like confusion matrix.
Even after my attempt at some clarifying statements, I'd still guess many users will still not really understand what is difference between this and confusion matrix.

cgnorthcutt: `Returns the pairs of classes that are often mislabeled as one another.`

PR #272, opened 2022-06-07
[WIP] Label quality scores for multi-annotator

weijinglok:
This PR is still a WIP. This PR adds multi-annotator support for computing the label quality scores via the `label_quality_scores_multiannotator` function. 

TODO:
- [ ] Complete TODO for the `consensus_method` arg
- [ ] Write docstrings
- [ ] Add tests

### DIFF
--- a/cleanlab/rank.py
+++ b/cleanlab/rank.py
@@ -33,12 +33,14 @@ labels in data that was previously held-out.
 
 
 import numpy as np
-from typing import List
+from typing import List, Union, Tuple
 import warnings
 from cleanlab.internal.label_quality_utils import (
     _subtract_confident_thresholds,
     get_normalized_entropy,
 )
+from cleanlab.dataset import overall_label_health_score, rank_classes_by_label_quality
+import pandas as pd
 from sklearn.metrics import log_loss
 from sklearn.neighbors import NearestNeighbors
 
@@ -425,6 +427,133 @@ def get_label_quality_ensemble_scores(
         )
 
     return label_quality_scores
+
+
+def get_worst_class(labels, pred_probs):
+    ranked_noisy_classes = rank_classes_by_label_quality(labels, pred_probs).query(
+        "`Label Quality Score` < 1.0"
+    )
+    return ranked_noisy_classes["Class Index"][0] if len(ranked_noisy_classes) > 0 else np.nan
+
+
+def get_label_quality_scores_multiannotator(
+    labels_multiannotator: pd.DataFrame,
+    pred_probs: np.array,
+    *,
+    consensus_method: Union[str, List[str]] = "majority",
+    return_annotator_stats: bool = False,
+    kwargs: dict = {},
+) -> Union[pd.DataFrame, Tuple[pd.DataFrame, pd.DataFrame]]:
+    """Returns label quality scores for each example for each annotator.
+
+    This function computes the label quality scores for each example
+    for each annotator in a given classification dataset.
+
+    The score is between 0 and 1; lower scores
+    indicate labels less likely to be correct. For example:
+
+    1 - clean label (the given label is likely correct).
+    0 - dirty label (the given label is unlikely correct).
+
+    Parameters
+    ----------
+    labels_multiannotator : pd.DataFrame
+        2D pandas DataFrame of multiple given labels for each example with shape (N, M),
+        where N is the number of examples and M is the number of annotators.
+        labels_multiannotator[n][m] = given label for n-th example by m-th annotator.
+        For a dataset with K classes, the given labels must be an integer in 0, 1, ..., K-1 or
+        np.nan if not annotated. Column names should correspond to the annotators' ID.
+
+    pred_probs : np.array
+        An array of shape ``(N, K)`` of model-predicted probabilities,
+        ``P(label=k|x)``. Each row of this matrix corresponds
+        to an example `x` and contains the model-predicted probabilities that
+        `x` belongs to each possible class, for each of the K classes. The
+        columns must be ordered such that these probabilities correspond to
+        class 0, 1, ..., K-1.
+
+        **Caution**: `pred_probs` from your model must be out-of-sample!
+        You should never provide predictions on the same examples used to train the model,
+        as these will be overfit and unsuitable for finding label-errors.
+        To obtain out-of-sample predicted probabilities for every datapoint in your dataset, you can use :ref:`cross-validation <pred_probs_cross_val>`.
+        Alternatively it is ok if your model was trained on a separate dataset and you are only evaluating
+        data that was previously held-out.
+
+    consensus_method : str or List[str]
+        Describing which algorithm to reach the consensus label.
+        If a List is passed, then the 0th element of List is the method used to produce
+        consensus_label, quality_of_consensus, annotator_agreement (may or may not be affected by this argument).
+        The 1st, 2nd, 3rd, etc. elements of this List are output as extra columns in the returned DF with names formatted as:
+        consensus_label_SUFFIX, quality_of_consensus_SUFFIX
+        where SUFFIX = the str element of this list.
+
+    kwargs : Same additional arguments as for get_label_quality_scores().
+
+    Returns
+    -------
+    df : pandas DataFrame in which each row corresponds to one example, with columns:
+                quality_annotator_1, quality_annotator_2, ..., quality_annotator_M, annotator_agreement, consensus_label, quality_of_consensus
+            (Here _1, _2, ..., _M suffixes may be replaced column names used to describe annotators in `labels` input)
+    """
+
+    # Compute the overall label health score for each annotator (part of annotator_stats)
+    overall_label_health_score_df = labels_multiannotator.apply(
+        overall_label_health_score, args=[pred_probs], verbose=False
+    )
+
+    # Compute the number of labels labeled/annotated by each annotator (part of annotator_stats)
+    num_labeled = labels_multiannotator.count()
+
+    # Find the worst labeled class for each annotator (part of annotator_stats)
+    worst_class = labels_multiannotator.apply(get_worst_class, args=[pred_probs])
+
+    # Create annotator_stats DataFrame from its columns
+    annotator_stats = pd.DataFrame(
+        {
+            "overall_quality": overall_label_health_score_df,
+            "num_labeled": num_labeled,
+            "worst_class": worst_class,
+        }
+    )
+
+    # Compute the label quality scores for each annotators' labels
+    label_quality_scores_multiannotator = labels_multiannotator.apply(
+        get_label_quality_scores, args=[pred_probs], **kwargs
+    )
+
+    # Prefix column name referring to the annotators' label quality scores
+    label_quality_scores_multiannotator = label_quality_scores_multiannotator.add_prefix(
+        "quality_of_"
+    )
+
+    # Compute the consensus_labels
+    # TODO: deal with ties, conditional based on consensus_method, consensus_method can be a List[str], add dawid-skene
+    consensus_labels = labels_multiannotator.mode(axis=1)
+
+    # Compute the fraction of annotator disagreeing with the consensus labels
+    # i.e. 1 - fraction of annotator agreeing with the consensus labels
+    annotator_disagreement = labels_multiannotator.assign(consensus_label=consensus_labels).apply(
+        lambda s: 1.0 - s[:-1].value_counts(normalize=True)[s["consensus_label"]],
+        axis=1,
+    )
+
+    # Compute the label quality scores of the consensus labels
+    quality_of_consensus = get_label_quality_scores(
+        consensus_labels.to_numpy(), pred_probs, **kwargs
+    )
+
+    # Concatenate additional columns to the label_quality_scores_multiannotator DataFrame
+    (
+        label_quality_scores_multiannotator["consensus_label"],
+        label_quality_scores_multiannotator["annotator_disagreement"],
+        label_quality_scores_multiannotator["quality_of_consensus"],
+    ) = (consensus_labels, annotator_disagreement, quality_of_consensus)
+
+    return (
+        (label_quality_scores_multiannotator, annotator_stats)
+        if return_annotator_stats
+        else label_quality_scores_multiannotator
+    )
 
 
 def get_self_confidence_for_each_label(
--- a/cleanlab/rank.py
+++ b/cleanlab/rank.py
@@ -33,14 +33,12 @@ labels in data that was previously held-out.
 
 
 import numpy as np
-from typing import List, Union, Tuple
+from typing import List
 import warnings
 from cleanlab.internal.label_quality_utils import (
     _subtract_confident_thresholds,
     get_normalized_entropy,
 )
-from cleanlab.dataset import overall_label_health_score, rank_classes_by_label_quality
-import pandas as pd
 from sklearn.metrics import log_loss
 from sklearn.neighbors import NearestNeighbors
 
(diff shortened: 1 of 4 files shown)


cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+    Returns
+    -------
+    df : pandas DataFrame in which each row corresponds to one example, with columns:
+                quality_annotator_1, quality_annotator_2, ..., quality_annotator_M, annotator_agreement, consensus_label, quality_of_consensus
+            (Here _1, _2, ..., _M suffixes may be replaced column names used to describe annotators in `labels` input)
+    """
+
+    # Compute the overall label health score for each annotator (part of annotator_stats)
+    overall_label_health_score_df = labels_multiannotator.apply(
+        overall_label_health_score, args=[pred_probs], verbose=False
```
jwmueller, line 501: for these `apply()` calls, should `args = (pred_probs,)` as tuple instead or does not it not matter?

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
     return label_quality_scores
 
 
+def get_worst_class(labels, pred_probs):
+    ranked_noisy_classes = rank_classes_by_label_quality(labels, pred_probs).query(
+        "`Label Quality Score` < 1.0"
+    )
+    return ranked_noisy_classes["Class Index"][0] if len(ranked_noisy_classes) > 0 else np.nan
```
jwmueller, line 436: This function seems like it belongs in dataset.py?  And should it be user facing or not?  If not, should have underscore or live in utils instead

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+
+
+def get_label_quality_scores_multiannotator(
```
jwmueller, line 439: Let's move all multiannotator stuff to a new file named something like `multiannotator.py` ?

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+
+    consensus_method : str or List[str]
+        Describing which algorithm to reach the consensus label.
+        If a List is passed, then the 0th element of List is the method used to produce
+        consensus_label, quality_of_consensus, annotator_agreement (may or may not be affected by this argument).
+        The 1st, 2nd, 3rd, etc. elements of this List are output as extra columns in the returned DF with names formatted as:
+        consensus_label_SUFFIX, quality_of_consensus_SUFFIX
+        where SUFFIX = the str element of this list.
+
+    kwargs : Same additional arguments as for get_label_quality_scores().
```
jwmueller, line 490: Keyword arguments to pass into ``get_label_quality_scores()``.

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+    0 - dirty label (the given label is unlikely correct).
+
+    Parameters
+    ----------
+    labels_multiannotator : pd.DataFrame
+        2D pandas DataFrame of multiple given labels for each example with shape (N, M),
+        where N is the number of examples and M is the number of annotators.
+        labels_multiannotator[n][m] = given label for n-th example by m-th annotator.
+        For a dataset with K classes, the given labels must be an integer in 0, 1, ..., K-1 or
+        np.nan if not annotated. Column names should correspond to the annotators' ID.
```
jwmueller, line 465: "if not annotated" -> "if this annotator did not label a particular example"

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+
+        **Caution**: `pred_probs` from your model must be out-of-sample!
+        You should never provide predictions on the same examples used to train the model,
+        as these will be overfit and unsuitable for finding label-errors.
+        To obtain out-of-sample predicted probabilities for every datapoint in your dataset, you can use :ref:`cross-validation <pred_probs_cross_val>`.
+        Alternatively it is ok if your model was trained on a separate dataset and you are only evaluating
+        data that was previously held-out.
+
+    consensus_method : str or List[str]
+        Describing which algorithm to reach the consensus label.
```
jwmueller, line 483: "Describing which algorithm to reach the consensus label"
->
"For each example, which method should be used to aggregate labels from multiple annotators into a single consensus label".

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+        If a List is passed, then the 0th element of List is the method used to produce
```
jwmueller, line 484: Add extra sentence before this:

A List may be passed if you want to consider multiple methods for producing consensus labels.
If a List is passed, then the 0th element of List is the method used to produce columns "consensus_label", "quality_of_consensus", "annotator_agreement" in the returned DataFrame.

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+        consensus_label, quality_of_consensus, annotator_agreement (may or may not be affected by this argument).
+        The 1st, 2nd, 3rd, etc. elements of this List are output as extra columns in the returned DF with names formatted as:
```
jwmueller, line 486: [nit] "DF", try not to use abbreviations in documentation

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+        consensus_label_SUFFIX, quality_of_consensus_SUFFIX
+        where SUFFIX = the str element of this list.
```
jwmueller, line 488: "of this list." ->
"of this list, which must correspond to a valid method for computing consensus labels".

cleanlab/rank.py
jwmueller: Options include: ["majority"]

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+
+    kwargs : Same additional arguments as for get_label_quality_scores().
+
+    Returns
+    -------
+    df : pandas DataFrame in which each row corresponds to one example, with columns:
+                quality_annotator_1, quality_annotator_2, ..., quality_annotator_M, annotator_agreement, consensus_label, quality_of_consensus
+            (Here _1, _2, ..., _M suffixes may be replaced column names used to describe annotators in `labels` input)
```
jwmueller, line 496: "column names used to describe annotators in `labels` input" ->
"column names in `labels` DataFrame used to ID the annotators"

jwmueller: [nit] don't need the parentheses

jwmueller: Add definitions:

"annotator_agreement" is the fraction of annotators that agree with the label chosen by the majority of the annotators (only considering those annotators that labeled a particular example).

"consensus_label" is the single label that is best for each example (you can control how it is derived from all annotators' labels via the argument: ``consensus_method``).

"quality_of_consensus" is cleanlab's label quality score that quantifies how likely the consensus_label is correct.

jwmueller: "quality_annotator_M" is the label quality score for the labels provided by annotator M (is nan for examples which this annotator did not label).

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+        consensus_label, quality_of_consensus, annotator_agreement (may or may not be affected by this argument).
+        The 1st, 2nd, 3rd, etc. elements of this List are output as extra columns in the returned DF with names formatted as:
+        consensus_label_SUFFIX, quality_of_consensus_SUFFIX
+        where SUFFIX = the str element of this list.
+
+    kwargs : Same additional arguments as for get_label_quality_scores().
+
+    Returns
+    -------
+    df : pandas DataFrame in which each row corresponds to one example, with columns:
```
jwmueller, line 494: Can we use more informative variable name like: `label_quality_df`

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+                quality_annotator_1, quality_annotator_2, ..., quality_annotator_M, annotator_agreement, consensus_label, quality_of_consensus
+            (Here _1, _2, ..., _M suffixes may be replaced column names used to describe annotators in `labels` input)
+    """
```
jwmueller, line 497: Sorry didn't specify this before, but could be nice to also add column in the returned DF:

`num_annotations` which is the number of annotators that have labeled each example.

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+    )
+    return ranked_noisy_classes["Class Index"][0] if len(ranked_noisy_classes) > 0 else np.nan
+
+
+def get_label_quality_scores_multiannotator(
+    labels_multiannotator: pd.DataFrame,
+    pred_probs: np.array,
+    *,
+    consensus_method: Union[str, List[str]] = "majority",
+    return_annotator_stats: bool = False,
```
jwmueller, line 444: missing documentation for this argument

jwmueller: Returned annotator_stats DF should also be documented

jwmueller: Let's add another arg here:
`verbose = True`

If verbose: then the full `annotator_stats` DF is printed out during the execution of this function.

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+    )
+
+    # Compute the number of labels labeled/annotated by each annotator (part of annotator_stats)
+    num_labeled = labels_multiannotator.count()
+
+    # Find the worst labeled class for each annotator (part of annotator_stats)
+    worst_class = labels_multiannotator.apply(get_worst_class, args=[pred_probs])
+
+    # Create annotator_stats DataFrame from its columns
+    annotator_stats = pd.DataFrame(
```
jwmueller, line 511: Let's move the code that computes `annotator_stats` into a separate function `compute_annotator_stats()` and just call it inside this one. 
Furthermore, you should only call `compute_annotator_stats` if `verbose == True` or if `return_annotator_stats == True`.

Let's only call the `compute_annotator_stats()` function at the very end of this one.

jwmueller: Your current code to compute the annotator_stats is a bit off because it does not consider the consensus_label at all.

The `compute_annotator_stats()` function should take as an extra input the `df` returned by this function. 
It should be report the annotator_stats based on the df returned by this function instead of recomputing them from scratch.  The df already contains the quality_score for each example from each annotator so you can easily compute these statistics from it.

jwmueller: `annotator_stats` should contain another column:

"agreement_with_consensus" which equals the fraction of examples (out of those labeled by this annotator and at least 2 other annotators) where this annotator did not agree with the consensus label. I think you could omit tied consensus labels from this calculation.

cleanlab/rank.py
jwmueller: Define each option for reaching consensus label:

"majority" means consensus labels are reached via a simple majority vote among annotators, with ties broken via ??

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+        get_label_quality_scores, args=[pred_probs], **kwargs
+    )
+
+    # Prefix column name referring to the annotators' label quality scores
+    label_quality_scores_multiannotator = label_quality_scores_multiannotator.add_prefix(
+        "quality_of_"
+    )
+
+    # Compute the consensus_labels
+    # TODO: deal with ties, conditional based on consensus_method, consensus_method can be a List[str], add dawid-skene
```
jwmueller, line 530: Let's break ties randomly for now.
So if there are 4 labels with an equal number of annotators voting for each one (and no more common label), then we'd just choose one of these 4 at random.

cleanlab/rank.py
```diff
@@ -33,12 +33,14 @@
...
 import numpy as np
-from typing import List
+from typing import List, Union, Tuple
 import warnings
 from cleanlab.internal.label_quality_utils import (
     _subtract_confident_thresholds,
     get_normalized_entropy,
 )
+from cleanlab.dataset import overall_label_health_score, rank_classes_by_label_quality
+import pandas as pd
```
jwmueller, line 43: Can you update description of the PR to show a code example for using this new functionality, as well as the output it prints and what the returned objects look like?

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+    pred_probs: np.array,
+    *,
+    consensus_method: Union[str, List[str]] = "majority",
+    return_annotator_stats: bool = False,
+    kwargs: dict = {},
+) -> Union[pd.DataFrame, Tuple[pd.DataFrame, pd.DataFrame]]:
+    """Returns label quality scores for each example for each annotator.
+
+    This function computes the label quality scores for each example
+    for each annotator in a given classification dataset.
```
jwmueller, line 450: Replace first 2 sentences with something like:

"This function is for multiclass classification datasets where examples have been labeled by multiple annotators (not necessarily the same number of annotators per example).  It computes quality scores for each annotator's labels and other useful values like how confident we are that the current consensus label is actually correct."

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+    # Compute the consensus_labels
+    # TODO: deal with ties, conditional based on consensus_method, consensus_method can be a List[str], add dawid-skene
+    consensus_labels = labels_multiannotator.mode(axis=1)
+
+    # Compute the fraction of annotator disagreeing with the consensus labels
+    # i.e. 1 - fraction of annotator agreeing with the consensus labels
+    annotator_disagreement = labels_multiannotator.assign(consensus_label=consensus_labels).apply(
+        lambda s: 1.0 - s[:-1].value_counts(normalize=True)[s["consensus_label"]],
+        axis=1,
+    )
```
jwmueller, line 538: Is this most efficient? 
For a particular row of  `labels_multiannotator`, don't we only need to compute: 
`np.mean(s != s["consensus_label"])`

I don't think we have to compute the value_counts here right?

jwmueller (changes requested):
Great start so far!

cleanlab/rank.py
jwmueller: The code should hardcode the available options for `consensus_method` and explicitly check that user has chosen a valid option.

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+
+    The score is between 0 and 1; lower scores
+    indicate labels less likely to be correct. For example:
+
+    1 - clean label (the given label is likely correct).
+    0 - dirty label (the given label is unlikely correct).
+
+    Parameters
+    ----------
+    labels_multiannotator : pd.DataFrame
```
jwmueller, line 460: Code should check that there are no rows with all np.nan in this DataFrame.
Raise ValueError otherwise.

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+
+    # Compute the label quality scores of the consensus labels
+    quality_of_consensus = get_label_quality_scores(
+        consensus_labels.to_numpy(), pred_probs, **kwargs
+    )
+
+    # Concatenate additional columns to the label_quality_scores_multiannotator DataFrame
+    (
+        label_quality_scores_multiannotator["consensus_label"],
+        label_quality_scores_multiannotator["annotator_disagreement"],
```
jwmueller, line 548: Code should raise Warning if there are no examples with 2 or more annotators. Warning should state that user may prefer to simply run `get_label_quality_scores()` on the provided label for each example and the additional functionality of this method will likely not be useful.

jwmueller: Code should raise Warning if there are no examples with 2 or more annotators where all annotators agree. It is indicative of something weird going on in how the user has formatted their inputs.

cleanlab/rank.py
```diff
@@ -427,6 +429,133 @@ def get_label_quality_ensemble_scores(
...
+        label_quality_scores_multiannotator["quality_of_consensus"],
+    ) = (consensus_labels, annotator_disagreement, quality_of_consensus)
+
+    return (
+        (label_quality_scores_multiannotator, annotator_stats)
+        if return_annotator_stats
+        else label_quality_scores_multiannotator
+    )
+
+
```
jwmueller, line 558: Remember to add unit test where some annotators never chose a particular class. This is an edge case that may have bugs

jwmueller: Remember to add unit test where some examples only have one annotator.

jwmueller: Remember to add unit test with ties.

cleanlab/rank.py
weijinglok: done!

cleanlab/rank.py
weijinglok: done!

cleanlab/rank.py
weijinglok: done!

cleanlab/rank.py
weijinglok: done!

cleanlab/rank.py
weijinglok: done!

cleanlab/rank.py
weijinglok: have refactored the `compute_annotator_stats`, but have not added the `agreement_with_consensus` column

cleanlab/rank.py
weijinglok: have broke ties by picking the one with the highest `pred_probs`

cleanlab/rank.py
weijinglok: done!

cleanlab/rank.py
weijinglok: done! (good catch!)

cleanlab/rank.py
weijinglok: done!

PR #279, opened 2022-06-16
Provide example label formats in tutorials

jwmueller:
Provide concrete example of what labels might look like in all tutorials, to address common point of user confusion.

### DIFF
--- a/docs/source/tutorials/audio.ipynb
+++ b/docs/source/tutorials/audio.ipynb
@@ -495,7 +495,7 @@
     "id": "laui-jXMm6qR"
    },
    "source": [
-    "Based on the given labels and out-of-sample predicted probabilities, cleanlab can quickly help us identify label issues in our dataset. Here we request that the indices of the identified label issues should be sorted by cleanlab's _self-confidence_ score, which measures the quality of each given label via the probability assigned it in our model's prediction."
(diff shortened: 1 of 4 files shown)


docs/source/tutorials/audio.ipynb
```diff
@@ -346,6 +346,7 @@
     "df = pd.DataFrame(file_paths, columns=[\"wav_audio_file_path\"])\n",
     "df[\"label\"] = df.wav_audio_file_path.map(lambda x: int(Path(x).parts[-1].split(\"_\")[0]))\n",
     "# Note: Classes must be represented as integer indices 0, 1, ..., num_classes - 1\n",
+    "# Eg. for dataset with 7 examples from 3 classes, labels might be: np.array([2,0,0,1,2,0,1])\n",
```
anishathalye, line 349: labels just needs to be an array_like, so we could simplify this a tiny bit, omit the `np.array` part, and write just `[2,0,0,1,2,0,1]`

anishathalye (approved):

docs/source/tutorials/audio.ipynb
jwmueller: isn't numpy array preferred for efficiency?  Example should show most preferable input format right?

anishathalye: I doubt it affects efficiency much. Showing the np.array version seems fine.

jwmueller: I meant efficiency of users' overall code beyond cleanlab because they probably already have labels in pandas/numpy/tensor format

PR #282, opened 2022-06-20
Clarify labels format

cgnorthcutt:
* Simple PR that adds descriptions of the `labels` parameter for all major API functions. The following description is added (except for rank.py which does not support multi_label currently).

```
All the classes (0, 1, ..., and K-1) MUST be present in ``labels``, such that:
len(set(labels)) == pred_probs.shape[1] for single-labeled labels, e.g. [1,0,2,1,1,0...]
len(set(k for l in labels for k in l)) == pred_probs.shape[1] for multi_label, e.g. [[1,2],[1],[0],..]
```

### DIFF
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -126,6 +126,9 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False):
     labels : np.array
       A discrete vector of noisy labels, i.e. some labels may be erroneous.
       *Format requirements*: for dataset with K classes, labels must be in 0, 1, ..., K-1.
+      All the classes (0, 1, ..., and K-1) MUST be present in ``labels``, such that:
+      len(set(labels)) == pred_probs.shape[1] for single-labeled labels, e.g. [1,0,2,1,1,0...]
+      len(set(k for l in labels for k in l)) == pred_probs.shape[1] for multi_label, e.g. [[1,2],[1],[0],..]
 
     multi_label : bool, optional
       If ``True``, labels should be an iterable (e.g. list) of iterables, containing a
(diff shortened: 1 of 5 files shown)


cgnorthcutt:
@anishathalye @jwmueller  -- i could add a warning whenever this is not satisfied?

@jwmueller -- unless you want to support the subset of data case where pyx has more labels than those passed-in, in which case the warning would be confusing. This is not a very common use case, though and the warning would be helpful for 95%+ of users.

cleanlab/count.py
```diff
@@ -126,6 +126,9 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False):
     labels : np.array
       A discrete vector of noisy labels, i.e. some labels may be erroneous.
       *Format requirements*: for dataset with K classes, labels must be in 0, 1, ..., K-1.
+      All the classes (0, 1, ..., and K-1) MUST be present in ``labels``, such that:
+      len(set(labels)) == pred_probs.shape[1] for single-labeled labels, e.g. [1,0,2,1,1,0...]
+      len(set(k for l in labels for k in l)) == pred_probs.shape[1] for multi_label, e.g. [[1,2],[1],[0],..]
```
jwmueller, line 131: I don't think this will render well in the docs?

cgnorthcutt: right, i can add double ``

cleanlab/rank.py
```diff
@@ -123,6 +123,8 @@ def get_label_quality_scores(
     labels : np.array
       A discrete vector of noisy labels, i.e. some labels may be erroneous.
       *Format requirements*: for dataset with K classes, labels must be in 0, 1, ..., K-1.
+      All the classes (0, 1, ..., and K-1) MUST be present in ``labels``, such that: len(set(labels)) == pred_probs.shape[1]
+      Assumes single-labeled labels, e.g. format: np.array([1,0,2,1,1,0...])
```
jwmueller: "single-labeled labels" may confuse some people who think they need to know what this means.

Could instead say:

Note: multi-label classification is not supported by this method, each example must belong to a single class, e.g. format: ``labels = np.array([1,0,2,1,1,0...])``.

PR #286, opened 2022-06-21
Drop dependency on SciPy

anishathalye:
We aren't actually using SciPy within the cleanlab package, only in tests.


anishathalye:
Easy PR 🙃

cgnorthcutt (approved):
scipy is our largest dependency so this change significantly reduces install time on fresh environments/machines.

LGTM!

PR #294, opened 2022-06-28
Created FAQ Page in the Cleanlab documentation

ulya-tkch:
- Created base FAQ page in cleanlab documentation

### DIFF
--- a/docs/source/tutorials/faq.ipynb
+++ b/docs/source/tutorials/faq.ipynb
@@ -2,17 +2,17 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "35694031",
+   "id": "ffe0d62e",
    "metadata": {},
    "source": [
     "# FAQ\n",
     "\n",
-    "Answers to frequently asked questions about the [`cleanlab`](https://github.com/cleanlab/cleanlab) open source package"
+    "Answers to frequently asked questions about the [cleanlab](https://github.com/cleanlab/cleanlab) open source package"
    ]
   },
   {
    "cell_type": "markdown",
-   "id": "9a45259e",
+   "id": "eca36874",
    "metadata": {},
    "source": [
     "### 1.How do I format labels for Cleanlab?"
@@ -20,7 +20,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "28dccc59",
+   "id": "d5d0fbb3",
    "metadata": {},
    "source": [
     "Cleanlab only works with integer-encoded labels in the range `{0,1, ... K-1}` where `K = number_of_classes`. Final labels array should only contain integer values in the range  `{0, K-1}` and be of shape `(N,)` where `N = total_number_of_data_points`.\n",
@@ -32,8 +32,8 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
-   "id": "b9bd950e",
+   "execution_count": 8,
+   "id": "239d5ee7",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -46,7 +46,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "963fd8f0",
+   "id": "49c3927b",
    "metadata": {},
    "source": [
     "Here `arr` is the one-hot encoded array and `one_hot_encoded_axis` is the axis of the one-hot encoded feature columns. `int_arr` is the final integer-encoded array. After labels array is integer-encoded, see in [The Workflows of Data-Centric AI Tutorial](https://docs.cleanlab.ai/v2.0.0/tutorials/indepth_overview.html#Workflow-1:-Use-CleanLearning()-for-everything) how to pass it into the model. "
@@ -54,7 +54,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "7f1a7829",
+   "id": "1520a93f",
    "metadata": {},
    "source": [
     "### Can't find an answer to your question?\n",
--- a/docs/source/tutorials/faq.ipynb
+++ b/docs/source/tutorials/faq.ipynb
@@ -23,9 +23,9 @@
    "id": "d5d0fbb3",
    "metadata": {},
    "source": [
-    "Cleanlab only works with integer-encoded labels in the range `{0,1, ... K-1}` where `K = number_of_classes`. Final labels array should only contain integer values in the range  `{0, K-1}` and be of shape `(N,)` where `N = total_number_of_data_points`.\n",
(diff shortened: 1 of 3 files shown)


docs/source/tutorials/faq.ipynb
```diff
@@ -0,0 +1,89 @@
...
+    "\n",
+    "Answers to frequently asked questions about the [cleanlab](https://github.com/cleanlab/cleanlab) open source package"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "eca36874",
+   "metadata": {},
+   "source": [
+    "### 1.How do I format labels for Cleanlab?"
```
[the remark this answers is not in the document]
ulya-tkch: changed to no numbers!

docs/source/tutorials/faq.ipynb
```diff
@@ -0,0 +1,89 @@
...
+   "source": [
+    "### How do I format labels for Cleanlab?"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d5d0fbb3",
+   "metadata": {},
+   "source": [
+    "Cleanlab only works with integer-encoded labels in the range `{0,1, ... K-1}` where `K = number_of_classes`. Final labels array should only contain integer values in the range  `{0, K-1}` and be of shape `(N,)` where `N = total_number_of_data_points`.\n",
```
jwmueller, line 26: Final labels array
-> 
The ``labels`` array

docs/source/tutorials/faq.ipynb
```diff
@@ -0,0 +1,89 @@
...
+    "\n",
+    "**Text labels** should to be integer-encoded. An example of text to integer-encoding can be found in step 2 of [Text Classification with TensorFlow, Keras, and Cleanlab Tutorial](https://docs.cleanlab.ai/v2.0.0/tutorials/text.html#2.-Load-and-preprocess-the-IMDb-text-dataset). \n",
```
jwmueller, line 28: First sentence can be improved as:

**Text or string labels** should be mapped to integers for each possible value.

jwmueller: Just FYI:  links in the documentation should not be for specific version of cleanlab.  In this case the link should've been: have to work for every single version of the documentation, so there should never be a hardcoded global link to another internal page in the documentation.  In this case, the link should've simply been:  https://docs.cleanlab.ai/stable/tutorials/text.html#2.-Load-and-preprocess-the-IMDb-text-dataset

jwmueller: But actually in this case specifically, I would not use a link since the concept is so simple to explain. I would instead just say:

For example if your original data labels look like this: `["dog", "dog", "cat", "mouse", "cat"]`, you should feed them to Cleanlab like this: `labels = [1, 1, 0, 2, 0]` and keep track of which integer represents which class (classes were ordered alphabetically in this example).

docs/source/tutorials/faq.ipynb
```diff
@@ -0,0 +1,89 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "239d5ee7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np \n",
+    "\n",
+    "arr  = np.array([[0,1,0,0],[1,0,0,0],[0,0,0,1]]) # [3,4] array wih 4 labels\n",
```
jwmueller, line 42: Comment should clarify: 

This example has 3 labels (one per datapoint) where each label can be one of 4 possible classes.

jwmueller: I'd be wary of using 3 examples though and 4 classes, since cleanlab is not intended to be used with more classes than data, and there will coincidentally happen to be two instances of the number 3 (one for # examples, one for max-class-integer).

I'd instead consider changing this example to something like 4 examples, 3 classes.

docs/source/tutorials/faq.ipynb
```diff
@@ -0,0 +1,89 @@
...
+    "one_hot_encoded_axis = 1\n",
```
jwmueller, line 43: can probably delete this variable and just set `axis=1` below. We will have many Qs on the FAQ page eventually and want to keep answers short

docs/source/tutorials/faq.ipynb
```diff
@@ -0,0 +1,89 @@
...
+    "int_arr = np.argmax(arr, axis=one_hot_encoded_axis) # [3, ] array with label 0-3 assigned per row "
```
jwmueller, line 44: call this variable `labels`. Comment can clarify this is how labels should be represented when input into cleanlab functions.

docs/source/tutorials/faq.ipynb
```diff
@@ -0,0 +1,89 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "49c3927b",
+   "metadata": {},
+   "source": [
+    "Here `arr` is the one-hot encoded array and `one_hot_encoded_axis` is the axis of the one-hot encoded feature columns. `int_arr` is the final integer-encoded array. After labels array is integer-encoded, see in [The Workflows of Data-Centric AI Tutorial](https://docs.cleanlab.ai/v2.0.0/tutorials/indepth_overview.html#Workflow-1:-Use-CleanLearning()-for-everything) how to pass it into the model. "
```
jwmueller, line 52: I think we can omit this cell. It should be pretty obvious from the above how user can utilize `labels`. The FAQ answers are not intended to be full tutorials, just to help somebody get unstuck (and for cleanlab team members to be able to quickly point to a link when users ask questions!)

docs/source/tutorials/faq.ipynb
```diff
@@ -0,0 +1,89 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1520a93f",
+   "metadata": {},
+   "source": [
+    "### Can't find an answer to your question?\n",
+    "\n",
+    "If your issue is not addressed here, please refer to [Cleanlab Github issues](https://github.com/cleanlab/cleanlab/issues). \n",
```
jwmueller, line 62: I'd use this link instead:  https://github.com/cleanlab/cleanlab/issues?q=is%3Aissue

so that closed issues are also shown which may contain many answers

jwmueller: Also refer the person to our examples repo:
https://github.com/cleanlab/examples

as another source of answers

jwmueller: Finally you can also direct them to our slack community:
https://join.slack.com/t/cleanlab-community/shared_invite/zt-17lszn4hv-gg2FhZPXYfljq_l01uo92g

PR #295, opened 2022-06-28
Allow CleanLearning to use validation data in each fold

huiwengoh:
Added an arg in CleanLearning.fit() which allows users to pass in a callable function to map the validation data into a format that is accepted by `clf`'s `fit()` method.

### DIFF
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -372,9 +372,20 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
           than during cross-validation.
 
         validation_func : callable, optional
-          Optional callable function that takes two arguments, `X_val`, `y_val`.
-          Specifies how to map the validation data split in the cross validation
-          step into the appropriate format to pass into `clf`'s ``fit()`` method.
+          Optional callable function that takes two arguments, `X_val`, `y_val`, and returns a dict
+          of keyword arguments passed into to `clf.fit()` which may be functions of the validation
+          data in each cross-validation fold. Specifies how to map the validation data split in each
+          cross-validation fold into the appropriate format to pass into `clf`'s ``fit()`` method.
+          e.g. if your model's ``fit()`` method is call using `clf.fit(X, y, X_validation, y_validation)`,
+          then you could set `validation_func = f` where
+          `def f(X_val, y_val): return {"X_validation": X_val, "y_validation": y_val}`
+
+          Note that `validation_func` will be ignored in the final call to `clf.fit()` on the
+          cleaned subset of the data. This argument is only for allowing `clf` to access the
+          validation data in each cross-validation fold (eg. for early-stopping or hyperparameter-selection
+          purposes). If you want to pass in validation data even in the final training of `clf.fit()`
+          on the cleaned data subset, you should explicitly pass in that data yourself
+          (eg. via `clf_final_kwargs` or `clf_kwargs`).
 
         Returns
         -------
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -272,6 +272,7 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
         sample_weight=None,
         clf_kwargs={},
         clf_final_kwargs={},
+        validation_func=None,
     ):
         """
         Train the model `clf` with error-prone, noisy labels as if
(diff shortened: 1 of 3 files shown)


cleanlab/classification.py
```diff
@@ -370,6 +371,11 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
           This can be useful for training differently in the final ``fit()``
           than during cross-validation.
 
+        validation_func : callable, optional
+          Optional callable function that takes two arguments, `X_val`, `y_val`.
+          Specifies how to map the validation data split in the cross validation
+          step into the appropriate format to pass into `clf`'s ``fit()`` method.
+
```
jwmueller, line 378: provide a simple example?  

Example: If your model's fit function looks like this: ``clf.fit(X,y,X_validation,y_validation)``, then you could set: ``validation_func = f``, with: ``def f(X_val, y_val): return {"X_validation": X_val, "y_validation": y_val}``

cleanlab/classification.py
```diff
@@ -370,6 +371,11 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
           This can be useful for training differently in the final ``fit()``
           than during cross-validation.
 
+        validation_func : callable, optional
+          Optional callable function that takes two arguments, `X_val`, `y_val`.
```
jwmueller, line 375: replace "." with:

, and returns a dict of keyword arguments passed into to `clf.fit()` which may be functions of the validation data in each cross-validation fold.

cleanlab/classification.py
jwmueller: Add final statement to docstring:

Note that ``validation_func`` will be ignored in the final call to `clf.fit()` on the cleaned subset of the data. This argument is only for allowing `clf` to access the validation data in each cross-validation fold (eg. for early-stopping or hyperparameter-selection purposes). If you want to pass in validation data even in the final training of `clf.fit()` on the cleaned data subset, you should explicitly pass in that data yourself (eg. via ``clf_final_kwargs`` or ``clf_kwargs``).

cleanlab/count.py
```diff
@@ -764,8 +769,14 @@ def estimate_confident_joint_and_cv_pred_proba(
                 )
                 missing_class_inds[missing_class] = dup_idx
 
+        # Map validation data into appropriate format to pass into classifier clf
+        if validation_func is None:
+            validation_kwargs = {}
+        elif callable(validation_func):
+            validation_kwargs = validation_func(X_holdout_cv, s_holdout_cv)
+
```
jwmueller, line 777: normally would not require type-based error checking, but we need to do something here since you're using if/elif structure and there is no "else" statement. 

I'd just add:
```
else:
    raise TypeError("validation_func must be callable function with args: X_val, y_val")
```

cleanlab/count.py
```diff
@@ -862,6 +874,11 @@ def estimate_py_noise_matrices_and_cv_pred_proba(
     clf_kwargs : dict, optional
       Optional keyword arguments to pass into `clf`'s ``fit()`` method.
 
+    validation_func : callable, optional
+      Optional callable function that takes two arguments, `X_val`, `y_val`.
+      Specifies how to map the validation data split in the cross validation
+      step into the appropriate format to pass into `clf`'s ``fit()`` method.
+
```
jwmueller, line 881: Since I suggested many additions to the docstring for this in CleanLearning.fit(), here I'd now just say:

```
validation_func : callable, optional
    Specifies how to map the validation data split in cross-validation as input for ``clf.fit()``.  For details, see the documentation of :py:meth:`CleanLearning.fit<cleanlab.classification.CleanLearning.fit>`

```

cleanlab/count.py
```diff
@@ -928,6 +947,11 @@ def estimate_cv_predicted_probabilities(
     clf_kwargs : dict, optional
       Optional keyword arguments to pass into `clf`'s ``fit()`` method.
 
+    validation_func : callable, optional
+      Optional callable function that takes two arguments, `X_val`, `y_val`.
```
jwmueller, line 951: Would again replace docstring here with the shorter snippet above that just tells user to reference CleanLearning.fit

cleanlab/count.py
```diff
@@ -1008,6 +1034,11 @@ def estimate_noise_matrices(
...
```
jwmueller, line 1038: Would again replace docstring here with the shorter snippet above that just tells user to reference CleanLearning.fit

cleanlab/classification.py
huiwengoh: done!

cleanlab/classification.py
huiwengoh: done!

cleanlab/count.py
huiwengoh: done!

cleanlab/count.py
huiwengoh: done!

cleanlab/count.py
huiwengoh: done!

cleanlab/count.py
huiwengoh: done!

jwmueller (approved):
Great stuff!

PR #301, opened 2022-06-29
Proper validation of `labels` values/format across package

jwmueller:
Applies same function to check the labels and pred_probs objects across all main user-functions in the package.
Other changes:

* Fixed bugs in: label-checking code and unit tests.

* Refactored code to be better structured.

* Updated some pandas calls to be forwards-compatible.


jwmueller:
Thanks @huiwengoh for pointing out one of the existing bugs!

huiwengoh (commented):
Thanks for making the change! I think it looks good!

cleanlab/filter.py
```diff
@@ -735,7 +510,7 @@ class 0, 1, ..., K-1.
       labeled with high confidence.
     """
 
-    assert len(pred_probs) == len(labels)
+    assert_valid_inputs(X=None, y=labels, pred_probs=pred_probs, multi_label=multi_label)
```
cgnorthcutt, line 513: how much slower is the package by checking the extra things checked by `assert_valid_inputs` -- can you run a few things before and after and %timeit?

cleanlab/rank.py
```diff
@@ -85,7 +87,7 @@ def order_label_issues(
 
     """
 
-    assert len(pred_probs) == len(labels)
+    assert_valid_inputs(X=None, y=labels, pred_probs=pred_probs, multi_label=False)
```
cgnorthcutt, line 90: if you hard code `multi_label=False` here (which i agree is correct), you also need to specify in the comments at the top and in the `labels` field that this function does not support multi_label (since rank does not support multi_label) and this method depends on rank to get the quality scores.

cgnorthcutt (changes requested):
Good PR. Three high level things to check everywhere:

* how is speed affected (should be negligible but just double check before merging)
* what about all the functions in rank, count, and filter that dont call `assert_valid_inputs` but take `pred_probs` and `labels` as inputs?
* combining the two points above, avoid calling `assert_valid_inputs` over and over on data that's already verified as valid since it was already checked by a calling function.

cleanlab/rank.py
jwmueller: That function never supported multi_label AFAICT, so it's not a change from this PR.

The docstring for `labels` already refers user to the docstring in `get_label_quality_scores()` which says multilabel classification is not supported.

cgnorthcutt: ah that works, although i think mentioning this at the end of the overall function header would be helpful to users instead of having to following a reference chain.

cleanlab/filter.py
cgnorthcutt: i would assume so, too. It just makes reviewing PRs like this a lot easier if you're able to share the time to run pytest tests/ before and after this PR

Side note -- we should definitely add that feature to the CI & CD sometime soon. @ryansingman  @anishathalye

cleanlab/rank.py
cgnorthcutt: as a user, i usually check the docstring of the function i'm calling while i'm writing code, and dont check the comment at the top of a module. Up to you. Agree it makes the docstrings redundant and verbose to include such things in every function header. I'm only recommending in this case since that line of code specifically assumes it and the assert will break if multi_labels are passed in.

cleanlab/filter.py
anishathalye: Btw, talking about just asymptotics doesn't really capture concrete running time that well. We should be using concrete numbers from benchmarks.

Also, the `(pred_probs < 0)` allocates a M x N array, and the `(pred_probs > 1)` allocates a whole other M x N array. This can be done in O(1) space instead.

cleanlab/rank.py
jwmueller: I'd prefer not to have half of a docstring for the function. So either it should copy the entire (long) docstring of the `labels` argument from `get_label_quality_scores()`, or just point user to that function's docstring as it currently does.  It's unclear with half a docstring why this would be the most relevant info to convey vs other info such as label formatting.

cleanlab/filter.py
jwmueller: Sorry I wasn't referring to asymptotics when i mentioned: in practical terms these lines are negligible extra time for the functions in which they are called.

(not very reliable) timing of `pytest` (all unit tests)

user+sys time for existing master branch:  1855.51s
user+sys time for this PR code version:  1844.67s

This is not a good benchmark obviously... firstly because the inputs are small. 
For big inputs it'll obviously take longer since there are two new linear-time checks, but as I mentioned the practical effect is negligible compared to the runtime of the original function in which the checks have been added.

cgnorthcutt (approved):
LGTM. Thanks @jwmueller !

cleanlab/classification.py
```diff
@@ -512,31 +512,29 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
                     if "sample_weight" in self.clf_final_kwargs:
-                        print("Fitting final model on the clean data with custom sample_weight...")
+                        print("Fitting final model on the clean data with custom sample_weight ...")
                     else:
+                        if (
+                            "sample_weight" in inspect.getfullargspec(self.clf.fit).args
+                            and self.noise_matrix is None
+                        ):
+                            print(
+                                "Cannot utilize sample weights for final training. To utilize must either specify noise_matrix "
```
cgnorthcutt, line 524: nit but This can be a bit clearer. 

i'd state it as 'Cannot use sample weights for final training because your model's fit function doesn't support the parameter 'sample_weight'.

cgnorthcutt, line 524: nit but would help the user understand what's going on here with the following comment instead:

"Cannot use sample weights for final training!
Details about why this matters: sample weights can be helpful during final training to take into account for the amount of removed data in each class. This helps the learned model to have the correct prior. To use sample weights, you need to either provide the noise_matrix or have previously called self.find_label_issues() instead of filter.find_label_issues() which computes them for you."

cgnorthcutt (approved):
small nit that i think helps educate users with whats going on

cleanlab/classification.py
cgnorthcutt: Note this could be a helpful as warning instead of print if we find out that it actually degrades performance in most real-world use cases. Seems fine as print statement for now.

jwmueller: added shortened version of this

PR #304, opened 2022-07-01
CL for multiannotator data

huiwengoh:
Extending CL to tasks where data are labeled by multiple annotators.

### DIFF
--- a/cleanlab/dataset.py
+++ b/cleanlab/dataset.py
@@ -102,6 +102,13 @@ def rank_classes_by_label_quality(
     return df.sort_values(by="Label Quality Score", ascending=True).reset_index(drop=True)
 
 
+def _get_worst_class(labels, pred_probs):
+    ranked_noisy_classes = rank_classes_by_label_quality(labels, pred_probs).query(
+        "`Label Quality Score` < 1.0"
+    )
+    return ranked_noisy_classes["Class Index"][0] if len(ranked_noisy_classes) > 0 else np.nan
+
+
 def find_overlapping_classes(
     labels=None,
     pred_probs=None,
--- a/cleanlab/dataset.py
+++ b/cleanlab/dataset.py
@@ -104,8 +104,7 @@ def rank_classes_by_label_quality(
 
 def _get_worst_class(labels, pred_probs):
     """Returns the class with the lowest *Label Quality Score*, i.e. the most problematic class.
-    If all classes have a Label Quality Score of 1.0 (all classes have no issues), this function will
-    return NaN.
+    If all classes have a Label Quality Score of 1.0 (all classes have no issues), this function will return NaN.
 
     **Parameters**: For parameter info, see the docstring of :py:func:`find_overlapping_classes <cleanlab.dataset.find_overlapping_classes>`.
 
--- a/cleanlab/dataset.py
+++ b/cleanlab/dataset.py
@@ -103,6 +103,18 @@ def rank_classes_by_label_quality(
 
 
 def _get_worst_class(labels, pred_probs):
+    """Returns the class with the lowest *Label Quality Score*, i.e. the most problematic class.
+    If all classes have a Label Quality Score of 1.0 (all classes have no issues), this function will
+    return NaN.
+
+    **Parameters**: For parameter info, see the docstring of :py:func:`find_overlapping_classes <cleanlab.dataset.find_overlapping_classes>`.
+
+    Returns
+    -------
+    int
+        The index representing the class with the lowest *Label Quality Score*, returns NaN if
+        all classes do not have issues.
+    """
     ranked_noisy_classes = rank_classes_by_label_quality(labels, pred_probs).query(
         "`Label Quality Score` < 1.0"
     )
--- a/cleanlab/multiannotator.py
+++ b/cleanlab/multiannotator.py
@@ -20,9 +20,63 @@ def _get_label_quality_scores_with_NA(
     return label_quality_scores
 
 
+def get_consensus_labels(
+    labels_multiannotator: pd.DataFrame,
+    pred_probs: np.array,
+    method: str = "majority",
+) -> np.array:
+    """Returns consensus labels aggregated from multiple annotators, determined by method specified.
+
+    Parameters
+    ----------
+    labels_multiannotator : pd.DataFrame
+        2D pandas DataFrame of multiple given labels for each example with shape (N, M),
+        where N is the number of examples and M is the number of annotators.
+        For more details, labels in the same format expected by the :py:func:`get_label_quality_scores_multiannotator <cleanlab.multiannotator.get_label_quality_scores_multiannotator>`.
+
+    pred_probs : np.array
+        An array of shape ``(N, K)`` of model-predicted probabilities, ``P(label=k|x)``.
+        For details, predicted probabilities in the same format expected by the :py:func:`get_label_quality_scores_multiannotator <cleanlab.multiannotator.get_label_quality_scores_multiannotator>`.
+
+    method : str (Options: ["majority"])
+        The method used to aggregate labels from multiple annotators into a single consensus label. Default is by using the majority vote.
+        Options include:
+        - ``"majority"``: means consensus labels are reached via a simple majority vote among annotators, with ties broken via ``pred_probs``.
+
+    Returns
+    -------
+    consensus_labels: np.array
+        An array of shape ``(N,)`` with the consensus labels aggregated from all annotators.
+    """
+    valid_methods = ["majority"]
+
+    if method == "majority":
+        mode_labels_multiannotator = labels_multiannotator.mode(axis=1)
+        consensus_labels = []
+        for i in range(len(mode_labels_multiannotator)):
+            consensus_labels.append(
+                int(
+                    mode_labels_multiannotator.iloc[i][
+                        pred_probs[i][
+                            mode_labels_multiannotator.iloc[i].dropna().astype(int).to_numpy()
+                        ].argmax()
+                    ]
+                )
+            )
+    else:
+        raise ValueError(
+            f"""
+            {method} is not a method for determining a consensus!
+            Please choose a valid method: {valid_methods}
+            """
+        )
+
+    return np.array(consensus_labels)
+
+
 def compute_multiannotator_stats(
-    labels_multiannotator,
-    pred_probs,
+    labels_multiannotator: pd.DataFrame,
+    pred_probs: np.array,
     # label_quality_scores_multiannotator=None, # is this still needed? it is never called in the functioned
 ) -> pd.DataFrame:
     """Returns overall statistics about each annotator.
@@ -149,7 +203,7 @@ def get_label_quality_scores_multiannotator(
           calculated using weighted product of `label_quality_score` of `consensus_labels` and `annotator_agreement`
         Here annotator_1, annotator_2, ..., annotator_M suffixes may be replaced column names in ``labels_multiannotator`` DataFrame used to ID the annotators.
 
-    annotator_stats : pandas.DataFrame
+    annotator_stats : pandas.DataFrame, optional (returned if `return_annotator_stats=True`)
         Returns overall statistics about each annotator.
         For details, see the documentation of :py:func:`compute_multiannotator_stats<cleanlab.multiannotator.compute_multiannotator_stats>`
     """
@@ -174,6 +228,13 @@ def get_label_quality_scores_multiannotator(
         )
 
     # Raise error if labels_multiannotator has <= 1 column
+    if len(labels_multiannotator.columns) <= 1:
+        raise ValueError(
+            """
+                labels_multiannotator must have more than one column.
+                If there is only one annotator, use cleanlab.rank.get_label_quality_scores instead
+            """
+        )
 
     # Raise error if labels_multiannotator only has 1 set of labels
 
@@ -194,18 +255,11 @@ def get_label_quality_scores_multiannotator(
 
     # Compute the consensus_labels
     # TODO: conditional based on consensus_method, consensus_method can be a List[str], add dawid-skene
-    mode_labels_multiannotator = labels_multiannotator.mode(axis=1)
-    consensus_labels = []
-    for i in range(len(mode_labels_multiannotator)):
-        consensus_labels.append(
-            int(
-                mode_labels_multiannotator.iloc[i][
-                    pred_probs[i][
-                        mode_labels_multiannotator.iloc[i].dropna().astype(int).to_numpy()
-                    ].argmax()
-                ]
-            )
-        )
+    consensus_labels = get_consensus_labels(
+        labels_multiannotator=labels_multiannotator,
+        pred_probs=pred_probs,
+        method=consensus_method,
+    )
 
     # Compute the fraction of annotator agreeing with the consensus labels
     annotator_agreement = labels_multiannotator.assign(consensus_label=consensus_labels).apply(
--- a/cleanlab/multiannotator.py
+++ b/cleanlab/multiannotator.py
@@ -4,7 +4,7 @@ from cleanlab.rank import get_label_quality_scores
 from cleanlab.dataset import overall_label_health_score, _get_worst_class
 import pandas as pd
 
-# new functions - unsure where to place them yet
+# TODO: unsure where to place this function? should it go to cleanlab.rank or stay here
 def _get_label_quality_scores_with_NA(
     labels: pd.Series,
     pred_probs: np.array,
@@ -48,21 +48,18 @@ def get_consensus_labels(
     consensus_labels: np.array
         An array of shape ``(N,)`` with the consensus labels aggregated from all annotators.
     """
+
     valid_methods = ["majority"]
+    consensus_labels = np.empty(len(labels_multiannotator), dtype="int32")
 
     if method == "majority":
         mode_labels_multiannotator = labels_multiannotator.mode(axis=1)
-        consensus_labels = []
         for i in range(len(mode_labels_multiannotator)):
-            consensus_labels.append(
-                int(
-                    mode_labels_multiannotator.iloc[i][
-                        pred_probs[i][
-                            mode_labels_multiannotator.iloc[i].dropna().astype(int).to_numpy()
-                        ].argmax()
-                    ]
-                )
-            )
+            consensus_labels[i] = mode_labels_multiannotator.iloc[i][
+                pred_probs[
+                    i, mode_labels_multiannotator.iloc[i].dropna().astype(int).to_numpy()
+                ].argmax()
+            ]
     else:
         raise ValueError(
             f"""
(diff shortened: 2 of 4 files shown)


tests/test_multiannotator.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+    # Generate our noisy labels using the noise_matrix for specified number of annotators.
+    s = pd.DataFrame(
+        np.vstack(
+            [generate_noisy_labels(true_labels_train, noise_matrix) for _ in range(num_annotators)]
+        ).transpose()
+    )
+
+    # Each annotator only labels approximately 75% of the dataset
```
jwmueller, line 61: maybe we want to test with more sparsity? Eg. each annotator only labels say ~20%?

jwmueller: If more sparsity is utilized, probably need to delete rows that have no annotators (or keep at least 1 annotator for such rows)

cleanlab/dataset.py
```diff
@@ -102,6 +102,13 @@ def rank_classes_by_label_quality(
     return df.sort_values(by="Label Quality Score", ascending=True).reset_index(drop=True)
 
 
+def _get_worst_class(labels, pred_probs):
```
jwmueller, line 105: Would be good to add a one-line docstring summarizing purpose of this method

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+    )
+
+    # Create multi-annotator stats DataFrame from its columns
+    return pd.DataFrame(
+        {
+            "overall_quality": overall_label_health_score_df,
+            "num_labeled": num_labeled,
+            "worst_class": worst_class,
+        }
+    )
```
jwmueller, line 255: Add:  `# TODO:  update this method once multi-annotator functionality is finalized.`

This function looks good for now!  But we may not want to use dataset.py  methods for this in the end, because dataset.py methods do not consider the other annotators' labels at all and attempt to estimate ground truth labels from each annotator independently (which is likely bad idea when some annotators hardly labeled any data).

Eventually we probably want to first figure out our guess of the best "consensus_label" and then use that as the ground-truth label which we use to rate each annotator (as well as their worst label).

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
+import numpy as np
+import pandas as pd
+from typing import List, Union, Tuple
+from cleanlab.rank import get_label_quality_scores
+from cleanlab.dataset import overall_label_health_score, _get_worst_class
+import warnings
+
+# TODO: unsure where to place this function? should it go to cleanlab.rank or stay here
```
jwmueller, line 8: I'd add a one-line docstring and then move this method to: internal/label_quality_utils.py


https://github.com/cleanlab/cleanlab/blob/master/cleanlab/internal/label_quality_utils.py

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+        to an example `x` and contains the model-predicted probabilities that
+        `x` belongs to each possible class, for each of the K classes. The
+        columns must be ordered such that these probabilities correspond to
+        class 0, 1, ..., K-1.
+        **Caution**: `pred_probs` from your model must be out-of-sample!
+        You should never provide predictions on the same examples used to train the model,
+        as these will be overfit and unsuitable for finding label-errors.
+        To obtain out-of-sample predicted probabilities for every datapoint in your dataset, you can use :ref:`cross-validation <pred_probs_cross_val>`.
+        Alternatively it is ok if your model was trained on a separate dataset and you are only evaluating
+        data that was previously held-out.
```
jwmueller, line 298: Can just link to docstring of `get_label_quality_scores()` method and write that this `pred_probs` should be of the same format as the one for that method.

This reduces docstring duplication

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+    1 - clean label (the given label is likely correct).
+    0 - dirty label (the given label is unlikely correct).
+
+    Parameters
+    ----------
+    labels_multiannotator : pd.DataFrame
+        2D pandas DataFrame of multiple given labels for each example with shape (N, M),
+        where N is the number of examples and M is the number of annotators.
+        labels_multiannotator[n][m] = given label for n-th example by m-th annotator.
+        For a dataset with K classes, the given labels must be an integer in 0, 1, ..., K-1 or
```
jwmueller, line 283: [nit]  the given labels -> each given label

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+        Here annotator_1, annotator_2, ..., annotator_M suffixes may be replaced column names in ``labels_multiannotator`` DataFrame used to ID the annotators.
+
+    annotator_stats : pandas.DataFrame, optional (returned if `return_annotator_stats=True`)
+        Returns overall statistics about each annotator.
+        For details, see the documentation of :py:func:`compute_multiannotator_stats<cleanlab.multiannotator.compute_multiannotator_stats>`
+    """
+
+    # Raise error if labels_multiannotator has NaN rows
+    if labels_multiannotator.isna().all(axis=1).any():
+        raise ValueError("labels_multiannotator cannot have rows with all np.NaN.")
```
jwmueller, line 340: [nit] can just say "NaN" instead of "np.NaN"

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+    consensus_method : str or List[str]
+        For each example, which method should be used to aggregate labels from multiple annotators into a single consensus label.
+        Options include:
+        - ``"majority"``: means consensus labels are reached via a simple majority vote among annotators, with ties broken via ``pred_probs``.
+        A List may be passed if you want to consider multiple methods for producing consensus labels.
+        If a List is passed, then the 0th element of List is the method used to produce columns "consensus_label", "quality_of_consensus", "annotator_agreement" in the returned DataFrame.
+        The 1st, 2nd, 3rd, etc. elements of this List are output as extra columns in the returned ``pandas DataFrame`` with names formatted as:
+        consensus_label_SUFFIX, quality_of_consensus_SUFFIX
+        where SUFFIX = the str element of this list, which must correspond to a valid method for computing consensus labels.
+
```
jwmueller, line 309: Let's add an explicit example here once there are few more methods

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+    if labels_multiannotator.apply(lambda s: len(s.dropna()) == 1, axis=1).all():
+        raise ValueError(
+            """Each example only has one label, collapse the labels into a 1-D array and use
+            cleanlab.rank.get_label_quality_scores instead"""
+        )
+
+    # Raise warning if no examples with 2 or more annotators agree
+    if labels_multiannotator.apply(
+        lambda s: np.array_equal(s.dropna().unique(), s.dropna()), axis=1
+    ).all():
```
jwmueller, line 369: maybe we can just move this check later in the code and do it based on the `annotator_agreement` column to avoid having to do extra compute?

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+        labels=labels[pd.notna(labels)],
+        pred_probs=pred_probs[pd.notna(labels)],
+        **kwargs,
+    )
+    label_quality_scores = pd.Series(np.nan, index=np.arange(len(labels)))
+    label_quality_scores[pd.notna(labels)] = label_quality_scores_subset
+    return label_quality_scores
+
+
+def get_consensus_labels(
```
jwmueller, line 24: Let's make this helper method (with underscore prefix): `_get_consensus_labels()`

So we don't have to commit to supporting this exact API forever.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+            f"""
+            {method} is not a valid consensus method!
+            Please choose a valid consensus_method: {valid_methods}
+            """
+        )
+
+    return consensus_labels
+
+
+def get_annotator_agreement(
```
jwmueller, line 76: Let's make this helper method (with underscore prefix): `_get_annotator_agreement()`

So we don't have to commit to supporting this exact API forever.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+        .apply(
+            lambda s: np.mean(s.drop("consensus_label") == s["consensus_label"]),
+            axis=1,
+        )
+        .to_numpy()
+    )
+
+
+# TODO: add other ways to the get quality score (methods that can be passed to get_label_quality_score)
+def get_quality_of_consensus(
```
jwmueller, line 109: Let's make this helper method (with underscore prefix): `_get_quality_of_consensus()`

So we don't have to commit to supporting this exact API forever.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+        An array of shape ``(N,)`` with the quality score of the consensus.
+        TODO: better explanation of this
+    """
+    lqs_consensus_labels = get_label_quality_scores(consensus_labels, pred_probs, **kwargs)
+    frac = pred_probs[range(len(consensus_labels)), (consensus_labels)] / num_annotations
+    quality_of_consensus = frac * lqs_consensus_labels + (1 - frac) * annotator_agreement
+    return quality_of_consensus
+
+
+def compute_consensus_stats(
```
jwmueller, line 146: Let's make this helper method (with underscore prefix): `_compute_consensus_stats()`

So we don't have to commit to supporting this exact API forever.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+    )
+    label_quality_scores_multiannotator = label_quality_scores_multiannotator.add_prefix(
+        "quality_annotator_"
+    )
+
+    # Count number of non-NaN values for each example
+    num_annotations = labels_multiannotator.count(axis=1).to_numpy()
+    label_quality_scores_multiannotator["num_annotations"] = num_annotations
+
+    # Compute consesnsus labels, annotator agreement and quality of consensus
```
jwmueller, line 384: typo: consesnsus

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+        )
+
+    # Raise error if labels_multiannotator has <= 1 column
+    if len(labels_multiannotator.columns) <= 1:
+        raise ValueError(
+            """labels_multiannotator must have more than one column. 
+            If there is only one annotator, use cleanlab.rank.get_label_quality_scores instead"""
+        )
+
+    # Raise error if labels_multiannotator only has 1 set of labels per example
```
jwmueller, line 359: "only has 1 set of labels per example" -> "only has 1 label per example"

jwmueller: (language in original version was a bit confusing to me)

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+        An array of shape ``(N,)`` with the consensus labels aggregated from all annotators.
+    """
+
+    valid_methods = ["majority"]
+    consensus_labels = np.empty(len(labels_multiannotator), dtype="int32")
+
+    # TODO: is there a more efficient/elegant way to do this?
+    if method == "majority":
+        mode_labels_multiannotator = labels_multiannotator.mode(axis=1)
+        for i in range(len(mode_labels_multiannotator)):
```
jwmueller, line 59: Can think more about this, but at least you should only have to loop over those examples `i` where `mode_labels_multiannotator[i]` is longer than 1 in the first place right?  So restricting the loop to only those examples should already provide big speedup (could even have special handling of examples with only 2 labels if you've already computed `num_annotations` at this point, since we know these will have ties).

jwmueller: Another alternative is to do `value_counts()` and implement the mode + tie-breaking yourself (eg. could simply add the pred_probs to the value_counts if called with `sort=False`). 

But would probably first want to some timing benchmarks/googling to see how much slower `value_counts()` is in comparison since it involves extra computation than `mode()`.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+        For more details, labels in the same format expected by the :py:func:`get_label_quality_scores_multiannotator <cleanlab.multiannotator.get_label_quality_scores_multiannotator>`.
+
+    pred_probs : np.array
+        An array of shape ``(N, K)`` of model-predicted probabilities, ``P(label=k|x)``.
+        For details, predicted probabilities in the same format expected by the :py:func:`get_label_quality_scores_multiannotator <cleanlab.multiannotator.get_label_quality_scores_multiannotator>`.
+
+    num_annotations : np.array
+        An array of shape ``(N,)`` with the number of annotators that have labeled each example.
+
+    method : str (Options: ["majority"])
```
jwmueller, line 169: current argname is "consensus_method"

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+
+
+def get_consensus_labels(
+    labels_multiannotator: pd.DataFrame,
+    pred_probs: np.array,
+    method: str,
+) -> np.array:
+    """Returns consensus labels aggregated from multiple annotators, determined by method specified.
+
+    Parameters
```
jwmueller, line 31: To reduce docstring duplication, let's try to refer to the documentation of `get_label_quality_scores_multiannotator()` and `compute_multiannotator_stats()`

as much as possible in these helper methods. Those two are the main functions that should have the most complete-looking documentation.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+        pred_probs=pred_probs,
+        num_annotations=num_annotations,
+        annotator_agreement=annotator_agreement,
+        **kwargs,
+    )
+
+    return (consensus_labels, annotator_agreement, quality_of_consensus)
+
+
+def compute_multiannotator_stats(
```
jwmueller, line 206: What do you think about changing this to `get_annotator_stats()` ?

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+    return pd.DataFrame(
+        {
+            "overall_quality": overall_label_health_score_df,
+            "num_labeled": num_labeled,
+            "worst_class": worst_class,
+        }
+    )
+
+
+def get_label_quality_scores_multiannotator(
```
jwmueller, line 258: what do you think about shortening to: `get_label_quality_multiannotator()` ?

(since this doesn't just return scores)

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+        To obtain out-of-sample predicted probabilities for every datapoint in your dataset, you can use :ref:`cross-validation <pred_probs_cross_val>`.
+        Alternatively it is ok if your model was trained on a separate dataset and you are only evaluating
+        data that was previously held-out.
+
+    consensus_method : str or List[str]
+        For each example, which method should be used to aggregate labels from multiple annotators into a single consensus label.
+        Options include:
+        - ``"majority"``: means consensus labels are reached via a simple majority vote among annotators, with ties broken via ``pred_probs``.
+        A List may be passed if you want to consider multiple methods for producing consensus labels.
+        If a List is passed, then the 0th element of List is the method used to produce columns "consensus_label", "quality_of_consensus", "annotator_agreement" in the returned DataFrame.
```
jwmueller, line 305: I don't think method should affect "annotator_agreement" for now?

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,436 @@
...
+
+    kwargs : dict
+        Keyword arguments to pass into ``get_label_quality_scores()``.
+
+    Returns
+    -------
+    label_quality_score_multiannotator : pandas.DataFrame
+        pandas DataFrame in which each row corresponds to one example, with columns:
+        - ``quality_of_annotator_1``, ``quality_of_annotator_2``, ..., ``quality_of_annotator_M``: the label quality score for the labels provided by annotator M (is ``NaN`` for examples which this annotator did not label).
+        - ``consensus_labels``: the single label that is best for each example (you can control how it is derived from all annotators' labels via the argument: ``consensus_method``)
```
jwmueller, line 327: Just `consensus_label` since there's only 1 per example?

cleanlab/multiannotator.py
huiwengoh: It does because `annotator_agreement` is dependent on the consensus, which is dependent on the consensus method specified. 
There might be a better way to write this, it might currently be confusing as we do not return `annotator_agreement` scores for the other methods (although it could be implicitly used in the calculation of `quality_of_consensus`)

cleanlab/multiannotator.py
huiwengoh: Yup that makes sense, in my head it was `labels` because it was an array of labels (hence plural), but that thought is inconsistent with the rest of the naming. Will make the change!

cleanlab/multiannotator.py
huiwengoh: changed all `compute`s to `get`s !

tests/test_multiannotator.py
huiwengoh: TODO: edge case here that will cause `get_multiannotator_stats` to fail - `dataset.overall_label_health_score` requires that there are no missing labels, which is not guaranteed here (some annotators might label too few examples, or bad annotator might label same labels for all examples)

cleanlab/dataset.py
```diff
@@ -102,6 +102,25 @@ def rank_classes_by_label_quality(
...
+    """Returns the class with the lowest *Label Quality Score*, i.e. the most problematic class.
+    If all classes have a Label Quality Score of 1.0 (all classes have no issues), this function will
+    return NaN.
```
jwmueller, line 108: [nit] odd spacing

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,467 @@
...
+    lqsm_columns = label_quality_scores_multiannotator.columns.tolist()
+    new_index = (
+        lqsm_columns[labels_multiannotator.shape[1] :]
+        + lqsm_columns[: labels_multiannotator.shape[1]]
+    )
+    label_quality_scores_multiannotator = label_quality_scores_multiannotator.reindex(
+        columns=new_index
+    )
+
+    annotator_stats = get_multiannotator_stats(labels_multiannotator, pred_probs)
```
jwmueller, line 455: should only be computed if we need it (ie. if verbose or if return_annotator_stats)

tests/test_multiannotator.py
```diff
@@ -0,0 +1,164 @@
...
+        # "n": n,
+    }
+
+
+# Global to be used by all test methods. Only compute this once for speed.
+data = make_data()
+
+
+# TODO: fix this test to work with NaN values - also unsure if this test should sit in this file
+# logic of current test does not work - randomly deleting data will break it (put in another file?)
```
jwmueller, line 109: I think this test can live here since it's currently the only context where `_get_worst_class()` is ever used.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,467 @@
...
+
+    return (consensus_label, annotator_agreement, quality_of_consensus)
+
+
+# TODO: update this method once multi-annotator functionality is finalized
+def get_multiannotator_stats(
+    labels_multiannotator: pd.DataFrame,
+    pred_probs: np.array,
+    # label_quality_scores_multiannotator=None, # is this still needed? it is never called in the function
+) -> pd.DataFrame:
```
jwmueller, line 222: Let's also support `quality_method` as an arg here (with str values). 
For now let's use the current logic `overall_label_health_score` if `quality_method="auto"`.

Let's also support `quality_method="agreement"` in which case we report fraction of this annotator's annotations that agreed with the consensus label.

cleanlab/multiannotator.py
jwmueller: I think we resolved this in discussion but feel free to unresolve if not

cleanlab/multiannotator.py
jwmueller: More generally, I think we should support any `quality_method` that can be passed into `get_label_quality_multiannotator()`.

In the case where `quality_method` here is one of the options supported for `get_label_quality_multiannotator()`, we can do the following:

Compute the quality-score of each of one annotator's labels as if it had replaced consensus-label in the consensus-quality-scoring inside of `get_label_quality_multiannotator()`.  Then report the average of the resulting quality-scores over all labels that this annotator provided, as the quality of this annotator.

For example if user specified to use self-confidence label-quality score, you can simply compute the self-confidence of each label provided by one annotator and then report their average to score that annotator.

huiwengoh: Added `agreement_with_consensus` as a column

TODO: better way to support `quality_method` options, currently there is a mismatch between quality methods for consensus stats and multiannotator stats (eg. `quality_method == 'agreement` makes sense for consensus stats but not multiannotator stats). We might have to separate these two args if the way to aggregate consensus vs annotator quality differs too much.

cleanlab/multiannotator.py
huiwengoh: Updated so that less computation is done if `len(mode_labels_multiannotator[i]) == 1`, but still requires looping over all examples

PR #305, opened 2022-07-05
Added support to build KNN graph with only training data

ulya-tkch:
* Added default classifier for get_knn_distance_ood_scores()
* Question: Should we create default NN with 10 neighbors always or 'k' if k exists?
* TODO: Unit testing push


cleanlab/rank.py
```diff
@@ -584,6 +585,9 @@ def get_knn_distance_ood_scores(
     avg_nbrs_distances : np.array
       Average distance to k-nearest neighbors for each datapoint which is used as a score for OOD detection.
     """
+    # if nbrs is not provided, then use default KNN classifier (k=10)
+    if nbrs is None:
+        nbrs = NearestNeighbors(n_neighbors=10, metric="cosine").fit(features)
```
jwmueller, line 590: > Should we create default NN with 10 neighbors always or 'k' if k exists?

Let's do k if k exists, but 10 as the default otherwise if k = None and nbrs not given.

cleanlab/rank.py
```diff
@@ -572,21 +572,27 @@ def get_knn_distance_ood_scores(
...
       Average distance to k-nearest neighbors for each datapoint which is used as a score for OOD detection.
     """
+    # if nbrs is not provided, then use default KNN classifier
+    if nbrs is None:
+        n_neighbors = k if k is not None else 10
+        nbrs = NearestNeighbors(n_neighbors=n_neighbors, metric="cosine").fit(features)
 
     # number of neighbors specified when fitting instantiated NearestNeighbors class object
-    max_k = nbrs.n_neighbors
+    max_k = nbrs.n_neighbors if nbrs else 10
```
jwmueller, line 595: I think max_k should also take into account if k was specified or not. 
I.e. 
```
if nbrs:
    max_k = nbrs.n_neighbors
elif k is not None:
    max_k = k
else:
    max_k = 10
```

jwmueller: Ah scratch that ^ if the code is already implementing the same logic but in a different way?

I guess the odd part in the way the code is structured right now is why would the `else 10` clause ever be triggered?

Based on the existing code (copied below) it seems that cannot happen or maybe I'm just missing something:

```
if nbrs is None:
        n_neighbors = k if k is not None else 10
        nbrs = NearestNeighbors(n_neighbors=n_neighbors, metric="cosine").fit(features)

    # number of neighbors specified when fitting instantiated NearestNeighbors class object
    max_k = nbrs.n_neighbors if nbrs else 10
```

cleanlab/rank.py
```diff
@@ -557,7 +557,7 @@ def get_confidence_weighted_entropy_for_each_label(
 
 
 def get_knn_distance_ood_scores(
-    features: np.array, nbrs: NearestNeighbors, k: int = None
+    features: np.array, nbrs: NearestNeighbors = None, k: int = None
```
calebchiam, line 560: nit: favor long form > short form, i.e. use 'neighbors'. That being said, calling this argument `knn` might be even clearer.

cleanlab/rank.py
calebchiam: +1. Cleaner construction might be to just set `k=10` in the function definition.

Then:
```
if nbrs is None:
  nbrs = NearestNeighbors(n_neighbors=k, metric="cosine").fit(features)
```

and
```
if k > nbrs.n_neighbors:
  k = nbrs.n_neighbors
  [warn that we cap k at nbrs.n_neighbors]

avg_nbrs_distances = distances[:, :k].mean(axis=1)

```

ulya-tkch: Way cleaner thank you guys!

tests/test_rank.py
```diff
@@ -403,42 +393,59 @@ def test_get_knn_distance_ood_scores():
...
+        features=X_with_ood,
+        k=instantiated_k,  # this should use default classifier (same as above) and k = instantiated_k
+    )
+
+    avg_knn_distances_default_k = rank.get_outlier_scores(
+        features=X_with_ood  # default k should be set to 10 == instantiated_k
+    )
+
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
```
jwmueller, line 438: Generally should not be catching warnings in unit test.  You can inform the test that it is expected to throw a particular type of warning via something like this for example:
```
@pytest.mark.filterwarnings("ignore::UserWarning")
```
as demonstrated here:
https://github.com/cleanlab/cleanlab/blob/master/tests/test_classification.py#L152

tests/test_rank.py
```diff
@@ -403,42 +393,59 @@ def test_get_knn_distance_ood_scores():
...
+        avg_knn_distances = rank.get_outlier_scores(
+            features=X_with_ood,
+            knn=knn,
+            k=25,  # this should throw warn, k should be set to instantiated_k
         )
+
+        assert len(w) == 1
+        assert "Value of k changed" in str(w[-1].message)
+
+    # Score sums should be equal because the three classifiers used have identical params and fit
```
jwmueller, line 449: classifiers?

tests/test_rank.py
```diff
@@ -403,42 +393,59 @@ def test_get_knn_distance_ood_scores():
...
     # Fit nearest neighbors on X_train
-    nbrs = NearestNeighbors(n_neighbors=5).fit(X_train)
+    knn = NearestNeighbors(n_neighbors=5).fit(X_train)
 
     # Get KNN distance as OOD score
     k = 5
-    knn_distance_ood_score = rank.get_knn_distance_ood_scores(
-        features=X_test_with_ood, nbrs=nbrs, k=k
-    )
+    knn_distance_ood_score = rank.get_outlier_scores(features=X_test_with_ood, knn=knn, k=k)
```
jwmueller, line 400: Add assert after this to verify the k is properly set?  

One failure mode I can see for this code is the default k=10 accidentally somehow overwrites a user-specified value of k.  Current unit tests seems like they'd still pass even if the code internally was using the wrong k value, so would be good to have test where this cannot happen (see also below comment).

tests/test_rank.py
```diff
@@ -403,42 +393,59 @@ def test_get_knn_distance_ood_scores():
...
+            knn=knn,
+            k=25,  # this should throw warn, k should be set to instantiated_k
         )
+
+        assert len(w) == 1
+        assert "Value of k changed" in str(w[-1].message)
+
+    # Score sums should be equal because the three classifiers used have identical params and fit
+    assert avg_knn_distances.sum() == avg_knn_distances_default_model.sum()
+    assert avg_knn_distances_default_k.sum() == avg_knn_distances.sum()
```
jwmueller, line 451: maybe also verify the distances are **not** the same if the user specified a different value of k like k=1?

cleanlab/rank.py
```diff
@@ -556,52 +556,54 @@ def get_confidence_weighted_entropy_for_each_label(
...
 
     Parameters
     ----------
     features : np.array
       Feature matrix of shape (N, M), where N is the number of datapoints and M is the number of features.
       This is the "query set" of features for each datapoint which are used for nearest neighbor search.
 
-    nbrs : sklearn.neighbors.NearestNeighbors
+    knn : sklearn.neighbors.NearestNeighbors
       Instantiated NearestNeighbors class object that's been fitted on a dataset in the same feature space.
```
jwmueller, line 571: NearestNeighbors -> ``NearestNeighbors`` (note the double backticks required for docstring rendering of code-font).

tests/test_rank.py
ulya-tkch: Sounds good, I'll do this from now on

cleanlab/rank.py
ulya-tkch: Why are we not using `` `` for all the other code-esq stuff in the header? i.e. k=min(10, max_k) -> ``k=min(10, max_k)``

jwmueller: It should be using the double backtick for all code-esq stuff in all docstrings.  Sorry the original code PR was merged a bit too quickly, so there is lots of room for improvement in this PR :D 

It'd be helpful if you can go through all the docstrings for this function and properly double backtick the code

cleanlab/rank.py
```diff
@@ -556,52 +556,54 @@ def get_confidence_weighted_entropy_for_each_label(
...
       Note that the distance metric and n_neighbors is specified when instantiating this class.
+      If knn=None, then by default knn=sklearn.neighbors.NearestNeighbors(n_neighbors=k, metric="cosine").fit(features)
```
jwmueller, line 573: knn=None and the knn=sklearn... should also be enclosed in double backticks to be formatted as code in the docstring.

cleanlab/rank.py
```diff
@@ -556,52 +556,54 @@ def get_confidence_weighted_entropy_for_each_label(
...
       See: https://scikit-learn.org/stable/modules/neighbors.html
 
-    k : int, default=None
-      Number of neighbors to use when calculating average distance to neighbors.
-      This value k needs to be less than or equal to max_k which is the n_neighbors used when fitting instantiated NearestNeighbors class object.
-      If k=None, then by default k=min(10, max_k) is used where max_k is extracted from the given nbrs.
+    k : int, default=None Number of neighbors to use when calculating average distance to neighbors. This value k
+    needs to be less than or equal to max_k which is the n_neighbors used when fitting instantiated NearestNeighbors
+    class object. If k=None, then by default k=min(10, max_k) is used where max_k is extracted from the given knn.
```
jwmueller, line 578: > If k=None, then by default k=min(10, max_k) is used where max_k is extracted from the given knn

I actually don't think this is good behavior. If `k = None and knn !=None`, then can we just set `k = n_neighbors` ?  Any other behavior seems counterintuitive to me.

tests/test_rank.py
```diff
@@ -403,42 +393,59 @@ def test_get_knn_distance_ood_scores():
...
-    nbrs = NearestNeighbors(n_neighbors=5).fit(X_train)
+    # Create NN classifiers with small instantiated k and fit on data
+    knn = NearestNeighbors(n_neighbors=instantiated_k, metric="cosine").fit(X_with_ood)
 
-    with pytest.raises(AssertionError) as e:
-        _ = rank.get_knn_distance_ood_scores(
-            features=X_test_with_ood, nbrs=nbrs, k=10  # this should throw an assertion error
+    avg_knn_distances_default_model = rank.get_outlier_scores(
+        features=X_with_ood,
+        k=instantiated_k,  # this should use default classifier (same as above) and k = instantiated_k
```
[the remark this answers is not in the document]
ulya-tkch: Estimator is more filtting!

cleanlab/rank.py
```diff
@@ -556,52 +556,54 @@ def get_confidence_weighted_entropy_for_each_label(
...
+    If knn is not provided, then by default k=10.
```
jwmueller, line 578: Let's remove all mentions of `max_k` from docstring, sorry that should've never been in there before. In general not a good idea to use internal variable names in the docstring.

cleanlab/rank.py
```diff
@@ -556,52 +556,54 @@ def get_confidence_weighted_entropy_for_each_label(
...
       Instantiated NearestNeighbors class object that's been fitted on a dataset in the same feature space.
       Note that the distance metric and n_neighbors is specified when instantiating this class.
+      If knn=None, then by default knn=sklearn.neighbors.NearestNeighbors(n_neighbors=k, metric="cosine").fit(features)
       See: https://scikit-learn.org/stable/modules/neighbors.html
 
-    k : int, default=None
-      Number of neighbors to use when calculating average distance to neighbors.
-      This value k needs to be less than or equal to max_k which is the n_neighbors used when fitting instantiated NearestNeighbors class object.
-      If k=None, then by default k=min(10, max_k) is used where max_k is extracted from the given nbrs.
+    k : int, default=None Number of neighbors to use when calculating average distance to neighbors. This value k
```
jwmueller, line 574: The description of the arg should be on a newline

cleanlab/rank.py
ulya-tkch: Actually good point, for ``k=None`` to be true now ``k=None`` would need to be explicitly passed into the function. ``k=10`` if there is no k specified,  ``k=max_k`` if ``k>n_neighbors`` and ``k=k`` otherwise. Need to update this

cleanlab/rank.py
```diff
@@ -556,52 +556,54 @@ def get_confidence_weighted_entropy_for_each_label(
...
+    needs to be less than or equal to max_k which is the n_neighbors used when fitting instantiated NearestNeighbors
```
jwmueller, line 575: Let's update the first 2 sentences of docstring to the following (the later sentence should also be correspondingly updated once the logic is updated):

Number of neighbors to use when calculating outlier score (average distance to neighbors).  If an existing ` ``knn`` ` object is provided, you can still specify that outlier score should use a different value of k than originally used in the ` ``knn`` `, as long as your specified value of k is smaller than the one originally used.

cleanlab/rank.py
jwmueller: add: 

with ` ``k=10`` ` if you do not specify ` ``k`` ` here.

cleanlab/rank.py
```diff
@@ -556,52 +556,54 @@ def get_confidence_weighted_entropy_for_each_label(
...
     Parameters
     ----------
     features : np.array
       Feature matrix of shape (N, M), where N is the number of datapoints and M is the number of features.
       This is the "query set" of features for each datapoint which are used for nearest neighbor search.
 
-    nbrs : sklearn.neighbors.NearestNeighbors
+    knn : sklearn.neighbors.NearestNeighbors
       Instantiated NearestNeighbors class object that's been fitted on a dataset in the same feature space.
       Note that the distance metric and n_neighbors is specified when instantiating this class.
```
jwmueller, line 570: Add sentence:

You can also pass in a subclass of ` ``sklearn.neighbors.NearestNeighbors`` ` which allows you to use faster approximate neighbor libraries as long as you wrap them behind the same sklearn API.

cleanlab/rank.py
```diff
@@ -556,52 +556,54 @@ def get_confidence_weighted_entropy_for_each_label(
...
 
-def get_knn_distance_ood_scores(
-    features: np.array, nbrs: NearestNeighbors, k: int = None
-) -> np.array:
-    """Returns the KNN distance out-of-distribution (OOD) score for each datapoint.
+def get_outlier_scores(features: np.array, knn: NearestNeighbors = None, k: int = 10) -> np.array:
+    """Returns the KNN distance outlier score for each datapoint.
 
-    This is a function to compute OOD scores where higher scores indicate the datapoint is more likely to be OOD.
+    This is a function to compute outlier scores where higher scores indicate the datapoint is more likely to be an outlier.
```
jwmueller, line 560: (Note cleanlab library tries to use vocab "example" to refer to each datapoint).  I'd update the main docstring to something like:

Returns an outlier score for each example based on its feature values, where higher scores indicate examples more likely to be outliers.  The score is based on the average distance between the example and its K nearest neighbors in the dataset (in feature space).

cleanlab/rank.py
```diff
@@ -556,52 +556,54 @@ def get_confidence_weighted_entropy_for_each_label(
...
 
     Parameters
     ----------
     features : np.array
       Feature matrix of shape (N, M), where N is the number of datapoints and M is the number of features.
       This is the "query set" of features for each datapoint which are used for nearest neighbor search.
```
jwmueller, line 566: I'd update docstring to:

Feature array of shape (N, M), where N is the number of examples and M is the number of features used to represent each example (each row corresponds to one example).  All features should be numeric.  For unstructured data (eg. images, text, categorical values, ...), you should provide vector embeddings to represent each example (e.g. extracted from some pretrained neural network).

cleanlab/rank.py
```diff
@@ -556,52 +556,54 @@ def get_confidence_weighted_entropy_for_each_label(
     return label_quality_scores
 
 
-def get_knn_distance_ood_scores(
-    features: np.array, nbrs: NearestNeighbors, k: int = None
-) -> np.array:
-    """Returns the KNN distance out-of-distribution (OOD) score for each datapoint.
+def get_outlier_scores(features: np.array, knn: NearestNeighbors = None, k: int = 10) -> np.array:
```
jwmueller, line 557: np.array -> np.ndarray everywhere

cleanlab/rank.py
```diff
@@ -556,52 +554,59 @@ def get_confidence_weighted_entropy_for_each_label(
...
     Returns
     -------
-    avg_nbrs_distances : np.array
-      Average distance to k-nearest neighbors for each datapoint which is used as a score for OOD detection.
+    avg_knn_distances : np.array
+      Average distance to k-nearest neighbors for each datapoint which is used as a score for outlier detection.
     """
+    # if knn is not provided, then use default KNN as estimator
+    if knn is None:
+        knn = NearestNeighbors(n_neighbors=k, metric="cosine").fit(features)
```
jwmueller, line 590: after this (while still inside of the `if` clause), we should do:

```
features = None
```
so that it'll be None in the later call to: `knn.kneighbors(features)`.

When kneighbors gets None as its input it returns the neighbors for the training data:
https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.NearestNeighbors.html#sklearn.neighbors.NearestNeighbors.kneighbors

which is what we want to not waste computation (and to get the number of neighbors correct, otherwise the datapoint itself was being counted as a neighbor of itself)

cleanlab/rank.py
jwmueller: May also help to explicitly list exactly what methods of sklearn.neighbors.NearestNeighbors their custom class actually needs to support. I think it's only `fit()`, `kneighbors()`, and the`.n_neighbors` attribute, but double check this.

cleanlab/rank.py
ulya-tkch: Wouldn't the same problem occur in the original call:
```
knn = NearestNeighbors(n_neighbors=k, metric="cosine").fit(features)
outlier_score = get_outlier_scores(features, knn, k)
```

ulya-tkch: Hacky way to enforce that would be
```
if np.array_equal(knn.__dict__['_fit_X'], features):
        features = None
```

jwmueller: Good point!  I'd just clarify the following things at the end of the `knn` arg docstring:

If you specify ` ``knn`` ` here and wish to find outliers in the same data you already passed into ` ``knn.fit(features)`` `, then you should simply specify ` ``features = None`` ` here.  This requires that your ` ``knn.kneighbors(None)`` `
returns the distances to the datapoints it was ` ``fit()`` ` on.  Not recommended but if you specify ` ``features`` `  here again as the same features previously used in ` ``knn.fit()`` `, then each example will be also counted as one of its own neighbors and the average KNN distances will be skewed downward (although their relative ordering will be preserved).

jwmueller: I'd want to avoid your route in order to not impose more requirements on what the user's custom `knn` class has to satisfy (we want to be compatible with as many approximate neighbor libraries as possible here)

ulya-tkch: @jwmueller is this hacky way ok, that is my most recent push.

jwmueller: Let's not do that for the reasons mentioned above. We really don't want the user's custom knn class to have to conform a lot of specifics, and I don't think it's a big deal if they happened to pass in features twice. So let's just clarify this issue in the docstring as I mentioned above and not do anything about this issue in the source code.

ulya-tkch: Docstring for ``knn`` looked SO long if I included that entire (great!) explanation. Shortened it a little, let me know if that still conveys same information.

jwmueller: shortening my overly long statements is always a good idea :) 
New docstring looks great!

cleanlab/rank.py
```diff
@@ -556,52 +554,68 @@ def get_confidence_weighted_entropy_for_each_label(
...
-      Feature matrix of shape (N, M), where N is the number of datapoints and M is the number of features.
-      This is the "query set" of features for each datapoint which are used for nearest neighbor search.
+    features : np.ndarray
+      Feature array of shape ``(N, M)``, where N is the number of examples and M is the number of features used to represent each example.
+      All features should be numeric. For unstructured data (eg. images, text, categorical values, ...), you should provide
+      vector embeddings to represent each example (e.g. extracted from some pretrained neural network).
 
-    nbrs : sklearn.neighbors.NearestNeighbors
-      Instantiated NearestNeighbors class object that's been fitted on a dataset in the same feature space.
+    knn : sklearn.neighbors.NearestNeighbors
```
jwmueller, line 573: knn : sklearn.neighbors.NearestNeighbors, default = None

cleanlab/rank.py
```diff
@@ -556,52 +554,68 @@ def get_confidence_weighted_entropy_for_each_label(
...
 
     Returns
     -------
-    avg_nbrs_distances : np.array
-      Average distance to k-nearest neighbors for each datapoint which is used as a score for OOD detection.
+    avg_knn_distances : np.ndarray
+      Average distance to k-nearest neighbors for each example which is used as a score for outlier detection.
     """
+    # if knn is not provided, then use default KNN as estimator
+    if knn is None:
```
jwmueller, line 597: In the case where `knn is None` would be good to raise ValueError if `k > len(features)`. While we could also make the code for that scenario, I don't think it is worth supporting.

ulya-tkch:
How should we handle type check error:
``features`` has type ``np.ndarray`` but we set ``features=None`` to avoid duplicate fitting. Should we set default none for features? Is it bad our function has all defaults?

anishathalye:
Setting defaults is fine, and having a default be `None` is fine, but this should be expressed in the type system. You can do that using [`Optional`](https://docs.python.org/3/library/typing.html#typing.Optional).

```python
from typing import optional

x: Optional[int] = None  # this type checks
x = 3  # and so does this

def f(y: Optional[int]):
  if y is None:
    return 42
  else:
    return y + 1  # this type checks thanks to flow-sensitive typing
```

jwmueller:
> Setting defaults is fine, and having a default be None is fine, but this should be expressed in the type system. You can do that using [Optional](https://docs.python.org/3/library/typing.html#typing.Optional).

But that is not exactly what's going on here because this `features` variable is not optional. It is being used as a function argument and then being sometimes set to None inside the function before it is passed to another external library's function.  Wouldn't declaring `features` as Optional type be misleading because it actually cannot be None for our function, we are just sometimes setting it to None inside of our function after we've utilized it. Not sure what the best practice for that with python typing would be, still to declare it as `Optional`?

cleanlab/rank.py
```diff
@@ -556,52 +554,73 @@ def get_confidence_weighted_entropy_for_each_label(
...
 
-    nbrs : sklearn.neighbors.NearestNeighbors
-      Instantiated NearestNeighbors class object that's been fitted on a dataset in the same feature space.
+    knn : sklearn.neighbors.NearestNeighbors, default = None
+      Instantiated ``NearestNeighbors`` class object that's been fitted on a dataset in the same feature space.
       Note that the distance metric and n_neighbors is specified when instantiating this class.
+      You can also pass in a subclass of ``sklearn.neighbors.NearestNeighbors`` which allows you to use faster
+      approximate neighbor libraries as long as you wrap them behind the same sklearn API.
+      If you specify ``knn`` here and wish to find outliers in the same data you already passed into ``knn.fit(features)``, you should specify ``features = None`` here if your ``knn.kneighbors(None)``
+      returns the distances to the datapoints it was ``fit()`` on.
```
anishathalye, line 579: I found the above sentence confusing. Is the user ever supposed to specify `features=None`?

anishathalye (commented):
I hadn't read the code carefully / understood the API. Based on the comment that I found confusing: is the user ever supposed to set `features=None`? If yes, the type in the function argument should be `Optional[np.ndarray]`, otherwise it can stay `np.ndarray`. If we only set `features=None` internally in certain cases, then we should use a different variable name internally. Obviously, choose a better variable name than this, but `features2: Optional[np.ndarray] = features; ...`.

If it's the case that either `features` or `knn` should be non-`None`, we could express that in the type system in a kind of clunky way (e.g. a single `Union[np.ndarray, NearestNeighbors]` argument), though I think the cleaner way is to have both be `Optional[_]` and then do a run-time check, `if features is not None: { do features stuff } elif knn is not None: { do knn stuff } else: raise ValueError('either features or knn must be non-None')`.

By the way, if we run mypy with [`--allow-redefinition`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-allow-redefinition), we can write code like this:

```python
from typing import Optional

# external API: x is an int
def f(x: int) -> int:
    # internally, f wants to use x as an Optional[int]
    x: Optional[int] = x
    if x == 42:
        x = None
    if x is None:
        return 0
    return 1
```

This type checks with `mypy --allow-redefinition`.

cleanlab/rank.py
ulya-tkch: Either features or knn can be ``None`` but both cannot be ``None`` at the same time. I like the idea of going with two Optional arguments and then adding a runtime check to make sure at least one of the values is not ``None``.

ulya-tkch:
TODO: reverse scoring for outlier severity

cleanlab/rank.py
```diff
@@ -556,52 +555,87 @@ def get_confidence_weighted_entropy_for_each_label(
...
+    k: int = 10,
+    t: int = 1,
 ) -> np.ndarray:
-    """Returns the KNN distance out-of-distribution (OOD) score for each datapoint.
+    """Returns the KNN distance outlier score for each example.
 
-    This is a function to compute OOD scores where higher scores indicate the datapoint is more likely to be OOD.
+    Returns an outlier score for each example based on its feature values, where higher scores indicate examples more
+    likely to be outliers. The score is based on the average distance between the example and its
+    K nearest neighbors in the dataset (in feature space).
```
jwmueller, line 568: Add sentence:

Scores lie in [0,1] with values near 0 indicating examples that are less typical in the dataset distribution.

cleanlab/rank.py
```diff
@@ -556,52 +555,87 @@ def get_confidence_weighted_entropy_for_each_label(
...
-      This is the "query set" of features for each datapoint which are used for nearest neighbor search.
+    features : np.ndarray
+      Feature array of shape ``(N, M)``, where N is the number of examples and M is the number of features used to represent each example.
+      All features should be numeric. For unstructured data (eg. images, text, categorical values, ...), you should provide
+      vector embeddings to represent each example (e.g. extracted from some pretrained neural network).
 
-    nbrs : sklearn.neighbors.NearestNeighbors
-      Instantiated NearestNeighbors class object that's been fitted on a dataset in the same feature space.
+    knn : sklearn.neighbors.NearestNeighbors, default = None
+      Instantiated ``NearestNeighbors`` class object that's been fitted on a dataset in the same feature space.
```
jwmueller, line 578: "class object" -> "object"

cleanlab/rank.py
```diff
@@ -556,52 +555,87 @@ def get_confidence_weighted_entropy_for_each_label(
...
-      If k=None, then by default k=min(10, max_k) is used where max_k is extracted from the given nbrs.
+    k : int, default=10
+      Number of neighbors to use when calculating outlier score (average distance to neighbors). If an existing ``knn``
+      object is provided, you can still specify that outlier score should use a different value of ``k`` than originally
+      used in the ``knn``, as long as your specified value of ``k`` is smaller than the one originally used.
+      If k is larger, then by default ``k = n_neighbors`` where ``n_neighbors`` is defined in the ``knn``.
+      If k is not provided, then by default ``k = 10``.
+
+    t : int, default=1
+      Parameter controlling magnitude of similarity score transformation. Transformation is ``exp(-x*t)`` where ``x`` is the average distance to k-nearest neighbors.
```
jwmueller, line 596: Controls transformation of distances between examples into similarity scores that lie in [0,1]. The transformation applied is ...

cleanlab/rank.py
```diff
@@ -556,52 +555,87 @@ def get_confidence_weighted_entropy_for_each_label(
...
 
     Returns
     -------
-    avg_nbrs_distances : np.array
-      Average distance to k-nearest neighbors for each datapoint which is used as a score for OOD detection.
+    avg_knn_distances : np.ndarray
+      Similarity score for each example where score shows likeleyhood example is not an outlier. Smaller scores mean example more likeley to be an outlier. Similarity score is a transformation of each example's average distance to k-nearest neighbors.
```
jwmueller, line 601: Description of the scores are already given above so let's just shorten this to:

Score for each example that roughly corresponds to the likelihood this example stems from the same distribution as the dataset ``features`` (i.e. is not an outlier).

cleanlab/rank.py
```diff
@@ -556,52 +555,87 @@ def get_confidence_weighted_entropy_for_each_label(
...
+    # neighbor of each point is the point itself, at a distance of zero.
+    distances, _ = knn.kneighbors(features)
 
     # Calculate average distance to k-nearest neighbors
-    avg_nbrs_distances = distances[:, :k].mean(axis=1)
+    avg_knn_distances = distances[:, :k].mean(axis=1)
 
-    return avg_nbrs_distances
+    # Global rescale outlier_scores to range 0-1 with 0 = most concerning
+    distance_to_score = np.exp(-1 * avg_knn_distances * t)
```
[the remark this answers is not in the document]
ulya-tkch: Changed to outlier_scores to agree with 'get_ outlier_scores' function naming convention

cleanlab/rank.py
```diff
@@ -556,52 +555,87 @@ def get_confidence_weighted_entropy_for_each_label(
...
+      If k is not provided, then by default ``k = 10``.
+
+    t : int, default=1
+      Parameter controlling magnitude of similarity score transformation. Transformation is ``exp(-x*t)`` where ``x`` is the average distance to k-nearest neighbors.
 
     Returns
     -------
-    avg_nbrs_distances : np.array
-      Average distance to k-nearest neighbors for each datapoint which is used as a score for OOD detection.
+    avg_knn_distances : np.ndarray
```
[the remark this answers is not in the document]
ulya-tkch: [nit] So I did this but I don't think it is intuitive to have an array called "outlier_scores" where higher score is less likeley to be an outlier. Something like "in_distribution_scores" seems better. Only raising this because this array name will be in the function description and may be confusing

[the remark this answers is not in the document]
ulya-tkch: Renaming would probably require renaming entire function though

tests/test_rank.py
```diff
@@ -403,42 +392,139 @@ def test_get_knn_distance_ood_scores():
...
 
-    # Create OOD datapoint
+@pytest.mark.filterwarnings("ignore::UserWarning")
+def test_default_k_and_model_get_outlier_scores():
+    # Testing using 'None' as model param and correct setting of default k as max_k
+
+    # Create dataset with OOD example
+    X = data["X_test"]
     X_ood = np.array([[999999999.0, 999999999.0]])
+    X_with_ood = np.vstack([X, X_ood])
```
jwmueller, line 439: maybe I'm missing something but it seems none of the unit tests are explicitly verifying that X_ood has the smallest outlier score among all the datapoints? 

We should definitely explicitly verify that when `X_with_ood` is passed in, the final index of `outlier_scores` contains the smallest value in this vector.

tests/test_rank.py
```diff
@@ -403,42 +392,139 @@ def test_get_knn_distance_ood_scores():
...
+            k=15,  # this should throw TypeError because knn=None and features=None
+        )
+    except Exception as e:
+        assert "Both knn and features arguments" in str(e)
+        with pytest.raises(ValueError) as e:
+            rank.get_outlier_scores(
+                features=None,
+                knn=None,
+                k=15,  # this should throw TypeError because knn=None and features=None
+            )
```
jwmueller, line 518: This seems basically the same logic as the below block except for the `k=15` argument. Is it really testing something different and thus worth keeping?

ulya-tkch: Testing different k does not affect exception throw but there is no real reason for this (pretty sure it doesnt/wouldnt)

tests/test_rank.py
ulya-tkch: This was being checked using argsort but I made it more readable!

tests/test_rank.py
jwmueller: Ya let’s get rid of this block then. It’s a little confusing to have two different blocks that are testing the same thing

tests/test_rank.py
jwmueller: Ah got it, looks fine sorry I missed it!

jwmueller (approved):
Amazing work on this!

PR #306, opened 2022-07-06
Add static type checking

anishathalye:
Our package doesn't have type annotations everywhere, so we can't use
mypy in strict mode just yet. Still, adding type checking in CI is
valuable, so we don't have unchecked annotations in our code.

This patch includes basic fixes to make type checking pass, including
switching the incorrect `np.array` type annotation for `np.ndarray` and
adding some assertions for flow-sensitive typing.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -1,0 +1,22 @@
+[mypy]
+
+[mypy-sklearn.*]
+ignore_missing_imports = True
+
+[mypy-scikeras.*]
+ignore_missing_imports = True
+
+[mypy-tensorflow.*]
+ignore_missing_imports = True
+
+[mypy-fasttext.*]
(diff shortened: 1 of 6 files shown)


.mypy.ini
```diff
@@ -0,0 +1,22 @@
+[mypy]
```
jwmueller, line 1: Add instructions/comment about this in DEVELOPMENT.md ?

jwmueller (approved):
Great catches!

.mypy.ini
anishathalye: I don't think there's much to say, it shouldn't come up for the average developer, and when it does, mypy prints instructions / it's pretty standard to add ignored libraries when they don't have type stubs.

Issue #288, opened 2022-06-24
Give better error messages when a class has no examples in `find_label_issues`


anishathalye:
Currently, we produce various internal errors when a class is not represented. We should either support missing classes, or we should produce something like a `ValueError("find_label_issues requires that each class has at least one example")`.

Right now, we produce a variety of internal errors (data-dependent).

# Steps to reproduce

## Example 1

```python
import cleanlab
import numpy as np

labels = np.array([0, 0, 1, 1])
pred_probs = np.array([[0.1, 0.7, 0.2], [0.1, 0.8, 0.1], [0.7, 0.2, 0.1], [0.8, 0.1, 0.1]])
issues = cleanlab.filter.find_label_issues(labels, pred_probs, n_jobs=1)
```

Produces the result:

```
Traceback (most recent call last):
  File ".../error.py", line 6, in <module>
    issues = cleanlab.filter.find_label_issues(labels, pred_probs)
  File ".../src/cleanlab/env/lib/python3.9/site-packages/cleanlab/filter.py", line 457, in find_label_issues
    confident_joint, cl_error_indices = compute_confident_joint(
  File ".../src/cleanlab/env/lib/python3.9/site-packages/cleanlab/count.py", line 435, in compute_confident_joint
    pred_probs_bool = pred_probs >= thresholds - 1e-6
ValueError: operands could not be broadcast together with shapes (4,3) (2,)
```

## Example 2

```python
import cleanlab
import numpy as np

labels = np.array([0, 0])
pred_probs = np.array([[0.3, 0.7], [0.2, 0.8]])
issues = cleanlab.filter.find_label_issues(labels, pred_probs, n_jobs=1)
```

Produces the result:

```
Traceback (most recent call last):
  File ".../error.py", line 6, in <module>
    issues = cleanlab.filter.find_label_issues(labels, pred_probs, n_jobs=1)
  File ".../src/cleanlab/env/lib/python3.9/site-packages/cleanlab/filter.py", line 561, in find_label_issues
    label_issues_masks_per_class = [_prune_by_count(k, args) for k in range(K)]
  File ".../src/cleanlab/env/lib/python3.9/site-packages/cleanlab/filter.py", line 561, in <listcomp>
    label_issues_masks_per_class = [_prune_by_count(k, args) for k in range(K)]
  File ".../src/cleanlab/env/lib/python3.9/site-packages/cleanlab/filter.py", line 204, in _prune_by_count
[...]

Issue #44, opened 2020-09-10
Can the class number computed from psx not from s


Mickey-Guo:
Thanks for your contributions!
In the function `cleanlab.latent_estimation.compute_confident_joint`, if K is None, I wonder if I can replace this line with `K = np.array(psx).shape[1]` to avoid some bugs if one or more classes in the dataset is of zero sample.
And there may be other lines using this kind of code. Is the toolkit compatible with the situation above? Thanks!

cgnorthcutt:
Cleanlab requires at least 1 example in every class, otherwise there is nothing to train on.

Dealing with inputs like yours is on the road map, but for now, you need to re format your input to be labels from 0, 1, 2... And psx and s must not have any zero size classes.

Mickey-Guo:
> Cleanlab requires at least 1 example in every class, otherwise there is nothing to train on.
> 
> Dealing with inputs like yours is on the road map, but for now, you need to re format your input to be labels from 0, 1, 2... And psx and s must not have any zero size classes.

That is. If `s` has no zero size class but the predicted result `psx` does, which I guess would happen if some class size is rather small, cleanlab cannot handle this situation for now. Right?

cgnorthcutt:
Thanks for the follow-up. Cleanlab will never remove less than 5 examples from any class. It's guaranteed. So could explain further?

Mickey-Guo:
Thanks for your answer. I can understand this as I notice `MIN_NUM_PER_CLASS = 5` is defined in pruning.py to guarantee this.
I found another problem that might happen if I manually set thresholds rather than calculate them from psx. If I set some class’s threshold very high like 0.999, because of the following code in the function `cleanlab.latent_estimation.compute_confident_joint`:
```
y_confident = true_label_guess[at_least_one_confident]
s_confident = s[at_least_one_confident]
confident_joint = confusion_matrix(y_confident, s_confident).T 
```
The union set of `y_confident` and `s_confident` may not cover all classes. This would result in wrong size confusion matrix. And the following step of Confident Learning would fail. 
So I think if I want to set thresholds manually, I have to make sure that `y_confident` and `s_confident` would cover all classes. What do you think about it?

cgnorthcutt:
Hi @Mickey-Guo  thanks for this feedback. Can you share a complete, minimum code example to achieve this error?

Slicerkao:
[...]

PR #308, opened 2022-07-09
error for missing classes, consistency on determining num_classes, code cleanup 

jwmueller:
1.  Adds explicit informative error statements when there is is missing class in labels 
I left # TODO comments in there for this logic so it is easily removed when we support missing classes @cgnorthcutt  (just search for: 'missing classes' in the code).

Addresses issue:  https://github.com/cleanlab/cleanlab/issues/288

2.  Aims to use the `get_num_classes()` helper function to establish the number of classes throughout the codebase. This way we can ensure consistency.  Since `get_num_classes` refers to `pred_probs` as its number 1 source of truth for the number of classes, this addresses issue:
https://github.com/cleanlab/cleanlab/issues/44

3.  Miscellaneous code cleanup.

### DIFF
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -39,6 +39,7 @@ from cleanlab.internal.util import (
     csr_vstack,
     append_extra_datapoint,
     train_val_split,
+    get_num_classes,
 )
 from cleanlab.internal.latent_algebra import (
     compute_inv_noise_matrix,
@@ -276,8 +277,8 @@ def _compute_confident_joint_multi_label(
         of confident joint as a baseline proxy for the label issues. This
         sometimes works as well as filter.find_label_issues(confident_joint)."""
 
-    # [0, 1, 2, ... num_classes - 1] Assumes all classes are represented in labels
-    unique_classes = range(pred_probs.shape[1])
+    # Assumes all classes are represented in labels: [0, 1, 2, ... num_classes - 1]
+    unique_classes = range(get_num_classes(labels=labels, pred_probs=pred_probs))
     if thresholds is None:
         # the avg probability of class given that the label is represented.
         thresholds = get_confident_thresholds(labels, pred_probs, multi_label=True)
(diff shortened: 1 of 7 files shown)


cleanlab/internal/label_quality_utils.py
```diff
@@ -85,19 +85,16 @@ def get_normalized_entropy(pred_probs: np.ndarray, min_allowed_prob: float = 1e-
...
       Minimum allowed probability value. Entries of `pred_probs` below this value will be clipped to this value.
       Ensures entropy remains well-behaved even when `pred_probs` contains zeros.
     Returns
     -------
     entropy : np.array (float)
-
+      Each element is the normalized entropy of the corresponding row of ``pred_probs``.
     """
 
     num_classes = pred_probs.shape[1]
```
huiwengoh, line 100: This should also call `get_num_classes()`

cleanlab/count.py
```diff
@@ -1176,8 +1178,10 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
     """
 
     if multi_label:
-        # [0, 1, 2, ... num_classes - 1]
-        unique_classes = range(pred_probs.shape[1])  # Assumes all classes are represented in labels
+        # Assumes all classes are represented in labels: [0, 1, 2, ... num_classes - 1]
+        unique_classes = range(
```
huiwengoh, line 1182: I think `unique_classes` can be defined before the if/else block, I scrolled down the code a couple lines and right now `pred_probs.shape[1]` is still being called in the else block (should also be changed to `get_num_classes()` but probably was just missed out in your previous commit), but since they both use `unique_classes`, moving it out would be less code duplication?

huiwengoh (changes requested):
Just pointed out a couple of missed changes!

cleanlab/internal/label_quality_utils.py
jwmueller: I actually purposefully left the num_classes determination here based on pred-probs because here the "number of classes" is not the same as the official number of classes the cleanlab package determines in a user's dataset (note this function doesn't even take in any user data/labels).  In this function, the `num_classes` variable is just a placeholder the number of columns of `pred_probs.shape`

cleanlab/count.py
jwmueller: good catch!

huiwengoh (approved):
I think it looks good!

PR #310, opened 2022-07-14
Added outlier detection tutorial into docs

ulya-tkch:


cgnorthcutt:
Do we know this outperforms OOD (out of distribution) detection using confident learning? The baseline being to simply keep track of examples where no probability for any class exceeds the class threshold and a better baseline being a hyper-parameter that determines what fraction of the class thresholds to use (instead of the average, some percentile).

Because thats an easy baseline and also more natural with the current package offering, it would make sense to first benchmark that before sharing this as the tutorial for 'how to ood with cleanlab'

docs/source/tutorials/knn_outlier.ipynb
```diff
@@ -0,0 +1,447 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "c1f65acd",
+   "metadata": {},
+   "source": [
+    "# Detecting Outliers with Cleanlab and Sklearn\n",
```
jwmueller, line 8: "Cleanlab and Sklearn"
-> 
Cleanlab and PyTorch Image Models (timm)

docs/source/tutorials/knn_outlier.ipynb
```diff
@@ -0,0 +1,447 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7b552a12",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Package installation (hidden on docs website).\n",
+    "# If running on Colab, may want to use GPU (select: Runtime > Change runtime type > Hardware accelerator > GPU)\n",
+    "\n",
+    "dependencies = [\"cleanlab\", \"matplotlib\", \"torch\", \"pylab\", \"keras\", \"sklearn\", \"timm\", ]\n",
```
jwmueller, line 48: I don't think `pylab` is recommended way to install / work with matplotlib anymore these days?  I may be wrong, but would stick with matplotlib's official recommendation on how to import / install it

docs/source/tutorials/knn_outlier.ipynb
```diff
@@ -0,0 +1,447 @@
...
+    "\n",
```
jwmueller, line 49: Sometimes these tutorials can break if the dependency packages get updated in some way that is not compatible with the other dependencies.

So from now on, let's start keeping track of which exact package versions the author of tutorial is using.  Can you add a comment here:

```
# package versions we used: matplot==X.X.X, torch==Y.Y.Y, ...
```

docs/source/tutorials/knn_outlier.ipynb
```diff
@@ -0,0 +1,447 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "802dece2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%time\n",
```
jwmueller, line 256: Don't need to time our cells in the tutorials. These are the same notebooks that will be displayed at docs.cleanlab.ai, so they should be as concise as possible.

docs/source/tutorials/knn_outlier.ipynb
```diff
@@ -0,0 +1,447 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "afd151d5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%time\n",
+    "import timm\n",
```
jwmueller, line 239: move imports to start of tutorial, so if it fails, user can first install the requisite packages rather than only realizing half-way through that a dependency is missing.

docs/source/tutorials/knn_outlier.ipynb
```diff
@@ -0,0 +1,447 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1ae35d12",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Load cifar10 datasets\n",
```
jwmueller, line 127: Is it possible to get the datasets from torchvision instead of keras?  torchvision would be a more natural dependency for users to install with torch + timm, especially we can be more sure that future versions of torchvisions will work with existing/past/future versions of torch/timm.

No worries if it's a lot of work to change this, hopefully future versions of keras.datasets will play nice with various versions of torch & timm.  In the past, I've seen incompatibility issues with having multiple DL frameworks installed from different version.

docs/source/tutorials/knn_outlier.ipynb
```diff
@@ -0,0 +1,447 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%time\n",
+    "import timm\n",
+    "\n",
+    "# This cell can take ~1-2 mins\n",
+    "model = timm.create_model('resnet50', pretrained=True, num_classes=0) # download the model from timm\n",
+    "trainX_torch = torch.from_numpy(trainX.swapaxes(1,3)) # turn trainX into tensor and fix channel dimension\n",
+    "train_feature_embeddings = model(trainX_torch) # fit model to train images and get train feature embeddings\n",
```
jwmueller, line 244: One reason for the RAM issues is that you're doing a single forward pass through the net with all of the data at once. This should be broken up into a loop over small mini-batches of the data and the results concatenated back into a single array (especially important for GPU which has less RAM)

jwmueller: same for test data below, we always only want to forward pass a small mini-batch through the network, never an entire dataset

docs/source/tutorials/knn_outlier.ipynb
```diff
@@ -0,0 +1,447 @@
...
+   "source": [
+    "plot_images(trainX,trainy)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7cf11eb8",
+   "metadata": {},
+   "source": [
+    "The test set on the other hand still visibily contains the non-animal images: `[ship, airplane, automobile, truck]`. If we consider `trainX` to be the representation of data, then these non-animal images in test dataset `testX` become outliers."
```
jwmueller, line 207: careful about using term "representation", I'd only reserve that term as a synonym for the feature embeddings of the data

jwmueller: In this case, I'd do: "the representation of data"
->
"representative of the normal data distribution"

docs/source/tutorials/index.rst
```diff
@@ -12,4 +12,5 @@ Tutorials
    audio
    dataset_health
    pred_probs_cross_val
+   knn_outlier
```
jwmueller, line 15: Let's just rename it to more generic term: outliers

and the notebook to be outliers.ipynb

docs/source/tutorials/outliers.ipynb
```diff
@@ -0,0 +1,615 @@
...
+    "# E.g. if viewing master branch documentation:\n",
+    "#     !pip install git+https://github.com/cleanlab/cleanlab.git\n",
+    "```"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "48cb3aed",
+   "metadata": {},
```
jwmueller, line 76: Note that to ensure this cell is hidden on docs.cleanlab.ai, we want to replace metadata with:
```
"metadata": {
    "nbsphinx": "hidden"
   },
```

If you try building the docs locally, you should then see that this cell does not appear in the built docs html files

docs/source/tutorials/outliers.ipynb
```diff
@@ -0,0 +1,615 @@
...
+    "false_positive_idxs = set(outlier_indices).intersection(set(animal_idxs))\n",
+    "FPR = len(false_positive_idxs) / (len(false_positive_idxs) + len(animal_idxs))\n",
+    "print(f'Number of false positives detected: {len(false_positive_idxs)}\\nFalse positive rate: {round(FPR,4)} ')"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "740a8774",
+   "metadata": {},
```
jwmueller, line 578: want to also have this cell be hidden as demonstrated in above comment

docs/source/tutorials/outliers.ipynb
```diff
@@ -0,0 +1,629 @@
...
+   "metadata": {
+    "nbsphinx": "hidden"
+   },
+   "outputs": [],
+   "source": [
+    "# Package installation (hidden on docs website).\n",
+    "# If running on Colab, may want to use GPU (select: Runtime > Change runtime type > Hardware accelerator > GPU)\n",
+    "# Package versions we used: matplotlib==3.5.1, numpy==1.21.6, torch==1.11.0, scikit-learn==1.0.2, torchvision==0.12.0, timm==0.5.4, cleanlab==2.0.0\n",
+    "\n",
+    "dependencies = [\"matplotlib\", \"torch\", \"torchvision\", \"sklearn\", \"timm\"]\n",
```
jwmueller, line 85: BTW we generally want to include `cleanlab` in the dependencies variable.
The cell below will check if each of the dependencies has been installed and tell the user they should install it if not (so would be good to also check this for cleanlab)

docs/source/tutorials/outliers.ipynb
```diff
@@ -0,0 +1,627 @@
...
+   "id": "740a8774",
+   "metadata": {
+    "nbsphinx": "hidden"
+   },
+   "outputs": [],
+   "source": [
+    "# Note: This cell is only for docs.cleanlab.ai, if running on local Jupyter or Colab, please ignore it.\n",
+    "num_animals = 0\n",
+    "for i in range(15):\n",
+    "    _,y = top_outlier_subset[0]\n",
```
jwmueller, line 597: just FYI there's small bug here I believe; should be `i` not `0` right?

jwmueller: Also just FYI it's typically not a good idea to hardcode a number like 15 as done here. 
Instead all 15s would be replaced by: `len(top_outlier_subset)`

jwmueller: Don't worry about these though, I'll make the updates in my future PR editing this notebook

ulya-tkch: Oh yeah i should be the value! 
I hardcoded 15 since 15 images are shown for the top outliers so it would make sense to make sure we're at least visualizing good data but len(top_outlier_subset) works the same for this case.

jwmueller (approved):
Good stuff! I'll make a follow-up PR to further improve some of the language / syntax

PR #315, opened 2022-07-21
Fix bugs in default KNN k-value

jwmueller:
Addressed subtle bugs in how default value of `k` was being set for outlier-scoring.

Thanks @elisno for finding these bugs!


ulya-tkch (approved):
Logic looks good! `If knn = None` and `elif k = None` might be slightly complicated logic but makes sense in this case.

Issue #307, opened 2022-07-07
Switch to strict static type checking


anishathalye:
#306 adds static type checking; it would be nice to switch over to mypy strict mode (`--strict`) once we have full coverage with type annotations.

elisno:
Maybe start the switch but allow specific warnings/errors?

E.g. the following only results in 27 errors overall:

```bash
mypy --install-types --non-interactive \
--strict --allow-untyped-defs --allow-untyped-calls --allow-any-generics \
--show-error-codes \
--exclude cleanlab/experimental \
cleanlab
```

Including the experimental features brings the total to 57 errors.

Once those errors have been dealt with, one can turn off the `--allow-X-Y` checks one by one?

jwmueller:
@elisno Would you like to help fix those 25 errors?  I agree that would be a good way to start imposing the strict mode!

elisno:
I'll see what I can do 👍

anishathalye:
Fantastic! Feel free to submit a PR that improves some subset of type annotations, you don't have to fix _everything_. Any improvement is appreciated :)

PR #317, opened 2022-07-21
validation.py: Annotate function args and return values

elisno:
Starting to fix strict mypy errors as discussed in #307.

This PR handles the internal.validation module:

I ran:
```bash
mypy --install-types --non-interactive --strict --allow-untyped-defs --allow-untyped-calls --allow-any-generics --show-error-codes --exclude cleanlab/experimental cleanlab/internal/validation.py
```

### Errors before PR

```
cleanlab/internal/validation.py:61: error: Argument 1 to "len" has incompatible type "Union[ndarray[Any, Any], generic]"; expected "Sized"  [arg-type]
cleanlab/internal/validation.py:72: error: Tuple index out of range  [misc]
cleanlab/internal/validation.py:74: error: Tuple index out of range  [misc]
Found 3 errors in 1 file (checked 1 source file)
```

### After PR

```
Success: no issues found in 1 source file
```

## Some thoughts

- I think (X, y) might need some custom Union type to handle both numpy arrays and pandas dataframe, etc.
- All of the "assert" functions return None.

### DIFF
--- a/cleanlab/internal/validation.py
+++ b/cleanlab/internal/validation.py
@@ -131,7 +131,7 @@ def assert_indexing_works(
         raise TypeError(msg)
 
 
-def labels_to_array(y: Union[list, np.ndarray, pd.Series, pd.DataFrame]) -> np.ndarray:
+def labels_to_array(y: Union[list, np.ndarray, np.generic, pd.Series, pd.DataFrame]) -> np.ndarray:
     """Converts different types of label objects to 1D numpy array and checks validity
 
     Parameters
--- a/cleanlab/internal/validation.py
+++ b/cleanlab/internal/validation.py
@@ -122,9 +122,9 @@ def assert_indexing_works(
         idx = [0, length_X - 1]
     try:
         if isinstance(X, (pd.DataFrame, pd.Series)):
-            _ = X.iloc[idx]
+            _ = X.iloc[idx]  # type: ignore
         else:
-            _ = X[idx]
+            _ = X[idx]  # type: ignore
     except:
         msg = "Data features X must support list-based indexing; i.e. one of these must work: \n"
         msg += "1)  X[index_list] where say index_list = [0,1,3,10], or \n"
--- a/cleanlab/internal/validation.py
+++ b/cleanlab/internal/validation.py
@@ -18,7 +18,7 @@
 Checks to ensure valid inputs for various methods.
 """
 
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Union
 import warnings
 import numpy as np
 import pandas as pd
@@ -26,7 +26,7 @@ import pandas as pd
 
 def assert_valid_inputs(
     X: Any,
-    y: np.ndarray,
+    y: Union[list, np.ndarray, pd.Series, pd.DataFrame],
     pred_probs: Optional[np.ndarray] = None,
     multi_label: bool = False,
 ) -> None:
@@ -88,7 +88,7 @@ def assert_valid_class_labels(y: np.ndarray) -> None:
     Assumes labels is 1D numpy array (not multi-label).
     """
     if y.ndim != 1:
-        raise ValueError("labels must by 1D numpy array.")
+        raise ValueError("labels must be 1D numpy array.")
 
     unique_classes = np.unique(y)
     if len(unique_classes) < 2:
@@ -131,10 +131,21 @@ def assert_indexing_works(
         raise TypeError(msg)
 
 
-def labels_to_array(y: np.ndarray) -> Any:
-    """Converts different types of label objects to 1D numpy array and checks validity"""
+def labels_to_array(y: Union[list, np.ndarray, pd.Series, pd.DataFrame]) -> np.ndarray:
+    """Converts different types of label objects to 1D numpy array and checks validity
+
+    Parameters
+    ----------
+    y : Union[list, np.ndarray, pd.Series, pd.DataFrame]
+        Labels to convert to 1D numpy array. Can be a list, numpy array, pandas Series, or pandas DataFrame.
+
+    Returns
+    -------
+    np.ndarray
+        1D numpy array of labels.
+    """
     if isinstance(y, pd.Series):
-        return y.values
+        return np.asarray(y.values)
     elif isinstance(y, pd.DataFrame):
         y = y.values
         if y.shape[1] != 1:
--- a/cleanlab/internal/validation.py
+++ b/cleanlab/internal/validation.py
@@ -21,7 +21,6 @@ Checks to ensure valid inputs for various methods.
 from typing import Any, List, Optional
 import warnings
 import numpy as np
-import numpy.typing as npt
 import pandas as pd
 
 
--- a/cleanlab/internal/validation.py
+++ b/cleanlab/internal/validation.py
@@ -18,6 +18,7 @@
 Checks to ensure valid inputs for various methods.
 """
 
+from cleanlab.typing import LabelLike, DatasetLike
 from typing import Any, List, Optional, Union
 import warnings
 import numpy as np
@@ -25,8 +26,8 @@ import pandas as pd
 
 
 def assert_valid_inputs(
-    X: Any,
-    y: Union[list, np.ndarray, pd.Series, pd.DataFrame],
+    X: DatasetLike,
+    y: LabelLike,
     pred_probs: Optional[np.ndarray] = None,
     multi_label: bool = False,
 ) -> None:
@@ -107,7 +108,7 @@ def assert_nonempty_input(X: Any) -> None:
 
 
 def assert_indexing_works(
-    X: Any, idx: Optional[List[int]] = None, length_X: Optional[int] = None
+    X: DatasetLike, idx: Optional[List[int]] = None, length_X: Optional[int] = None
 ) -> None:
     """Ensures we can do list-based indexing into ``X`` and ``y``.
     length_X is argument passed in since sparse matrix ``X``
@@ -131,12 +132,12 @@ def assert_indexing_works(
         raise TypeError(msg)
 
 
-def labels_to_array(y: Union[list, np.ndarray, np.generic, pd.Series, pd.DataFrame]) -> np.ndarray:
+def labels_to_array(y: Union[LabelLike, np.generic]) -> np.ndarray:
     """Converts different types of label objects to 1D numpy array and checks validity
 
     Parameters
     ----------
-    y : Union[list, np.ndarray, pd.Series, pd.DataFrame]
+    y : Union[LabelLike, np.generic]
         Labels to convert to 1D numpy array. Can be a list, numpy array, pandas Series, or pandas DataFrame.
 
     Returns
--- a/cleanlab/internal/validation.py
+++ b/cleanlab/internal/validation.py
@@ -18,12 +18,19 @@
 Checks to ensure valid inputs for various methods.
 """
 
+from typing import Any, List, Optional
 import warnings
 import numpy as np
+import numpy.typing as npt
 import pandas as pd
 
 
-def assert_valid_inputs(X, y, pred_probs=None, multi_label=False):
+def assert_valid_inputs(
+    X: npt.NDArray,
+    y: npt.NDArray,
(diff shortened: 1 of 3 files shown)


jwmueller:
For args that can be multiple types, I believe you can do: `from typing import Union`
and then declare `Union[Type1, Type2]` as their type, or even `Optional[Union[Type1, Type2]`.

cleanlab/internal/validation.py
```diff
@@ -18,12 +18,18 @@
...
 
+from typing import Any, List, Optional
 import warnings
 import numpy as np
 import pandas as pd
 
 
-def assert_valid_inputs(X, y, pred_probs=None, multi_label=False):
+def assert_valid_inputs(
+    X: np.ndarray,
```
jwmueller, line 28: Let's make this type `Any` because we want to support as many data formats as possible.

cleanlab/internal/validation.py
```diff
@@ -18,12 +18,18 @@
...
+    y: np.ndarray,
```
jwmueller, line 29: the possible valid types for `y` would include: 
list, np.ndarray, np.generic, pd.Series, pd.DataFrame

cleanlab/internal/validation.py
```diff
@@ -95,12 +101,14 @@ def assert_valid_class_labels(y):
...
 
-def assert_nonempty_input(X):
+def assert_nonempty_input(X: Optional[np.ndarray]) -> None:
     if X is None:
         raise ValueError("Data features X cannot be None. Currently X is None.")
 
 
-def assert_indexing_works(X, idx=None, length_X=None):
+def assert_indexing_works(
+    X: np.ndarray, idx: Optional[List[int]] = None, length_X: Optional[int] = None
```
jwmueller, line 110: Type of `X` should by `Any`

cleanlab/internal/validation.py
```diff
@@ -123,7 +131,7 @@ def assert_indexing_works(X, idx=None, length_X=None):
         raise TypeError(msg)
 
 
-def labels_to_array(y):
+def labels_to_array(y: np.ndarray) -> Any:
```
jwmueller, line 134: the possible valid types for `y` would include:
list, np.ndarray, np.generic, pd.Series, pd.DataFrame

cleanlab/internal/validation.py
```diff
@@ -95,12 +101,14 @@ def assert_valid_class_labels(y):
         raise TypeError(msg)
 
 
-def assert_nonempty_input(X):
+def assert_nonempty_input(X: Optional[np.ndarray]) -> None:
```
jwmueller, line 104: Type of `X` should by `Any`

cleanlab/internal/validation.py
elisno: mypy isn't happy about using `len(y: np.generic)`, as it's undefined.

elisno: I'll start with everything except np.generic for now.

I'm trying not to change anything in the function bodies to avoid breaking any assumptions that are made in the rest of the codebase.

cleanlab/internal/validation.py
```diff
@@ -18,12 +18,18 @@
...
 
+from typing import Any, List, Optional, Union
 import warnings
 import numpy as np
 import pandas as pd
 
 
-def assert_valid_inputs(X, y, pred_probs=None, multi_label=False):
+def assert_valid_inputs(
+    X: Any,
```
anishathalye, line 28: I understand why we're using `Any` here, because we don't want to spell out all the different possibilities / we want to allow certain duck-typed things without having Cleanlab depend on extra libraries, but "Any" doesn't read great in type annotations. What if we create a type alias for this, perhaps in a new `cleanlab.typing` module, and call it `dataset_like` (which will just be Any):

```python
from typing import TypeAlias

dataset_like: TypeAlias = Any
```

cleanlab/internal/validation.py
```diff
@@ -18,12 +18,18 @@
...
+    y: Union[list, np.ndarray, pd.Series, pd.DataFrame],
```
anishathalye, line 29: Maybe worth introducing an alias for this too?

```python
label_like: TypeAlias = Union[list, np.ndarray, pd.Series, pd.DataFrame]
```

cleanlab/internal/validation.py
```diff
@@ -95,12 +101,14 @@ def assert_valid_class_labels(y):
...
 
-def assert_nonempty_input(X):
+def assert_nonempty_input(X: Any) -> None:
     if X is None:
         raise ValueError("Data features X cannot be None. Currently X is None.")
 
 
-def assert_indexing_works(X, idx=None, length_X=None):
+def assert_indexing_works(
+    X: Any, idx: Optional[List[int]] = None, length_X: Optional[int] = None
```
anishathalye, line 110: If we make the above change, this will also change to `dataset_like`

cleanlab/internal/validation.py
```diff
@@ -123,10 +131,21 @@ def assert_indexing_works(X, idx=None, length_X=None):
         raise TypeError(msg)
 
 
-def labels_to_array(y):
-    """Converts different types of label objects to 1D numpy array and checks validity"""
+def labels_to_array(y: Union[list, np.ndarray, np.generic, pd.Series, pd.DataFrame]) -> np.ndarray:
```
anishathalye, line 134: `label_like`

cleanlab/internal/validation.py
```diff
@@ -123,10 +131,21 @@ def assert_indexing_works(X, idx=None, length_X=None):
...
+    """Converts different types of label objects to 1D numpy array and checks validity
+
+    Parameters
+    ----------
+    y : Union[list, np.ndarray, pd.Series, pd.DataFrame]
```
anishathalye, line 139: would want to update this as well

anishathalye (commented):
Looks great overall; only one recommendation: let's consider using type aliases to increase clarity and avoid code duplication.

anishathalye (commented):
Looking good! One final question/change requested.

@jwmueller, what do you think of this `LabelLike` / `DatasetLike` plan? Also, if we ever decide to include all the libraries like TensorFlow etc. (which we probably won't), we can always narrow the definition of `LabelLike` from `Any` to the more specific `Union[...]`.

jwmueller:
@anishathalye @elisno I also vote for the `LabelLike` / `DatasetLike` plan, where all `Union` and `Any` statements are deferred to the definition of the  `LabelLike` / `DatasetLike` types, so that the function args solely say:

``` X : DatasetLike, 
      y : LabelLike,
```

jwmueller:
@anishathalye should the type names actually be lower-case like in your earlier comments?  I.e. 
```
dataset_like, label_like
```
vs.
```
DatasetLike, LabelLike
```

I'm not sure which is more Pythonic, but I've definitely encountered the lower-case ones more frequently.

elisno:
Glad you pointed out the capitalization. I used this as a reference


https://peps.python.org/pep-0484/#type-aliases

Maybe we want to look into `TypeVar` for parametrizing the generic types?

cleanlab/typing.py
```diff
@@ -0,0 +1,10 @@
+from typing import Any, Union
+import numpy as np
+import pandas as pd
+
+LabelLike = Union[list, np.ndarray, pd.Series, pd.DataFrame]
```
anishathalye, line 5: Can we type these as type aliases? (https://github.com/cleanlab/cleanlab/pull/317#discussion_r927905542)

cleanlab/typing.py
```diff
@@ -0,0 +1,10 @@
...
+"""Type for objects that behave like collections of labels."""
+
+
+DatasetLike = Any
```
anishathalye, line 9: Same thing as above here.

anishathalye (commented):
@jwmueller I think Python isn't super consistent with formatting here. Many of the types come from class names, so those are upper camel case, and the typing package seems to use upper camel case as well, so I'd prefer that here (so `LabelLike` and `DatasetLike`).

As far as generics, maybe that will be nice to add at some point, but we could do that in a separate PR / is lower priority than getting typing coverage over the whole package (have it pass `mypy --strict`).

cleanlab/typing.py
elisno: Only for Python 3.10.

elisno: https://docs.python.org/3/library/typing.html#typing.TypeAlias

elisno:
```python
>>> import sys
>>> sys.version
'3.8.5 (default, Sep  4 2020, 07:30:14) \n[GCC 7.3.0]'
>>> from typing import TypeAlias
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'TypeAlias' from 'typing'
```

elisno: @anishathalye Did you have something like this in mind?

```python
# Import TypeAlias on Python 3.10+
try:
    from typing import TypeAlias
except ImportError:
    TypeAlias = type(None)


LabelLike: TypeAlias = Union[list, np.ndarray, pd.Series, pd.DataFrame]
DatasetLike: TypeAlias = Any
```

anishathalye: Ah I didn't realize that was only added in 3.10. We want to support 3.6+ without too much compatibility code, so let's stick with the current code in the PR.

cleanlab/internal/validation.py
```diff
@@ -123,10 +131,21 @@ def assert_indexing_works(X, idx=None, length_X=None):
...
+        Labels to convert to 1D numpy array. Can be a list, numpy array, pandas Series, or pandas DataFrame.
+
+    Returns
+    -------
+    np.ndarray
+        1D numpy array of labels.
+    """
     if isinstance(y, pd.Series):
-        return y.values
+        return np.asarray(y.values)
```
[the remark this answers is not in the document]
anishathalye: Looking at the docs for `pd.Series.values`, [pd.Series.to_numpy()](https://pandas.pydata.org/docs/reference/api/pandas.Series.to_numpy.html#pandas.Series.to_numpy) might be better to use here.

cleanlab/internal/validation.py
anishathalye: No, I don't think we need type aliases for those. Probabilities are floats (that are between 0 and 1), scores are floats (that are between 0 and 1), a joint is a 2d numpy array of floats.

cleanlab/internal/validation.py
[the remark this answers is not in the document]
elisno:
```python
return y.values.to_numpy()
```
these mypy errors

```
cleanlab/internal/validation.py:151: error: Returning Any from function declared to return "ndarray[Any, Any]"
cleanlab/internal/validation.py:151: error: Item "ndarray[Any, Any]" of "Union[ExtensionArray, ndarray[Any, Any]]" has no attribute "to_numpy"
```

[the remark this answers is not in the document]
anishathalye: Oh I think it's just `y.to_numpy()`.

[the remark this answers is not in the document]
elisno: Down to one error:

```
cleanlab/internal/validation.py:151: error: Returning Any from function declared to return "ndarray[Any, Any]"
```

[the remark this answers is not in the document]
elisno: Maybe turn this return value into a variable and annotate it explicitly?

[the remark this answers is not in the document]
elisno: Similary to what is done here:
https://github.com/jwmueller/cleanlab/blob/a0e1227ae5657e520e106552360072c39ea7fccb/cleanlab/rank.py#L647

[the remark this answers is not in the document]
elisno: This works:

```python
if isinstance(y, pd.Series):
        y_series: np.ndarray = y.to_numpy()
        return y_series
```

anishathalye (approved):
LGTM! Thanks for the PR. Will wait for Jonas to click the merge button.

jwmueller (approved):
LGTM, thanks for the useful contribution and your many good suggestions!

PR #318, opened 2022-07-22
Updating tutorials hyperlink to 2.0.0 release

aravindputrevu:
By default, the link redirects to the older version of [Image tutorial](https://docs.cleanlab.ai/stable/tutorials/image.html)

With this change, it is configured to the 2.0.0 release. 

However, in the long term, I believe two changes are necessary.

1. Anchor the hyperlink to tutorials subtopic in documentation instead of image tutorial. (this needs docs code/content change)
2. Setup a redirect like [docs.cleanlab.ai ](https://docs.cleanlab.ai) to [docs.cleanlab.ai/tutorial](https://docs.cleanlab.ai/tutorials).  (now, it gives 404)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ cl.predict(test_data)
 cleanlab.dataset.health_summary(labels, confident_joint=cl.confident_joint)
 ```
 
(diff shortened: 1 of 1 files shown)


jwmueller:
Thanks for the contribution @aravindputrevu !

However the links are correct as is, and there is just a minor bug in the banner on the docs page at "stable/" (if you look closely you'll see the banner is not necessary as this page is not actually outdated and is already the 2.0.0 documentation, which is the latest version)

PR #319, opened 2022-07-23
Allow KNN object to be returned by get_outlier_scores, Improved Tutorial

jwmueller:
I think it's important that the KNN object can be optionally returned, otherwise it gets too confusing for users to be able to easily find outliers in additional test data as well, unless they know how the internal KNN was originally constructed.


cleanlab/rank.py
```diff
@@ -640,4 +645,7 @@ def get_outlier_scores(
 
     # Map outlier_scores to range 0-1 with 0 = most concerning
     outlier_scores = np.exp(-1 * avg_knn_distances * t)
```
elisno, line 647: Looks like there's a type inference problem when returning `outlier_scores`.

Running mypy with:

```bash
mypy --install-types --non-interactive --strict --allow-untyped-defs --allow-untyped-calls --allow-any-generics cleanlab/rank.py
```

gives this error for the function:

```
cleanlab/rank.py:651: error: Returning Any from function declared to return "Union[ndarray[Any, Any], Tuple[ndarray[Any, Any], Any]]"
```

Annotating the variable seems to solve this issue:

```suggestion
    outlier_scores:np.ndarray = np.exp(-1 * avg_knn_distances * t)
```

jwmueller: good callout, thanks!

ulya-tkch (approved):
Change makes sense. Tutorial readability improved, I like the new headers that group the entire KNN section together. Test adds coverage. Nice work!

PR #322, opened 2022-07-24
Add y argument as alternative to labels in CleanLearning.fit()

### LINKED ISSUE #281: Change labels arg -> y in CleanLearning.fit()
`CleanLearning.fit()` argument names should match those of sklearn fit()

elisno:
Set `label` as an optional keyword-only argument.
Anyone still using it in this method should get a deprecation warning.

I'd like someone else's input on the same argument in `find_label_issues`, but will leave it be in this PR.

https://github.com/cleanlab/cleanlab/blob/5973a21958f47fd8e7246ccf36c8304ff4f22e3b/cleanlab/classification.py#L606

Fixes #281

### DIFF
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -297,6 +297,10 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
           An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
           Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
 
+          Note
+          ----
+          This argument was previously called ``labels`` in some previous versions of cleanlab.
+
         pred_probs : np.array, optional
           An array of shape ``(N, K)`` of model-predicted probabilities,
           ``P(label=k|x)``. Each row of this matrix corresponds
@@ -388,6 +392,15 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
           purposes). If you want to pass in validation data even in the final training of `clf.fit()`
           on the cleaned data subset, you should explicitly pass in that data yourself
           (eg. via `clf_final_kwargs` or `clf_kwargs`).
+
+        labels : np.array or pd.Series
+          An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
+          Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
+
+          Note
+          ----
+
+          This argument is kept for API backwards compatibility. You should use now use `y` instead.
 
         Returns
         -------
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -263,7 +263,7 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
     def fit(
         self,
         X,
-        labels,
+        labels=None,
         *,
         pred_probs=None,
         thresholds=None,
@@ -274,6 +274,7 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
         clf_kwargs={},
         clf_final_kwargs={},
         validation_func=None,
+        y=None,
     ):
         """
         Train the model `clf` with error-prone, noisy labels as if
@@ -388,6 +389,9 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
           on the cleaned data subset, you should explicitly pass in that data yourself
           (eg. via `clf_final_kwargs` or `clf_kwargs`).
 
+        y: np.array or pd.Series, optional
+          Alternative argument that can be specified instead of `labels`. Specifying `y` has the same effect as specifying `labels`, and is offered as an alternative for compatibility with sklearn.
+
         Returns
         -------
         CleanLearning
@@ -415,6 +419,13 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
             so that the classifier trains as if it had all the true labels,
             not just the subset of cleaned data left after pruning out the label issues.
         """
+
+        if labels is not None and y is not None:
+            raise ValueError("You must specify either `labels` or `y`, but not both.")
+        if y is not None:
+            labels = y
+        if labels is None:
+            raise ValueError("You must specify `labels`.")
 
         self.clf_final_kwargs = {**clf_kwargs, **clf_final_kwargs}
 
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -263,7 +263,7 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
     def fit(
         self,
         X,
-        labels,
+        y,
         *,
         pred_probs=None,
         thresholds=None,
@@ -274,6 +274,7 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
         clf_kwargs={},
         clf_final_kwargs={},
         validation_func=None,
+        labels=None,
     ):
         """
         Train the model `clf` with error-prone, noisy labels as if
@@ -292,7 +293,7 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
           The classifier that this instance was initialized with,
           ``clf``, must be able to fit() and predict() data with this format.
 
-        labels : np.array or pd.Series
+        y : np.array or pd.Series
           An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
           Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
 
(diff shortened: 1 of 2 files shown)


elisno:
Should something like this be added to the docstring? What SemVer should be here?

```
"""
.. deprecated:: 2.1.0
          `labels` will be removed in cleanlab 3.0.0. Use `y` instead to provide the labels.
"""
```

tests/test_classification.py
```diff
@@ -634,13 +634,13 @@ def dimN_data(N):
...
-    X, labels = dimN_data(N)
+    X, y = dimN_data(N)
     cl = CleanLearning(clf=ReshapingLogisticRegression())
     # just make sure we don't crash...
-    cl.fit(X, labels)
+    cl.fit(X, y)
     cl.predict(X)
     cl.predict_proba(X)
-    cl.score(X, labels)
+    cl.score(X, y)
```
jwmueller, line 644: We should have one unit test that explicitly checks the named arguments `y` and `labels` with both work by trying to call cl.fit() with each of them. Let's mark this unit test with a comment:

# TODO: remove this test once `labels` is deprecated.

cleanlab/classification.py
```diff
@@ -292,7 +293,7 @@ def fit(
           The classifier that this instance was initialized with,
           ``clf``, must be able to fit() and predict() data with this format.
 
-        labels : np.array or pd.Series
+        y : np.array or pd.Series
```
jwmueller, line 296: Note there are many other instances in the docstrings throughout this file (particularly near the top) which reference variable \`labels\`. We need to update all these references to be \`y\` now.

cleanlab/classification.py
```diff
@@ -416,6 +417,17 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
             not just the subset of cleaned data left after pruning out the label issues.
         """
 
+        if y is not None and labels is not None:
+            raise TypeError(
+                "CleanLearning.fit() received both `y` and `labels` arguments. "
+                "`labels` is deprecated, so please use `y` instead."
+            )
+        elif labels is not None:
+            warnings.warn("labels is deprecated; use y", DeprecationWarning, 2)
```
jwmueller, line 426: Update message to read:

"Argument `labels` is deprecated and will be removed in future versions, pass the labels via argument `y` instead."

cleanlab/classification.py
```diff
@@ -292,7 +293,7 @@ def fit(
           The classifier that this instance was initialized with,
           ``clf``, must be able to fit() and predict() data with this format.
 
-        labels : np.array or pd.Series
+        y : np.array or pd.Series
           An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
           Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
```
jwmueller, line 298: > Should something like this be added to the docstring? What SemVer should be here?

"""
.. deprecated:: 2.1.0
          `labels` will be removed in cleanlab 3.0.0. Use `y` instead to provide the labels.
"""

Instead of that, let's just add the following extra statement at the end of the docstring for argument `y`:

Note: this argument was previously called ``labels`` in some previous versions of cleanlab.

jwmueller (changes requested):
Looking pretty good! Just have some minor changes to add before we merge this

cgnorthcutt:
Left comment on the original issue #281

cleanlab/classification.py
cgnorthcutt: Note I just changed all these from labels to y for the 2.0 release and now that 2.0 is released you are recommending changing them all back. Let's discuss.

tests/test_classification.py
elisno: You're right. Both `y` and `labels` need to be optional arguments for the deprecated signature to be valid:
```
cl.score(X=X, labels=y)
```

Two thoughts:

1. How should it be reflected in the docstring that `y` isn't "really" optional?
2. Some tests in `test_classification.py` still use `labels`, e.g. https://github.com/cleanlab/cleanlab/blob/5973a21958f47fd8e7246ccf36c8304ff4f22e3b/tests/test_classification.py#L476-L481
Should DeprecationWarnings be filtered for those tests or the signatures updated (`labels` -> `y`)? I'll add a new unit test that checks all the signature variants.

cleanlab/classification.py
elisno: @jwmueller Thanks for pointing out the module docstring. References have been fixed there.
The only other references are scoped to different functions like `find_label_issues` and `_process_label_issues_arg`.
Shouldn't I leave probabilities like 
https://github.com/cleanlab/cleanlab/blob/5973a21958f47fd8e7246ccf36c8304ff4f22e3b/cleanlab/classification.py#L301
,
https://github.com/cleanlab/cleanlab/blob/5973a21958f47fd8e7246ccf36c8304ff4f22e3b/cleanlab/classification.py#L692
as is?


---


@cgnorthcutt Hopefully we can "keep" `labels` as a type hint later on.
See https://github.com/cleanlab/cleanlab/pull/317#issuecomment-1193201089.

cgnorthcutt:
@elisno @jwmueller  how about we simplify this commit by leaving labels as the default but add y=None at the end of the fit parameters. In other words, keep the API consistent (support labels everywhere in the package) and then add y=None. Docstring fopr y can say something like "Alias for labels parameter. Use this instead of labels when to enable full sklearn compliance and compatibility, i.e. allows you to use named parameter y=... in CleanLearning models used inside the sklearn functions"

Then we dont have to replace labels as many places or worry about consistency with methods outside of CleanLearning either.

elisno:
Good idea. The alias-approach is much cleaner than deprecating "new features" in 2.0.0.

cleanlab/classification.py
```diff
@@ -263,7 +263,7 @@ def __init__(
     def fit(
         self,
         X,
-        labels,
+        y=None,
```
jwmueller, line 266: @elisno After discussion with @cgnorthcutt we would like to do the following new plan. Sorry for the extra work and change of plans. Please ignore the old comments that contradict this new plan!

jwmueller: **New plan**:  Keep the `labels` arg as is. And keep `labels` throughout the documentation.  This keeps the CleanLearning code args the same as for the rest of the package. Especially for `CleanLearning.find_label_issues()` we prefer the argname `labels` since that matches the argname for `filter.find_label_issues()`.  So this basically corresponds to reverting all the changes and going back to the what the code looked like before. 

That said, we want this code to work with arbitrary sklearn code the user may be using, such that the user can drop in CleanLearning in place of the sklearn model without changing anything.  This means that at least `CleanLearning.fit()` must be able support the arg `y` in-case the user's sklearn code was doing something like: `SklearnEstimator.fit(X=...,y=...)`.

What this means is the only change we want this PR to add from what the code looked like before is support for an extra arg: `y`.

This extra arg should appear at the end of the args list of `CleanLearning.fit()` and it should have docstring that looks something like this: 
```
y: np.array or pd.Series, optional
  Alternative argument that can be specified instead of `labels`. Specifying `y` has the same effect as specifying `labels`, and is offered as an alternative for compatibility with sklearn. 
```
New plan:  Keep the `labels` arg as is. And keep `labels` throughout the documentation.  This keeps the CleanLearning code args the same as for the rest of the package. Especially for `CleanLearning.find_label_issues()` we prefer the argname `labels` since that matches the argname for `filter.find_label_issues()`.  So this basically corresponds to reverting all the changes and going back to the what the code looked like before. 

That said, we want this code to work with arbitrary sklearn code the user may be using, such that the user can drop in CleanLearning in place of the sklearn model without changing anything.  This means that at least `CleanLearning.fit()` must be able support the arg `y` in-case the user's sklearn code was doing something like: `SklearnEstimator.fit(X=...,y=...)`.

What this means is the only change we want this PR to add from what the code looked like before is support for an extra arg: `y`.

This extra arg should appear at the end of the args list of `CleanLearning.fit()` and it should have docstring that looks something like this: 
```
y: np.array or pd.Series, optional
  Alternative argument that can be specified instead of `labels`. Specifying `y` has the same effect as specifying `labels`, and is offered as an alternative for compatibility with sklearn. 
```

Your code should still raise ValueError if `y` and `labels` are both specified, stating only one should be specified.

All other functions in this file besides `fit()` should only take in arg `labels`, not `y`. So fit() should just internally convert `labels = y` as soon as it sees `y` was specified.

anishathalye: Also, I'd suggest you make `y` a [keyword-only argument](https://peps.python.org/pep-3102/).

elisno: Note that for this to work, `labels` would have to be "optional" to allow the use of `CleanLearning.fit(X=..., y=...)`

jwmueller: Yep exactly, we do need to change `labels` to be optional arg. Also should raise ValueError if `labels` and `y` are both None.

elisno: One thing I have yet to look into: How about writing a decorator for aliasing `labels`? That way it could be explicitly "required" in the signature.

ulya-tkch: Should we make this change throughout tutorials ex. image and tabular tutorials call  `find_label_issues(labels=y, ...)`

cgnorthcutt: @ulya-tkch i think its fine to still use `labels=labels`. but if your tutorial ever uses sklearn APIs on a CleanLearning object, then we must use `y=labels` to avoid an error/for compatibilty.

jwmueller (approved):
Thanks for seeing through all the edits!  Great contribution

PR #323, opened 2022-07-25
Updated tutorials to include quickstart

ulya-tkch:
* Added quickstart instructions at top of tutorials
* Added working package versions for tutorials
* TODO question: @jwmueller is the quickstart at the top of text.ipynb or audio.ipynb better? (i.e. should we show CleanLearning option in quickstart or just find_label_issues)
* TODO question:  Do we go with `y` or `labels` as input into these functions. If the tutorial later uses `y`, should we still use labels? Some tutorials use `y` some `labels`, should we make this consistent?


docs/source/tutorials/audio.ipynb
```diff
@@ -54,6 +84,8 @@
    "outputs": [],
    "source": [
     "# Package installation (hidden on docs website).\n",
+    "# Package versions used: tensorflow==2.8.1 numpy==1.21.6 torch==1.11.0 pandas==1.3.2 scikit-learn==1.0.2 \n",
+    "# pathlib==1.0.1 torchaudio==0.11.0 speechbrain==0.5.11 cleanlab==2.0.0 \n",
```
jwmueller, line 88: One thing to confirm is that the package versions you've listed in the notebooks do not contradict those in docs/requirements.txt.

Can you just quickly confirm this?  If they contract, I'd be inclined to go with the more recent version of the package (updating docs/requirements.txt as necessary).  Be super careful about this though, I purposefully downgraded some package versions as their latest versions caused errors when running the tutorials

docs/source/tutorials/audio.ipynb
```diff
@@ -19,6 +19,36 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert alert-info\">\n",
+    "Quickstart\n",
+    "<br/>\n",
+    "    \n",
+    "Do you already have true `labels` and `pred_probs` from a model trained on audio files? Run the code below to get ranked label issue indices.\n",
```
jwmueller, line 30: "Do you already" can just be "Already"

ulya-tkch:
TODO: remove cleanlab package version from package versioning

jwmueller:
> Is the quickstart at the top of text.ipynb or audio.ipynb better?

I actually like having both as is, good call!  It's nice to demonstrate both `filter.find_label_issues()` in one and `CleanLearning.find_label_issues()` in the other.

docs/source/tutorials/text.ipynb
```diff
@@ -24,6 +24,54 @@
...
+    "\n",
+    "<div  class=markdown markdown=\"1\" style=\"background:white;margin:16px\">  \n",
+    "    \n",
+    "```python\n",
+    "\n",
+    "from cleanlab.classification import CleanLearning\n",
+    "\n",
+    "cl = CleanLearning(model)\n",
+    "_ = cl.fit(data, labels)\n",
+    "label_issues = cl.get_label_issues()\n",
```
jwmueller, line 46: Maybe also add at the end: 

`preds = cl.predict(test_data)  # predictions from a version of your model trained on auto-cleaned data`

docs/source/tutorials/text.ipynb
```diff
@@ -24,6 +24,54 @@
...
+    "ranked_label_issues = find_label_issues(\n",
+    "    labels,\n",
+    "    pred_probs,\n",
+    "    return_indices_ranked_by=\"self_confidence\",\n",
+    ")\n",
+    "    \n",
+    "\n",
+    "```\n",
+    "    \n",
+    "</div>\n",
```
jwmueller, line 71: seems like a lot of newline chars, but I assume they are necessary for good formatting. Just noting in case they're not necessary

docs/source/tutorials/text.ipynb
```diff
@@ -24,6 +24,54 @@
...
+    "cl = CleanLearning(model)\n",
+    "_ = cl.fit(data, labels)\n",
+    "label_issues = cl.get_label_issues()\n",
+    "    \n",
+    "\n",
+    "```\n",
+    "    \n",
+    "</div>\n",
+    "    \n",
+    "Do you already have true `labels` and `pred_probs` from a trained model trained on text data? Run the code below to get ranked label issue indices.\n",
```
jwmueller, line 53: Let's update this statement to read:

Is your model/data not compatible with CleanLearning?  You can instead run cross-validation on your model to get out-of-sample `pred_probs` (may already have these from a previously trained model).  Run the code below to get indices of examples with label issues ranked by their inferred severity.

jwmueller:
> Some tutorials use `y` some `labels`, should we make this consistent?

Yes it'd be great if you can go through all tutorials and consistently use the variable name `labels` in place of `y` in this PR.  I think `labels` is the clearest name we could use throughout.

docs/source/tutorials/text.ipynb
ulya-tkch: Super necessary 😅

docs/source/tutorials/audio.ipynb
ulya-tkch: Some (like numpy) are missing, I am adding them with the most recent version I ran?

jwmueller: Are you talking about them being missing from docs/requirements.txt?

docs/requirements.txt is only supposed to list *extra* dependencies that are not already cleanlab dependencies. 
You probably should not be listing versions of cleanlab dependencies here (just like we decided not to list the version of cleanlab).  Cleanlab dependencies are automatically installed with cleanlab, we shouldn't encourage user to separately install other versions of them.

docs/requirements.txt
```diff
@@ -25,3 +25,7 @@ huggingface_hub==0.7
 torchaudio==0.11.0
 fasttext==0.9.2
 timm==0.6.5
+numpy==1.21.6
```
jwmueller, line 28: Let's remove numpy and any other dependencies of cleanlab as well.

This file is only supposed to list extra dependencies which aren't installed when cleanlab is installed

docs/source/tutorials/text.ipynb
```diff
@@ -56,7 +106,7 @@
    "source": [
     "# Package installation (hidden on docs website).\n",
     "# If running on Colab, may want to use GPU (select: Runtime > Change runtime type > Hardware accelerator > GPU)\n",
-    "\n",
+    "# Package versions we used: tensorflow==2.8.1 keras==2.8.0 numpy==1.21.6 pandas==1.3.2 scikeras==0.6.1\n",
     "dependencies = [\"cleanlab\", \"sklearn\", \"tensorflow\", \"tensorflow_datasets\", \"scikeras\"]\n",
```
jwmueller, line 110: sorry not an issue introduced by your PR, but could we also get rid of the sklearn in the `dependencies` list from all the tutorials?

sklearn is already a dependency of cleanlab, so we should actually not be telling users to separately install it.

docs/source/tutorials/image.ipynb
```diff
@@ -22,6 +22,56 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert alert-info\">\n",
+    "Quickstart\n",
+    "<br/>\n",
+    "    \n",
+    "Already have an sklearn compatible `model`, image `data` and true `labels`? Run the code below to train your `model` and get label issues.\n",
```
jwmueller, line 33: "and true" -> "with given"

jwmueller: We should make this ^ change for all notebooks, sorry I didn't catch it before!  Re terminology, we should be extremely wary about ever calling anything the "true label"

docs/source/tutorials/audio.ipynb
```diff
@@ -19,6 +19,56 @@
...
+    "label_issues = cl.get_label_issues()\n",
+    "preds = cl.predict(test_data) # predictions from a version of your model \n",
+    "                              # trained on auto-cleaned data\n",
+    "\n",
+    "\n",
+    "```\n",
+    "    \n",
+    "</div>\n",
+    "    \n",
+    "Is your model/data not compatible with `CleanLearning`? You can instead run cross-validation on your model to get out-of-sample `pred_probs` on true `labels`. Then run the code below to get label issue indices ranked by their inferred severity.\n",
```
jwmueller, line 50: Let's delete this and only demonstrate `filter.find_label_issues()` in the quickstart.  This tutorial never mentions CleanLearning, and I think it's confusing to mention CleanLearning in this blob when it is never illustrated anywhere in the tutorial.  In general, these quickstart blobs should only cover functionality later illustrated in the tutorial.

docs/source/tutorials/dataset_health.ipynb
```diff
@@ -15,6 +15,32 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert alert-info\">\n",
+    "Quickstart\n",
+    "<br/>\n",
+    "    \n",
+    "Already have true `labels` and `pred_probs` from a model trained on your dataset? Run the code below to evaluate dataset health.\n",
```
jwmueller, line 26: See earlier comment, we almost never want to use phrase "true labels"

jwmueller: "dataset health" -> "the overall health of your dataset"

docs/source/tutorials/image.ipynb
```diff
@@ -22,6 +22,56 @@
...
+    "    \n",
+    "```python\n",
+    "\n",
+    "from cleanlab.classification import CleanLearning\n",
+    "\n",
+    "cl = CleanLearning(model)\n",
+    "_ = cl.fit(train_data, labels)\n",
+    "label_issues = cl.get_label_issues()\n",
+    "preds = cl.predict(test_data) # predictions from a version of your model \n",
+    "                              # trained on auto-cleaned data\n",
```
jwmueller, line 46: Let's delete all the CleanLearning stuff from this quickstart blob. This tutorial never shows how to do anything with CleanLearning, so it is confusing here

docs/source/tutorials/audio.ipynb
ulya-tkch: Oops ok I'll amend!

docs/source/tutorials/text.ipynb
ulya-tkch: Getting rid of sklearn, pandas, numpy and tqdm from here (which are all cleanlab dependencies in setup.py)

docs/requirements.txt
```diff
@@ -25,3 +22,6 @@ huggingface_hub==0.7
 torchaudio==0.11.0
 fasttext==0.9.2
 timm==0.6.5
+torch==1.11.0
+requests==2.28.0
+pathlib==1.0.1
```
jwmueller, line 27: I think pathlib is already included in python so this isn't needed?  https://pathlib.readthedocs.io/en/pep428/

docs/source/tutorials/text.ipynb
```diff
@@ -56,7 +106,7 @@
    "source": [
     "# Package installation (hidden on docs website).\n",
     "# If running on Colab, may want to use GPU (select: Runtime > Change runtime type > Hardware accelerator > GPU)\n",
-    "\n",
+    "# Package versions we used: tensorflow==2.8.1 keras==2.8.0 scikeras==0.6.1\n",
```
jwmueller, line 109: don't think keras needs to be declared anymore nowadays, it's just part of tensorflow

jwmueller: I've removed it

docs/source/tutorials/text.ipynb
```diff
@@ -56,7 +106,7 @@
    "source": [
     "# Package installation (hidden on docs website).\n",
     "# If running on Colab, may want to use GPU (select: Runtime > Change runtime type > Hardware accelerator > GPU)\n",
-    "\n",
+    "# Package versions we used: tensorflow==2.8.1 scikeras==0.6.1\n",
```
jwmueller, line 109: tensorflow_datasets datasets version used?

docs/source/tutorials/tabular.ipynb
```diff
@@ -55,6 +105,8 @@
    "outputs": [],
    "source": [
     "# Package installation (hidden on docs website).\n",
+    "# Package versions used: no extra packages\n",
+    "\n",
```
jwmueller, line 109: Let's delete for brevity.

jwmueller: I've deleted it

docs/requirements.txt
jwmueller: I've removed it

docs/source/tutorials/audio.ipynb
```diff
@@ -54,6 +84,7 @@
    "outputs": [],
    "source": [
     "# Package installation (hidden on docs website).\n",
+    "# Package versions used: tensorflow==2.8.1 torch==1.11.0 pathlib==1.0.1 torchaudio==0.11.0 speechbrain==0.5.11\n",
```
jwmueller, line 87: pathlib is not necessary to specify

jwmueller: I've removed it

docs/source/tutorials/image.ipynb
```diff
@@ -122,7 +153,7 @@
     "X /= 255.0  # Scale the features to the [0, 1] range\n",
     "X = X.reshape(len(X), 1, 28, 28)  # reshape into [N, C, H, W] for PyTorch\n",
     "\n",
-    "y = mnist.target.astype(\"int64\").to_numpy()  # 1D array of labels"
+    "labels = mnist.target.astype(\"int64\").to_numpy()  # 1D array of true labels"
```
jwmueller, line 156: should not use phrase "true labels" except in special circumstances

jwmueller: I've replaced with "given labels"

docs/source/tutorials/outliers.ipynb
```diff
@@ -534,9 +534,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "cleanlab-p38",
    "language": "python",
-   "name": "python3"
+   "name": "cleanlab-p38"
```
jwmueller, line 539: IMPORTANT: for the docs to build properly in CI, notebooks must have kernelspec that looks like this:
```
"kernelspec": {
   "display_name": "Python 3 (ipykernel)",
   "language": "python",
   "name": "python3"
...
```

instead of referencing arbitrary virtualenvs

jwmueller: I've fixed this

jwmueller (approved):
LGTM!  Added some more minor edits in various places

cgnorthcutt:
@ulya-tkch 
Note in the literature there are two types of labels 'true labels' (which are never observed) and 'given label's (which are observed). The two names are meaningfully different. They should never be used interchangeably for this reason. 

In Cleanlab Studio, we use the name 'given label' for the (potentially noisy) given label and 'corrected label' for our best guess (with user input) at the 'true label'.

PR #329, opened 2022-07-29
Re-order find_label_issues args for better clarity 

jwmueller:
Most important arg users keep overlooking is the `return_indices_ranked_by` so moved it to be the first keyword-arg.  Roughly ordered the other args based on the fraction of users I guesstimate ever specify (or should need to specify) them.

Also improves clarity of some docstrings in this function.

### DIFF
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -58,10 +58,10 @@ def find_label_issues(
     return_indices_ranked_by=None,
     rank_by_kwargs={},
     filter_by="prune_by_noise_rate",
+    multi_label=False,
     frac_noise=1.0,
     num_to_remove_per_class=None,
     min_examples_per_class=1,
-    multi_label=False,
     confident_joint=None,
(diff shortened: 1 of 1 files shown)


cgnorthcutt (changes requested):
Added one clarifying commit.

tl;dr - i think `multi_label` should stay at the top until we fix multi_label. Rest LGTM.

I am good with everything except moving `multi_label` to the end. I think that should stay at the top (just after the asterisk) because this is one of the most common issues I get emailed/see posted/users have is figuring out how to run this for multi-label. We can change this later if/when we change how multi-label works, but for now I think multi-label should stay at the top because many users don't realize they need to set this to true.

cgnorthcutt:
went ahead and moved multi_label up. but kept ranked_by and filter_by args as top-most. i agree these are most important. thanks jonas!

cgnorthcutt (approved):
LGTM now. thanks!

PR #333, opened 2022-08-08
CL functionality for multiannotator data

huiwengoh:
Adding functionality for cleanlab to support tasks with multiple annotators.

### DIFF
--- a/cleanlab/internal/label_quality_utils.py
+++ b/cleanlab/internal/label_quality_utils.py
@@ -104,24 +104,3 @@ def get_normalized_entropy(pred_probs: np.ndarray, min_allowed_prob: float = 1e-
     # Note that dividing by log(num_classes) changes the base of the log which rescales entropy to 0-1 range
     clipped_pred_probs = np.clip(pred_probs, a_min=min_allowed_prob, a_max=None)
     return -np.sum(pred_probs * np.log(clipped_pred_probs), axis=1) / np.log(num_classes)
-
-
-def _get_label_quality_scores_with_NA(
-    labels: pd.Series,
-    pred_probs: np.ndarray,
-    kwargs: dict = {},
-) -> pd.Series:
-    """Returns label quality scores for each datapoint.
-    Same functionality as ``get_label_quality_scores`` with additional support for
-    datasets with NaN values.
-    For more info about parameters and returns, see the docstring of :py:func:`get_label_quality_scores <cleanlab.rank.get_label_quality_scores>`.
-    """
-    mask = pd.notna(labels)
-    label_quality_scores_subset = rank.get_label_quality_scores(
-        labels=labels[mask].astype("int64"),
-        pred_probs=pred_probs[mask],
-        **kwargs,
-    )
-    label_quality_scores = pd.Series(np.nan, index=np.arange(len(labels)))
-    label_quality_scores[mask] = label_quality_scores_subset
-    return label_quality_scores
--- a/cleanlab/internal/label_quality_utils.py
+++ b/cleanlab/internal/label_quality_utils.py
@@ -17,6 +17,8 @@
 """Helper functions for computing label quality scores"""
 
 import numpy as np
+import pandas as pd
+from cleanlab import rank
 from cleanlab.count import get_confident_thresholds
 
 
@@ -102,3 +104,23 @@ def get_normalized_entropy(pred_probs: np.ndarray, min_allowed_prob: float = 1e-
     # Note that dividing by log(num_classes) changes the base of the log which rescales entropy to 0-1 range
     clipped_pred_probs = np.clip(pred_probs, a_min=min_allowed_prob, a_max=None)
     return -np.sum(pred_probs * np.log(clipped_pred_probs), axis=1) / np.log(num_classes)
+
+
+def _get_label_quality_scores_with_NA(
+    labels: pd.Series,
+    pred_probs: np.array,
+    kwargs: dict = {},
+) -> pd.Series:
+    """Returns label quality scores for each datapoint.
+    Same functionality as ``get_label_quality_scores`` with additional support for
+    datasets with NaN values.
+    For more info about parameters and returns, see the docstring of :py:func:`get_label_quality_scores <cleanlab.rank.get_label_quality_scores>`.
+    """
+    label_quality_scores_subset = rank.get_label_quality_scores(
+        labels=labels[pd.notna(labels)].astype("int32"),
+        pred_probs=pred_probs[pd.notna(labels)],
+        **kwargs,
+    )
+    label_quality_scores = pd.Series(np.nan, index=np.arange(len(labels)))
+    label_quality_scores[pd.notna(labels)] = label_quality_scores_subset
+    return label_quality_scores
--- a/cleanlab/internal/label_quality_utils.py
+++ b/cleanlab/internal/label_quality_utils.py
@@ -17,8 +17,6 @@
 """Helper functions for computing label quality scores"""
 
 import numpy as np
-import pandas as pd
-from cleanlab import rank
 from cleanlab.count import get_confident_thresholds
 
 
--- a/cleanlab/internal/label_quality_utils.py
+++ b/cleanlab/internal/label_quality_utils.py
@@ -116,11 +116,12 @@ def _get_label_quality_scores_with_NA(
     datasets with NaN values.
     For more info about parameters and returns, see the docstring of :py:func:`get_label_quality_scores <cleanlab.rank.get_label_quality_scores>`.
     """
+    mask = pd.notna(labels)
     label_quality_scores_subset = rank.get_label_quality_scores(
-        labels=labels[pd.notna(labels)].astype("int32"),
-        pred_probs=pred_probs[pd.notna(labels)],
+        labels=labels[mask].astype("int64"),
+        pred_probs=pred_probs[mask],
         **kwargs,
     )
     label_quality_scores = pd.Series(np.nan, index=np.arange(len(labels)))
-    label_quality_scores[pd.notna(labels)] = label_quality_scores_subset
+    label_quality_scores[mask] = label_quality_scores_subset
     return label_quality_scores
--- a/cleanlab/internal/label_quality_utils.py
+++ b/cleanlab/internal/label_quality_utils.py
@@ -108,7 +108,7 @@ def get_normalized_entropy(pred_probs: np.ndarray, min_allowed_prob: float = 1e-
 
 def _get_label_quality_scores_with_NA(
     labels: pd.Series,
-    pred_probs: np.array,
+    pred_probs: np.ndarray,
     kwargs: dict = {},
 ) -> pd.Series:
     """Returns label quality scores for each datapoint.
--- a/cleanlab/internal/label_quality_utils.py
+++ b/cleanlab/internal/label_quality_utils.py
@@ -35,10 +35,10 @@ def _subtract_confident_thresholds(
 
     Parameters
     ----------
-    labels : np.array
+    labels : np.ndarray
       Labels in the same format expected by the `cleanlab.count.get_confident_thresholds()` method.
 
-    pred_probs : np.array (shape (N, K))
+    pred_probs : np.ndarray (shape (N, K))
       Predicted-probabilities in the same format expected by the `cleanlab.count.get_confident_thresholds()` method.
 
     multi_label : bool, optional
@@ -53,7 +53,7 @@ def _subtract_confident_thresholds(
 
     Returns
     -------
-    pred_probs_adj : np.array (float)
+    pred_probs_adj : np.ndarray (float)
       Adjusted pred_probs.
     """
 
@@ -86,7 +86,7 @@ def get_normalized_entropy(pred_probs: np.ndarray, min_allowed_prob: float = 1e-
 
     Parameters
     ----------
-    pred_probs : np.array (shape (N, K))
+    pred_probs : np.ndarray (shape (N, K))
       Each row of this matrix corresponds to an example x and contains the model-predicted
       probabilities that x belongs to each possible class: P(label=k|x)
 
@@ -95,7 +95,7 @@ def get_normalized_entropy(pred_probs: np.ndarray, min_allowed_prob: float = 1e-
       Ensures entropy remains well-behaved even when `pred_probs` contains zeros.
     Returns
     -------
-    entropy : np.array (float)
+    entropy : np.ndarray (float)
       Each element is the normalized entropy of the corresponding row of ``pred_probs``.
     """
 
--- a/cleanlab/multiannotator.py
+++ b/cleanlab/multiannotator.py
@@ -183,7 +183,25 @@ def _get_annotator_agreement(
 def _get_annotator_agreement_with_annotators(
     labels_multiannotator: pd.DataFrame,
     num_annotations: np.ndarray,
-):
+) -> np.ndarray:
+    """Returns the average agreement of each annotators with other annotatos that label the same examples.
+
+    Parameters
+    ----------
+    labels_multiannotator : pd.DataFrame
+        2D pandas DataFrame of multiple given labels for each example with shape (N, M),
+        where N is the number of examples and M is the number of annotators.
+        For more details, labels in the same format expected by the :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`.
+    consensus_label : np.ndarray
+        An array of shape ``(N,)`` with the consensus labels aggregated from all annotators.
+
+    Returns
+    -------
+    annotator_agreement : np.ndarray
+        An array of shape ``(M,)`` where M is the number of annotators, with the agreement of each annotator with other
+        annotators that labeled the same examples.
+    """
+
     def get_single_annotator_agreement(
         labels_multiannotator: pd.DataFrame,
         num_annotations: np.ndarray,
@@ -210,18 +228,45 @@ def _get_annotator_agreement_with_annotators(
 def _get_post_pred_probs(
     labels_multiannotator: pd.DataFrame,
     consensus_label: np.ndarray,
-    pred_probs: np.ndarray,
+    prior_pred_probs: np.ndarray,
     num_annotations: np.ndarray,
     annotator_agreement: np.ndarray,
     quality_method: str,
-):
+) -> np.ndarray:
+    """Return the posterior predicted probabilites of each example given a specified quality method.
+
+    Parameters
+    ----------
+    labels_multiannotator : pd.DataFrame
+        2D pandas DataFrame of multiple given labels for each example with shape (N, M),
+        where N is the number of examples and M is the number of annotators.
+        For more details, labels in the same format expected by the :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`.
+    consensus_label : np.ndarray
+        An array of shape ``(N,)`` with the consensus labels aggregated from all annotators.
+    prior_pred_probs : np.ndarray
+        TODO: more detailed explanation that this is the model pred_probs
+        An array of shape ``(N, K)`` of posterior predicted probabilities, ``P(label=k|x)``.
+        For details, predicted probabilities in the same format expected by the :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`.
+    num_annotations : np.ndarray
+        An array of shape ``(N,)`` with the number of annotators that have labeled each example.
+    annotator_agreement : np.ndarray
+        An array of shape ``(N,)`` with the fraction of annotators that agree with each consensus label.
+    quality_method : str
+        Specifies the method used to calculate the quality of the consensus label.
+        For valid quality methods, view :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`
+
+    Returns
+    -------
+    post_pred_probs : np.ndarray
+        An array of shape ``(N, K)`` with the posterior predicted probabilities.
+    """
     valid_methods = [
         "auto",
         "agreement",
     ]
 
     if quality_method == "auto":
-        num_classes = get_num_classes(pred_probs=pred_probs)
+        num_classes = get_num_classes(pred_probs=prior_pred_probs)
         likelihood = np.mean(
             annotator_agreement[num_annotations != 1]
         )  # likelihood that any annotator will annotate the consensus label for any example
@@ -241,20 +286,20 @@ def _get_post_pred_probs(
         # compute model weight
         mask = num_annotations != 1
         consensus_label_subset = consensus_label[mask]
-        pred_probs_subset = pred_probs[mask]
+        prior_pred_probs_subset = prior_pred_probs[mask]
 
-        model_error = np.mean(np.argmax(pred_probs_subset, axis=1) != consensus_label_subset)
+        model_error = np.mean(np.argmax(prior_pred_probs_subset, axis=1) != consensus_label_subset)
         model_weight = (1 - (model_error / most_likely_class_error)) * np.sqrt(
             np.mean(num_annotations)
         )
 
         # compute weighted average
-        post_pred_probs = np.full(pred_probs.shape, np.nan)
+        post_pred_probs = np.full(prior_pred_probs.shape, np.nan)
         for i in range(len(labels_multiannotator)):
             example = labels_multiannotator.iloc[i].dropna()
             post_pred_probs[i] = [
                 np.average(
-                    [pred_probs[i, true_label]]
+                    [prior_pred_probs[i, true_label]]
                     + [
                         likelihood
                         if annotator_label == true_label
@@ -397,7 +442,7 @@ def _get_consensus_stats(
     post_pred_probs = _get_post_pred_probs(
         labels_multiannotator=labels_multiannotator,
         consensus_label=consensus_label,
-        pred_probs=pred_probs,
+        prior_pred_probs=pred_probs,
         num_annotations=num_annotations,
         annotator_agreement=annotator_agreement,
         quality_method=quality_method,
--- a/cleanlab/multiannotator.py
+++ b/cleanlab/multiannotator.py
@@ -888,12 +888,12 @@ def get_label_quality_multiannotator(
             "detailed_label_quality": detailed_label_quality,
             "annotator_stats": annotator_stats,
         }
-    if return_detailed_quality:
+    elif return_detailed_quality:
         return {
             "label_quality_multiannotator": label_quality_multiannotator,
             "detailed_label_quality": detailed_label_quality,
         }
-    if return_annotator_stats:
+    elif return_annotator_stats:
         return {
             "label_quality_multiannotator": label_quality_multiannotator,
             "annotator_stats": annotator_stats,
--- a/cleanlab/multiannotator.py
+++ b/cleanlab/multiannotator.py
@@ -4,7 +4,6 @@ from typing import List, Union, Tuple
 from cleanlab.rank import get_label_quality_scores
 from cleanlab.internal.util import get_num_classes
 from cleanlab.internal.label_quality_utils import _get_label_quality_scores_with_NA
-from cleanlab.dataset import _get_worst_class
 import warnings
 
 
@@ -622,15 +621,6 @@ def _get_annotator_stats(
         For details, see the documentation of :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`.
     """
 
-    def try_get_worst_class(labels, pred_probs):
-        try:
-            return _get_worst_class(labels, pred_probs)
-        except:
-            warnings.warn(
-                "worst_class labels for some annotators are NaN due to missing class labels"
-            )
-            return np.NaN
-
     annotator_quality = _get_annotator_quality(
         labels_multiannotator=labels_multiannotator,
         pred_probs=pred_probs,
@@ -654,9 +644,10 @@ def _get_annotator_stats(
     ).to_numpy()
 
     # Find the worst labeled class for each annotator
+    # TODO: potentially have to do tiebreaking here
     worst_class = labels_multiannotator.apply(
-        lambda s: try_get_worst_class(s[pd.notna(s)], pred_probs[pd.notna(s)])
-    )
+        lambda s: (s[pd.notna(s)] == consensus_label[pd.notna(s)]).groupby(s).mean().idxmin()
+    ).astype("int64")
 
     # Create multi-annotator stats DataFrame from its columns
     annotator_stats = pd.DataFrame(
--- a/cleanlab/multiannotator.py
+++ b/cleanlab/multiannotator.py
@@ -11,7 +11,7 @@ import warnings
 def convert_long_to_wide_dataset(
     labels_multiannotator_long: pd.DataFrame,
 ) -> pd.DataFrame:
-    """Converts a wide format dataset to long format which is suitable for passing into
+    """Converts a long format dataset to wide format which is suitable for passing into
     :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`.
     Dataframe must contain three columns named:
     1. ``task`` representing each example labeled by the annotators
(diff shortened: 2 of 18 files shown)


cleanlab/multiannotator.py
```diff
@@ -0,0 +1,829 @@
...
+from cleanlab.internal.util import get_num_classes
+from cleanlab.internal.label_quality_utils import _get_label_quality_scores_with_NA
+from cleanlab.dataset import _get_worst_class
+import warnings
+
+
+def convert_long_to_wide_dataset(
+    labels_multiannotator_long: pd.DataFrame,
+) -> pd.DataFrame:
+    """Converts a wide format dataset to long format which is suitable for passing into
```
jwmueller, line 14: I think you have it reversed in the description :) should be long to wide

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,829 @@
...
+    Dataframe must contain three columns named:
+    1. ``task`` representing each example labeled by the annotators
+    2. ``annotator`` representing each annotator
+    3. ``label`` representing the label given by an annotator for the corresponding task
+
+    Parameters
+    ----------
+    labels_multiannotator_long : pd.DataFrame
+        pandas DataFrame in long format with three columns named ``task``, ``annotator`` and ``label``
+    """
```
jwmueller, line 25: could clarify this:

Returns
--------

labels_multiannotator_wide : pd.DataFrame
   Object of the proper format to be passed as \`\`labels_multiannotator\`\` for the other \`\`cleanlab.multiannotator\`\` functions.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,829 @@
...
+    labels_multiannotator.columns.name = None
+    return labels_multiannotator
+
+
+def get_consensus_label(
+    labels_multiannotator: pd.DataFrame or np.ndarray,
+    pred_probs: np.ndarray = None,
+    consensus_method: str = "majority",
+) -> np.ndarray:
+    """Returns consensus labels aggregated from multiple annotators, determined by method specified.
```
jwmueller, line 39: could rephrase as:

Returns a single consensus label for each example, aggregated from the labels given by multiple annotators.

jwmueller: The discussion of which method is used for the aggregation should IMO be reserved for the proper parameters below. We don't want non-advanced users to already be thinking about methods/choices they have to make. We prefer the package seems easy-to-use and trusted to already be using good default settings.

cleanlab/multiannotator.py
huiwengoh: Oops good catch!

cleanlab/multiannotator.py
huiwengoh: Left the returns section out by accident, good catch!

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,829 @@
...
+
+    Parameters
+    ----------
+    labels_multiannotator : pd.DataFrame or np.ndarray
+        2D pandas DataFrame of multiple given labels for each example with shape (N, M),
```
ulya-tkch, line 44: np.ndarray can be passed but only pd.DataFrame is mentioned in the description

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,829 @@
...
+        where N is the number of examples and M is the number of annotators.
+        For more details, labels in the same format expected by the :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`.
+    pred_probs : np.ndarray, optional
+        An array of shape ``(N, K)`` of model-predicted probabilities, ``P(label=k|x)``.
+        For details, predicted probabilities in the same format expected by the :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`.
```
ulya-tkch, line 49: This suggests taking a look at `get_label_quality_multiannotator` for details but `get_label_quality_multiannotator` suggests taking a look at `get_label_quality_scores`. Maybe have one function contain the final docstring and link to itt?

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,859 @@
...
+        )
+
+    return consensus_label.astype("int64")
+
+
+def _get_annotator_agreement(
+    labels_multiannotator: pd.DataFrame,
+    consensus_label: np.ndarray,
+) -> np.ndarray:
+    """Returns the fractions of annotators that agree with the consensus labels. Note that the
```
ulya-tkch, line 164: [nit] Returns the fraction of annotators that agree with the consensus label per example.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,859 @@
...
+        axis=1,
+    )
+    return annotator_agreement.to_numpy()
+
+
+def _get_annotator_agreement_with_annotators(
+    labels_multiannotator: pd.DataFrame,
+    num_annotations: np.ndarray,
+) -> np.ndarray:
+    """Returns the average agreement of each annotators with other annotatos that label the same examples.
```
ulya-tkch, line 192: Returns the average agreement of each annotator with other annotators that label the same example.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,859 @@
...
+        - ``agreement_with_consensus``: fraction of examples where a given annotator agrees with the consensus label
+        - ``worst_class``: the class that is most frequently mislabeled by a given annotator
+    """
+
+    def try_get_worst_class(labels, pred_probs):
+        try:
+            return _get_worst_class(labels, pred_probs)
+        except:
+            warnings.warn(
+                "worst_class labels for some annotators are NaN due to missing class labels"
```
ulya-tkch, line 621: _get_worst_class function docstring says NaN is returned "if all classes do not have issues" but warning suggests "worst_class labels for some annotators are NaN due to missing class labels". Confusing in documentation which is the real reason for NaN

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,859 @@
...
+    pred_probs : np.ndarray
+        An array of shape ``(N, K)`` of model-predicted probabilities, ``P(label=k|x)``.
+        For details, predicted probabilities in the same format expected by the :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`.
+    consensus_label : np.ndarray
+        An array of shape ``(N,)`` with the consensus labels aggregated from all annotators.
+    num_annotations : np.ndarray
+        An array of shape ``(N,)`` with the number of annotators that have labeled each example.
+    annotator_agreement : np.ndarray
+        An array of shape ``(N,)`` with the fraction of annotators that agree with each consensus label.
+    quality_method : str
```
ulya-tkch, line 602: In other parts of codebase they specify if there is a default method in docstring 
i.e. quality_method : str, default = "auto"

cleanlab/multiannotator.py
huiwengoh: Ah missed that out! Added np array in description

cleanlab/multiannotator.py
huiwengoh: (bad spelling/grammar ...) 
Thanks for catching! Edits made

tests/test_multiannotator.py
```diff
@@ -0,0 +1,187 @@
...
+        ).transpose()
+    )
+
+    # column of labels without NaNs to test _get_worst_class
+    complete_labels = deepcopy(s)
+
+    # Each annotator only labels approximately 20% of the dataset
+    # (unlabeled points represented with NaN)
+    s = s.apply(lambda x: x.mask(np.random.random(n) < 0.8))
+    s.dropna(axis=1, how="all", inplace=True)
```
jwmueller, line 72: maybe need to add check that we didn't drop too many annotations?  Eg. the random seed may sometimes have different effects on different operating systems, so would be ideal to ensure this code will 100% produce a valid dataset regardless of the seed.

tests/test_multiannotator.py
```diff
@@ -0,0 +1,187 @@
...
+# Global to be used by all test methods. Only compute this once for speed.
+data = make_data()
+
+
+def test_convert_long_to_wide():
+    labels_long = make_data_long(data["labels"])
+    labels_wide = convert_long_to_wide_dataset(labels_long)
+
+    assert isinstance(labels_wide, pd.DataFrame)
+    # TODO: might add a test to see if the wide dataframe actually contains the right info
```
jwmueller, line 111: how about just checking one index (say 10) of labels_wide contains the proper values from the corresponding index of labels_long?

tests/test_multiannotator.py
```diff
@@ -0,0 +1,187 @@
...
+    # Assert that the worst class index should be the class with the highest noise
+    assert dataset._get_worst_class(labels, pred_probs) == data["noise_matrix"].diagonal().argmax()
+
+
+def test_label_quality_scores_multiannotator():
+    labels = data["labels"]
+    pred_probs = data["pred_probs"]
+
+    lqs_multiannotator = get_label_quality_multiannotator(labels, pred_probs)
+    assert isinstance(lqs_multiannotator, pd.DataFrame)
```
jwmueller, line 128: maybe check shape matches what we expect as well?  And (some of the) column names?

tests/test_multiannotator.py
```diff
@@ -0,0 +1,187 @@
...
+        np.array(labels), pred_probs, quality_method="agreement"
+    )
+
+    # test returning annotator stats
+    lqs_annotatorstats = get_label_quality_multiannotator(
+        labels, pred_probs, return_annotator_stats=True
+    )
+    assert isinstance(lqs_annotatorstats, tuple)
+    assert isinstance(lqs_annotatorstats[0], pd.DataFrame)
+    assert isinstance(lqs_annotatorstats[1], pd.DataFrame)
```
jwmueller, line 152: maybe check shape matches what we expect as well? And (some of the) column names?

jwmueller: maybe also check every annotator is present in: `lqs_annotatorstats[1]`

jwmueller: For both DFs, how about checking the values in each column (or row) lie within some bounds that we know they should satisfy, eg. nonnegative, between 0-1, etc.

tests/test_multiannotator.py
```diff
@@ -0,0 +1,187 @@
...
+            [0.4, 0.4, 0.2],
+            [0.3, 0.6, 0.1],
+            [0.75, 0.2, 0.05],
+            [0.1, 0.4, 0.5],
+            [0.2, 0.4, 0.4],
+            [0.2, 0.4, 0.4],
+        ]
+    )
+
+    consensus_label = get_consensus_label(labels_tiebreaks, pred_probs_tiebreaks)
```
jwmueller, line 187: Great test!  Can we add 1-2 more edge-case tests like this?  In particular for sparse setting where most examples only have 1 annotation, there are some annotators that don't overlap with others, etc.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,962 @@
...
+
+
+def _get_quality_of_consensus(
+    labels_multiannotator: pd.DataFrame,
+    consensus_label: np.ndarray,
+    pred_probs: np.ndarray,
+    num_annotations: np.ndarray,
+    annotator_agreement: np.ndarray,
+    quality_method: str = "auto",
+    kwargs: dict = {},
```
jwmueller, line 374: this should be named `XXX_kwargs` where XXX provides extra info about how these kwargs will be used.  This avoids confusion with the standard `**kwargs`

cleanlab/internal/label_quality_utils.py
```diff
@@ -17,6 +17,8 @@
 """Helper functions for computing label quality scores"""
 
 import numpy as np
+import pandas as pd
+from cleanlab import rank
```
jwmueller, line 21: I don't think these two imports are being used anywhere in this file?

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,928 @@
...
+            "annotator_quality": annotator_quality,
+            "num_labeled": num_labeled,
+            "agreement_with_consensus": agreement_with_consensus,
+            "worst_class": worst_class,
+        }
+    )
+
+    return annotator_stats.sort_values(by=["annotator_quality", "agreement_with_consensus"])
+
+
```
jwmueller, line 678: Let's move `get_label_quality_multiannotator` to the top of this file since it's the most important function.

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,928 @@
...
+    It computes quality scores for each annotator's labels and other useful values like how
+    confident we are that the current consensus label is actually correct.
+    The score is between 0 and 1; lower scores
+    indicate labels less likely to be correct. For example:
+    1 - clean label (the given label is likely correct).
+    0 - dirty label (the given label is unlikely correct).
+
+    Parameters
+    ----------
+    labels_multiannotator : pd.DataFrame
```
jwmueller, line 702: pd.DataFrame or np.ndarray

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,928 @@
...
+    pred_probs: np.ndarray,
+    *,
+    consensus_method: Union[str, List[str]] = "best_quality",
+    quality_method: str = "auto",
+    return_detailed_quality: bool = False,
+    return_annotator_stats: bool = False,  # sort by lowest overall_quality first
+    verbose: bool = True,
+    label_quality_score_kwargs: dict = {},
+) -> Union[pd.DataFrame, Dict[str, pd.DataFrame]]:
+    """Returns label quality scores for each example for each annotator.
```
jwmueller, line 690: [nit] "for each example for each annotator" -> "for each example and for each annotator"

cleanlab/multiannotator.py
```diff
@@ -0,0 +1,928 @@
...
+    This function is for multiclass classification datasets where examples have been labeled by
+    multiple annotators (not necessarily the same number of annotators per example).
+    It computes quality scores for each annotator's labels and other useful values like how
+    confident we are that the current consensus label is actually correct.
```
jwmueller, line 694: "like how confident we are that the current consensus label is actually correct"
->
"like one consensus label for each example that best accounts for the labels chosen by each annotator (and their quality), as well as a score for how confident we are that this consensus label is actually correct"

cleanlab/internal/label_quality_utils.py
huiwengoh: My bad, I removed a function from this file and forgot to get remove the imports. Deleted!

cleanlab/multiannotator.py
huiwengoh: `get_worst_class` function have been edited, above warnings are irrelevant now

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+   "metadata": {},
+   "source": [
+    "## 2. Create the data (can skip these details)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "69b5ddaa",
+   "metadata": {},
+   "source": [
```
jwmueller, line 117: Note the code in your markdown cell Pulldown  does not exactly match the code in the hidden python cell below.  In particular the hidden python cell has some extra code:


```
X = data_dict["X_train"]
multiannotator_labels = data_dict["multiannotator_labels"]
true_labels = data_dict["true_labels_train"]
```

jwmueller: I'd prefer all that code and the execution of:
```
data_dict = make_data()
```
to be an entirely separate python cell (not hidden).  So the only thing that should be in this markdown pulldown cell and in the hidden cell would be the definition of the function, not the usage of the function to define global variables used elsewhere in the notebook.

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e95d31b2",
+   "metadata": {},
+   "source": [
+    "For this tutorial we are using a toy dataset that has 50 annotators and 160 examples. There are three possible classes, `0`, `1` and `2`.\n",
+    "\n",
+    "Each annotator annotates approximately 10% of the examples. We also synthetically made the last 5 annotators in our toy dataset have much noisier labels than the rest of the annotators."
```
jwmueller, line 281: Add final cell here that prints a `head()` of the data that the user would have in practice (not the "true"-stuff).  We want to give reader a sense of what their data should look like. 

If we've designed our APIs right, the only work  the user needs to do is get their data formatted properly, so it's imperative we make it as clear as possible what the exact format should look like

jwmueller: When I run this code, the `multiannotator_labels` object contains floats as the labels, but I'd presume it would make more sense for these to be ints?

jwmueller: Basically you'll want to envision users trying to run our functions on every single malformatted dataset that could possibly exist :)  

And we want to preclude these issues from happening via a clear tutorial and docstrings!

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+   "source": [
+    "## 3. Get majority vote label and compute out-of-sample predicted probabilites"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c1857cc7",
+   "metadata": {},
+   "source": [
+    "Before training our machine learning model, we must first obtain the consensus labels from the annotators that labeled the data. The simplest way to obtain an initial set of consensus labels is to select it using majority vote."
```
jwmueller, line 312: can we add hyperlink here for: majority vote
to a stable website with clear definition/tutorial of what this is in the crowdsourcing context?

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "6ca92617",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "multiannotator_dict = get_label_quality_multiannotator(multiannotator_labels, cv_pred_probs)"
```
jwmueller, line 374: Let's think carefully about the name of this variable `multiannotator_dict` since users tend to just copy the tutorials' code, so their own code will probably have this variable in it too (and thus should have clearest name possible).

jwmueller: Alternative names could include:
`multiannotator_results` ,`multiannotator_analysis`, but I cant necessarily think of alternate names that are clearly superior to `multiannotator_dict`

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "14251ee0",
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "multiannotator_dict[\"detailed_label_quality\"].head()"
+   ]
```
jwmueller, line 423: Can we make the annotators have string IDs in the original dataset, and those IDs should be shown here as well in this DF

jwmueller: maybe let's reorder the columns more thematically so the column order is:

ID, annotator_quality, agreement_with_consensus, num_labeled, worst_class

What do you think?

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "bf945113",
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [],
+   "source": [
+    "multiannotator_dict[\"label_quality_multiannotator\"].head()"
```
jwmueller, line 402: maybe let's reorder the columns more thematically so the column order is: 
consensus_label, consensus_quality_score, annotator_agreement, num_annotations

What do you think?

jwmueller: Let's also shorten the string name of this dict key to just be "label_quality" instead of "label_quality_multiannotator" ?

I presume it's clear to user that everything in this module is related to multiannotator

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "cbd1e415",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from sklearn.linear_model import LogisticRegression\n",
+    "from sklearn.model_selection import cross_val_predict\n",
```
jwmueller, line 341: Let's move the imports to top of notebook so the user's code will fail earlier if they didn't have the proper installation for some reason

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "efe16638",
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "multiannotator_dict[\"annotator_stats\"].head(10)"
+   ]
```
jwmueller, line 443: what do you think about renaming the column "num_labeled" -> "num_examples_labeled" ?

jwmueller: maybe let's reorder the columns more thematically so the column order is:

ID, annotator_quality, agreement_with_consensus, num_labeled, worst_class

What do you think?

jwmueller: Can we make the annotators have string IDs in the original dataset, and those IDs should be shown here as well in this DF.

If the annotators have IDs, the IDs should also be the suffix of each column name in:
`multiannotator_dict["detailed_label_quality"]`

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+   "metadata": {
+    "nbsphinx": "hidden"
+   },
+   "outputs": [],
+   "source": [
+    "# Note: This cell is only for docs.cleanlab.ai, if running on local Jupyter or Colab, please ignore it.\n",
+    "\n",
+    "if majority_vote_accuracy >= cleanlab_label_accuracy:  # check cleanlab has improved prediction accuracy\n",
+    "    raise Exception(\"Cleanlab training failed to improve model accuracy.\")"
+   ]
```
jwmueller, line 526: Nice check!

jwmueller: Can we also add check that the known low-quality annotators got lower quality scores than the other annotators (eg. take the average of these 5 bad annotators' quality and compare it to avg-quality of the other annotators)

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "abd0fb0b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "improved_consensus_label = multiannotator_dict[\"label_quality_multiannotator\"][\"consensus_label\"]"
```
jwmueller, line 478: Type of `improved_consensus_label` does not appear to match `majority_vote_label`?

I think we should make them have the same type by default, probably np.ndarray

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+   "id": "b806d2ea",
+   "metadata": {
+    "nbsphinx": "hidden"
+   },
+   "outputs": [],
+   "source": [
+    "# Note: This cell is only for docs.cleanlab.ai, if running on local Jupyter or Colab, please ignore it.\n",
+    "\n",
+    "if majority_vote_accuracy >= cleanlab_label_accuracy:  # check cleanlab has improved prediction accuracy\n",
+    "    raise Exception(\"Cleanlab training failed to improve model accuracy.\")"
```
jwmueller, line 525: model accuracy -> consensus label accuracy

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+    "print(f\"Accuracy of cleanlab consensus labels = {cleanlab_label_accuracy}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2c20b2c9",
+   "metadata": {},
+   "source": [
+    "We can see that the accuracy of the consensus labels improved as a result of using cleanlab, which not only takes the annotators' labels into account, but also a model to compute better consensus labels."
+   ]
```
jwmueller, line 511: For completeness sake, final (non-hidden) cell of notebook could show retraining the logistic regression model on the improved consensus labels?

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a0d09bfa",
+   "metadata": {},
+   "source": [
+    "The `annotator_stats` DataFrame is sorted by increasing `annotator_quality`, showing us the worst annotators first.\n",
+    "\n",
+    "Notice that in the above table annotators with ids 45 to 49 have the worst annotator quality score, which is expected because we made the last 5 annotators systematically worse than the rest."
```
jwmueller, line 452: Can we add explicit check for this in the final hidden cell in this notebook?

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+    "multiannotator_labels = data_dict[\"multiannotator_labels\"]\n",
+    "true_labels = data_dict[\"true_labels_train\"]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e95d31b2",
+   "metadata": {},
+   "source": [
+    "For this tutorial we are using a toy dataset that has 50 annotators and 160 examples. There are three possible classes, `0`, `1` and `2`.\n",
```
jwmueller, line 279: What do think about us manually ensuring the last 5 examples in the dataset also have wrong consensus labels, by purposefully ensuring:
each of these examples only has one annotation, and this annotation is wrong.

This way we could better demonstrate the utility of the `consensus_quality_score`.  And the final hidden cell of this notebook could include a check that these bottom 5 examples receive low-ish consensus_quality_score

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,555 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from sklearn.linear_model import LogisticRegression\n",
+    "from sklearn.model_selection import cross_val_predict\n",
+    "\n",
+    "model = LogisticRegression()\n",
+    "\n",
+    "num_crossval_folds = 5  \n",
+    "cv_pred_probs = cross_val_predict(\n",
```
jwmueller, line 346: maybe better to just call this `pred_probs` since that's our canonical name for it throughout the codebase?

huiwengoh: That makes sense! I used `cv_pred_probs` because I saw it in some of our other tutorials ([audio](https://docs.cleanlab.ai/v2.0.0/tutorials/audio.html) and [workflow overviews](https://docs.cleanlab.ai/v2.0.0/tutorials/indepth_overview.html)). I can open a separate PR to change those

docs/source/tutorials/multiannotator.ipynb
huiwengoh: Moved the function call and subsequent lines to a non-hidden cell.

huiwengoh: I believe I also followed the format of other tutorials, so some of the ``make_data()`` calls in the existing tutorials might also be hidden. Can open a PR in the future to fix this

docs/source/tutorials/multiannotator.ipynb
huiwengoh: Moved! Again something prevalent in existing tutorials, can fix at a later date

docs/source/tutorials/multiannotator.ipynb
huiwengoh: Yes all labels are saved as dytpe float to allow for NaN values (which numpy defines as float), added note to clarify

docs/source/tutorials/multiannotator.ipynb
huiwengoh: I like `multiannotator_results`!

docs/source/tutorials/multiannotator.ipynb
huiwengoh: I originally arranged `num_annotations` to be the first column because that would be the only non-repeating column if we passed more than one `consensus_method` (in a list). But I think the order you proposed makes more sense since `consensus_label` is probably the most important column

docs/source/tutorials/multiannotator.ipynb
huiwengoh: I agree that `num_examples_labeled` would be a clearer name.
I'm thinking if we should order `num_examples_labeled` as the last column (it would be consistent with `num_annotations` being the last column in the `label_quality` df)

jwmueller: Sounds good!

huiwengoh: Are you proposing that `ID` is a column of its own in the `annotator_stats` df? 
Currently the annotator ID is used as the index (right now they have int IDs so it is not obvious), and the column names of `detailed_label_quality` is `quality_annotator_{annotatorID}`. Is that alright or should something be changed in terms of displaying IDs?
Will edit the tutorial annotator IDs to be strings instead of ints.

jwmueller: No sorry for confusion, let's just keep ID as string index when it exists!

I just wanted to make sure ID will show up somewhere if it is string

docs/source/tutorials/multiannotator.ipynb
huiwengoh: Is the purpose of this retraining to show how the user can use the improved model to predict test data and more? 
We are not trying to rerun our multiannotator functions again right (just confirming)

jwmueller: No it's just to clarify what the purpose of the improved consensus labels is. I don't think we should even report accuracy of the retrained model or anything like that

jwmueller: That said, after this extra cell you could (in text) just explain:  

We can repeatedly iterate this whole process of getting better consensus labels using the model's pred_probs and then retraining our model with the improved labels to get even better pred_probs!

huiwengoh: Not entirely sure I catch the best way to present this section, will be up in the new push and some feedback would be really helpful!

docs/source/tutorials/multiannotator.ipynb
jwmueller: Actually since we are creating the dataset and now that we want it to have annotator IDs, it has to be pandas. And I think you can in fact make pandas only have integer values with missing entries?

https://pandas.pydata.org/docs/user_guide/integer_na.html#:~:text=In%20Working%20with%20missing%20data,this%20may%20not%20matter%20much.

huiwengoh: Changed to used pd.NA types

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,753 @@
...
+    "annotator_stats = multiannotator_results[\"annotator_stats\"]\n",
+    "bad_annotator_idx = [\"A0046\", \"A0047\", \"A0048\", \"A0049\", \"A0050\"]\n",
+    "bad_annotator_mask = annotator_stats.index.isin(bad_annotator_idx)\n",
+    "\n",
+    "avg_annotator_quality_bad = np.mean(annotator_stats[bad_annotator_mask][\"annotator_quality\"])\n",
+    "avg_annotator_quality_good = np.mean(annotator_stats[~bad_annotator_mask][\"annotator_quality\"])\n",
+    "\n",
+    "if avg_annotator_quality_bad >= avg_annotator_quality_good:\n",
+    "    raise Exception(\"Low quality annotators have higher quality scores than good quality annotators\")"
+   ]
```
jwmueller, line 724: Let's also have check of the consensus-quality score down here?
Eg. comparing `worst_quality_accuracy` against 
`better_quality_accuracy` and raising Exception if they don't compare as expected.

docs/source/tutorials/multiannotator.ipynb
```diff
@@ -0,0 +1,753 @@
...
+   "source": [
+    "## 5. Retrain model using improved consensus labels"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8d4e31ab",
+   "metadata": {},
+   "source": [
+    "After obtaing the improved consensus labels, we can now retrain a machine learning model using these newly obtained labels. "
```
jwmueller, line 672: typo: obtaing

jwmueller (approved):
LGTM!  I just had a couple tiny remaining comments on the tutorial notebook.

After those are addressed and this is merged, I'll later update some of the language used in the tutorial notebook in a separate PR to better match our benchmarks/write-up.

PR #334, opened 2022-08-08
Allow missing classes in get_label_quality_scores

huiwengoh:
Updated `assert_valid_inputs` to take in optional arg `allow_missing_classes`, which will skip the check for missing classes in `labels`. `get_label_quality_scores` will no longer check input labels for missing classes.


jwmueller (approved):
LGTM!

cgnorthcutt:
quick check -- planning to remove this once missing classes are supported by default/automatically?

cgnorthcutt:
> Yes when will that be :)

Before 2.1 release!! <3

PR #335, opened 2022-08-08
Allow missing classes in assert_valid_class_labels

huiwengoh:
Allowing `assert_valid_class_labels` to take in optional arg `allow_missing_classes` to skip checks for missing classes in `labels`. 
A follow up to #334 which missed a helper function call that also checks for missing classes.

### DIFF
--- a/cleanlab/internal/validation.py
+++ b/cleanlab/internal/validation.py
@@ -101,8 +101,9 @@ def assert_valid_class_labels(
     y: np.ndarray,
     allow_missing_classes: bool = False,
 ) -> None:
-    """Check that labels is zero-indexed (first label is 0) and all classes present.
-    Assumes labels is 1D numpy array (not multi-label).
+    """Check that labels is properly formatted, i.e. a 1D array that is
+       zero-indexed (first label is 0) and all classes present (if ``allow_missing_classes is False``).
+       Assumes labels is 1D numpy array (not multi-label).
     """
     if y.ndim != 1:
         raise ValueError("labels must be 1D numpy array.")
--- a/cleanlab/internal/validation.py
+++ b/cleanlab/internal/validation.py
@@ -25,6 +25,7 @@ import numpy as np
 import pandas as pd
 
 
+# TODO: remove allow_missing_classes once supported
 def assert_valid_inputs(
     X: DatasetLike,
     y: LabelLike,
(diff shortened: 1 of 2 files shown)


cleanlab/internal/validation.py
```diff
@@ -96,22 +97,27 @@ def assert_valid_inputs(
...
-    if (unique_classes != np.arange(len(unique_classes))).any():
-        msg = "cleanlab requires zero-indexed integer labels (0,1,2,..,K-1), but in "
-        msg += "your case: np.unique(labels) = {}. ".format(str(unique_classes))
-        msg += "Every class in (0,1,2,..,K-1) must be present in labels as well."
-        raise TypeError(msg)
+        if (unique_classes != np.arange(len(unique_classes))).any():
+            msg = "cleanlab requires zero-indexed integer labels (0,1,2,..,K-1), but in "
+            msg += "your case: np.unique(labels) = {}. ".format(str(unique_classes))
+            msg += "Every class in (0,1,2,..,K-1) must be present in labels as well."
+            raise TypeError(msg)
```
elisno, line 120: I recommend inverting this condition and returning early.

```diff
-    if not allow_missing_classes:
-        unique_classes = np.unique(y)
-        if len(unique_classes) < 2:
-            raise ValueError("Labels must contain at least 2 classes.")
-
-        if (unique_classes != np.arange(len(unique_classes))).any():
-            msg = "cleanlab requires zero-indexed integer labels (0,1,2,..,K-1), but in "
-            msg += "your case: np.unique(labels) = {}. ".format(str(unique_classes))
-            msg += "Every class in (0,1,2,..,K-1) must be present in labels as well."
-            raise TypeError(msg)
+   if allow_missing_classes:
+       return None
+
+    unique_classes = np.unique(y)
+    if len(unique_classes) < 2:
+        raise ValueError("Labels must contain at least 2 classes.")
+
+    if (unique_classes != np.arange(len(unique_classes))).any():
+        msg = "cleanlab requires zero-indexed integer labels (0,1,2,..,K-1), but in "
+        msg += "your case: np.unique(labels) = {}. ".format(str(unique_classes))
+        msg += "Every class in (0,1,2,..,K-1) must be present in labels as well."
+        raise TypeError(msg)
```

Once missing classes are supported, you just have to remove two lines.

jwmueller: Thanks for the suggestion @elisno 
But it unfortunately won't be that simple once we support missing classes. We'll still want to check that the classes are integers and some other checks (like no class integer exceeds num_columns of `pred_probs`).  So I don't think we need to restructure the code like this now.

jwmueller (approved):
Added docstring clarification but LGTM

PR #336, opened 2022-08-09
Changed all docstring instances of np.array to np.ndarray

ulya-tkch:
* np.array is not a proper class name it is just a function to create np.ndarrays and therefore should not be parameter class


jwmueller (approved):
Thanks for the cleanup!

PR #338, opened 2022-08-12
Implementing get_ood_scores function

ulya-tkch:
* function structure


cleanlab/rank.py
```diff
@@ -652,3 +653,106 @@ def get_outlier_scores(
...
+    else:
+        raise ValueError(
+            f"""
+            {scoring_method} is not a valid ood scoring method!
+            Please choose a valid scoring_method: {valid_methods}
+            """
+        )
+
+    # Map ood_scores to range 0-1 with 0 = most concerning
+    ood_scores: np.ndarray = np.exp(-1 * ood_scores * t)
```
jwmueller, line 750: I don't think this exp(...) is how we should be transforming these scores to [0,1] with 0 = most concerning.

Note that least_confidence already satisfies this, so it does not need transformation (just don't do 1 minus it). Similarly 1 - normalized_entropy will also already satisfy this so no further transformation will be needed after that.

cleanlab/rank.py
```diff
@@ -652,3 +653,106 @@ def get_outlier_scores(
...
+            confident_thresholds = 0
+
+    if adjusted_pred_probs:
+        pred_probs = pred_probs - confident_thresholds
+        pred_probs += confident_thresholds.max()
+        pred_probs /= pred_probs.sum(axis=1)[:, None]
+
+    if scoring_method == "entropy":
+        ood_scores = get_normalized_entropy(pred_probs)
+    elif scoring_method == "self_confidence":
```
jwmueller, line 739: Let's change this method's name to: `"least_confidence"` to match our label-quality scoring paper and the uncertainty sampling literature.  Note it is slightly different than our self_confidence which measures the probability of the given label, not the probability of the argmax_class.

cleanlab/rank.py
```diff
@@ -652,3 +653,106 @@ def get_outlier_scores(
...
+            confident_thresholds = get_confident_thresholds(
+                labels, pred_probs, multi_label=False
+            )  # should we support multi_label?
+        else:
+            confident_thresholds = 0
+
+    if adjusted_pred_probs:
+        pred_probs = pred_probs - confident_thresholds
+        pred_probs += confident_thresholds.max()
+        pred_probs /= pred_probs.sum(axis=1)[:, None]
```
jwmueller, line 735: why not just use our existing: `_subtract_confident_thresholds()` function?

https://github.com/cleanlab/cleanlab/blob/master/cleanlab/internal/label_quality_utils.py#L23

cleanlab/rank.py
```diff
@@ -652,3 +653,106 @@ def get_outlier_scores(
...
+
+    if confident_thresholds is None:
+        if adjusted_pred_probs:
+            if labels is None:
+                raise ValueError(
+                    f"Cannot calculate adjusted_pred_probs without labels. Either pass in labels parameter or set adjusted_pred_probs = False."
+                )
+            confident_thresholds = get_confident_thresholds(
+                labels, pred_probs, multi_label=False
+            )  # should we support multi_label?
```
jwmueller, line 728: no let's not support multi_label in this function, it's unclear how that would work exactly

cleanlab/rank.py
```diff
@@ -652,3 +653,106 @@ def get_outlier_scores(
         return (outlier_scores, knn)
     else:
         return outlier_scores
+
+
+def get_ood_scores(
+    pred_probs: np.array,
+    confident_thresholds: Optional[np.ndarray] = None,
+    labels: Optional[np.ndarray] = None,
+    adjusted_pred_probs: bool = False,
```
jwmueller, line 662: change argname to `adjust_pred_probs` to match the other functions in `cleanlab.rank`. More generally, try to match the argnames / internal structure of the other existing functions in this file

cleanlab/rank.py
```diff
@@ -652,3 +653,106 @@ def get_outlier_scores(
         return (outlier_scores, knn)
     else:
         return outlier_scores
+
+
+def get_ood_scores(
+    pred_probs: np.array,
+    confident_thresholds: Optional[np.ndarray] = None,
+    labels: Optional[np.ndarray] = None,
```
jwmueller, line 661: Let's replace these 3 lines with:
```
pred_probs: np.array, *,
    labels: Optional[np.ndarray] = None,
    confident_thresholds: Optional[np.ndarray] = None,
```

Note I've changed the order of the latter two args, and also made them keyword only args, so that in the future we can swap their order again without breaking existing users' code.

cleanlab/rank.py
```diff
@@ -652,3 +653,106 @@ def get_outlier_scores(
         return (outlier_scores, knn)
     else:
         return outlier_scores
+
+
+def get_ood_scores(
+    pred_probs: np.array,
```
jwmueller, line 659: np.ndarray

cleanlab/rank.py
```diff
@@ -652,3 +653,106 @@ def get_outlier_scores(
...
+    confident_thresholds: Optional[np.ndarray] = None,
+    labels: Optional[np.ndarray] = None,
+    adjusted_pred_probs: bool = False,
+    scoring_method: str = "entropy",
```
jwmueller, line 663: shorten argname to `method` to better match other functions in this file

cleanlab/rank.py
```diff
@@ -652,3 +653,106 @@ def get_outlier_scores(
...
+    t: int = 1,
```
jwmueller, line 664: unnecessary arg, see comment below

cleanlab/rank.py
```diff
@@ -652,3 +653,106 @@ def get_outlier_scores(
...
+    return_thresholds: bool = False,
+) -> Union[np.ndarray, Tuple[np.ndarray, NearestNeighbors]]:
```
jwmueller, line 666: NearestNeighbors -> np.ndarray

cleanlab/rank.py
ulya-tkch: `_subtract_confident_thresholds() `calculates `confident_thresholds` intrenally. This would not work for some cases here such as user passes in `confident_thresholds` and no `labels`.

cleanlab/rank.py
ulya-tkch: Good idea!

cleanlab/rank.py
ulya-tkch: Prioritize same file format > other file formats?

cleanlab/rank.py
jwmueller: I see.  But I still think you could alter the `_subtract_confident_thresholds()` function to keep it working for the settings where it's currently being utilized (no change) while also working for this use-case here, right?  

Eg. could call it here with some dummy `labels = None` if there are no labels, and add an optional arg `confident_thresholds` (default value = None) in which case the _subtract...() function only computes the thresholds when this is None, otherwise uses the given thresholds.

jwmueller: Maybe I'm missing something and it's more work than is worthwhile to utilize _subtract_confident_thresholds here, but generally will want to minimize code duplication as much as possible!

cleanlab/rank.py
jwmueller: yep I think that's a reasonable heuristic :)

cleanlab/rank.py
ulya-tkch: Ok I can definitely open a PR doing that but without any alteration  using `_subtract_confident_thresholds()` won't work

ulya-tkch: Update: Added in newest commit. Not sure if setting pred_probs to optionally = None (to preserve variable ordering for other functions) was the cleanest way of doing that. Open to better suggestions!

ulya-tkch:
@jwmueller function returns `(ood_scores, confident_thresholds)` if confident_thresholds is not None. If `confident_thresholds=None` and `return_thresholds=True` then user warning is thrown and only ood_scores are returned.

Is that a good solution?

cleanlab/internal/label_quality_utils.py
```diff
@@ -15,30 +15,30 @@
...
 import numpy as np
+from typing import Optional
 from cleanlab.count import get_confident_thresholds
 
 
 def _subtract_confident_thresholds(
-    labels: np.ndarray,
+    labels: Optional[np.ndarray],
     pred_probs: np.ndarray,
+    confident_thresholds: Optional[np.ndarray] = None,
```
jwmueller, line 26: It's generally safer to add this extra arg to existing helper functions at the end of the arg-list, in case these helper functions are being used with positional arguments in the codebase.

ulya-tkch: Makes a lot of sense!

ulya-tkch:
TODO: 
* Switch to `class OutOfDistribution()`
* make adjust_pred_probs=True by default not false

ulya-tkch:
TODO: 
* Add docstring
* Add test_outlier.py

cleanlab/outlier.py
```diff
@@ -0,0 +1,175 @@
...
+    def _assert_valid_inputs(self, pred_probs, features):
+        if features is not None and pred_probs is not None:
+            raise ValueError(
+                f"Cannot fit to object to features and pred_probs. Pass in either one or the other."
+            )
+
+        if features is None and pred_probs is None:
+            raise ValueError(
+                f"Not enough information to compute scores. Pass in either features or pred_probs."
+            )
```
jwmueller, line 130: maybe also check features is 2D array if it's not None?

cleanlab/outlier.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if features is not None:
+            # fit to outlier scores
+            if verbose:
+                print("Computing outlier scores based on provided features ...")
+
+        if pred_probs is not None:
+            # fit to ood scores
+            if verbose:
+                print("Computing ood scores based on provided pred_probs ...")
```
jwmueller, line 63: as your comment mentions, why not move this into _shared_fit ?

cleanlab/outlier.py
```diff
@@ -0,0 +1,198 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
+
+"""
+Class to identify which examples are out of distribution.
+"""
+import warnings
+from typing import Optional
+import numpy as np
+from cleanlab.rank import get_ood_scores, get_outlier_scores
```
jwmueller, line 23: These can just be private methods defined inside this file, instead of public methods inside `rank.py`, right?

cleanlab/outlier.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class OutOfDistribution:
+    OUTLIER_PARAMS = tuple(("k", "t"))
+    OOD_PARAMS = tuple(("labels", "adjust_pred_probs", "method"))
```
jwmueller, line 28: Why do you think these should be tuples instead of list or set?

cleanlab/outlier.py
ulya-tkch: Yes! Moving those over today

cleanlab/outlier.py
ulya-tkch: Shared fit is called for fit() and fit_score(). Saying "Computing outlier scores based on provided features ..." does not make sense for fit function since we are not returning scores. 

Changing this to "Fitting OOD object based on provided features ..." and moving into _shared_fit() but let me know if I should revert

cleanlab/outlier.py
ulya-tkch: Set is best

cleanlab/outlier.py
```diff
@@ -0,0 +1,541 @@
...
+        assert isinstance(scores, np.ndarray)
+        return scores
+
+    def _get_params(self, param_keys) -> dict:
+        """
+        Helper method to get function specific dictionary of parameters (i.e. only those in param_keys).
+        """
+        return {k: v for k, v in self.params.items() if k in param_keys}
+
+    def _get_invalid_params(self, params, param_keys) -> list:
```
jwmueller, line 283: doesn't seem like this should be instance method (does not use `self` AFAICT)

cleanlab/outlier.py
```diff
@@ -0,0 +1,541 @@
...
+        """
+        Helper method to get list of parameters in param that are not in param_keys.
+        """
+        if params is None:
+            return []
+        return list(set(params.keys()).difference(set(param_keys)))
+
+    def _assert_valid_inputs(self, features, pred_probs):
```
jwmueller, line 291: doesn't seem like this should be instance method (does not use `self` AFAICT)

cleanlab/outlier.py
```diff
@@ -0,0 +1,541 @@
...
+        return scores
+
+    def _get_params(self, param_keys) -> dict:
+        """
+        Helper method to get function specific dictionary of parameters (i.e. only those in param_keys).
+        """
+        return {k: v for k, v in self.params.items() if k in param_keys}
+
+    def _get_invalid_params(self, params, param_keys) -> list:
+        """
```
jwmueller, line 284: What about also ensuring the validity of `params` in this function (if it was passed by user, ie. is not None)?

In particular could check that `params` does not contain an unknown keys, as well as any keys that are only valid when only `features` was provided (and same thing for when only `pred_probs` was provided.

cleanlab/outlier.py
```diff
@@ -0,0 +1,541 @@
...
+
+
+def _get_ood_features_scores(
+    features: Optional[np.ndarray] = None,
+    knn: Optional[NearestNeighbors] = None,
+    k: Optional[int] = None,
+    t: int = 1,
+    return_estimator: bool = False,
+) -> Union[np.ndarray, Tuple[np.ndarray, NearestNeighbors]]:
+    """Returns an outlier score for each example based on its feature values.
```
jwmueller, line 382: which is computed inversely proportional to the average distance between this example and its K nearest neighbors (in feature space).

cleanlab/outlier.py
```diff
@@ -0,0 +1,541 @@
...
+                  a different value of `k` than originally used in the ``knn``,
+                  as long as your specified value of `k` is smaller than the value originally used in ``knn``.
+            *  t : int, default=1
+                  Optional hyperparameter only for advanced users.
+                  Controls transformation of distances between examples into similarity scores that lie in [0,1].
+                  The transformation applied to distances `x` is `exp(-x*t)`.
+                  If you find your scores are all too close to 1, consider increasing `t`,
+                  although the relative scores of examples will still have the same ranking across the dataset.
+
+          If `pred_probs` is passed in, `params` could contian following keys:
```
jwmueller, line 190: typo: contian

cleanlab/outlier.py
```diff
@@ -0,0 +1,541 @@
...
+            *  adjust_pred_probs : bool, True
+                  Account for class imbalance in the label-quality scoring by adjusting predicted probabilities
+                  via subtraction of class confident thresholds and renormalization.
+                  Set this to ``False`` if you prefer to skip accounting for class-imbalance.
```
jwmueller, line 194: can also add that: You do not have to pass in `labels` in this case.

cleanlab/outlier.py
```diff
@@ -0,0 +1,541 @@
...
+                  See `Northcutt et al., 2021 <https://jair.org/index.php/jair/article/view/12125>`_.
```
jwmueller, line 195: keep track of TODO to replace this with link to our writeup describing this methodology.

tests/test_outlier.py
```diff
@@ -0,0 +1,552 @@
...
+
+    # TESTING: params dict is a copy
+    params_dict = {"k": 10, "t": 5}
+    OOD.fit(features=features, params=params_dict)
+    ood_params = OOD.params
+    params_dict = params_dict.update({"k": 20})
+    assert ood_params == OOD.params
+
+    # test calling functions with different params performs differently
+
```
jwmueller, line 241: also test that you can set `labels=None` while providing pred_probs if `adjust_pred_probs=False`?

cleanlab/outlier.py
elisno:
```suggestion
    @staticmethod
    def _assert_valid_inputs(features, pred_probs):
```
allows you to use `self._assert_valid_inputs(features, pred_probs)` within the scope of the class without setting `self` as the first argument.

cleanlab/outlier.py
elisno:
```suggestion
    @staticmethod
    def _get_invalid_params(params, param_keys) -> list:
```

cleanlab/outlier.py
ulya-tkch: Thank you! This is a lot cleaner :D

cleanlab/outlier.py
```diff
@@ -0,0 +1,545 @@
+# Copyright (C) 2017-2022  Cleanlab Inc.
```
jwmueller, line 1: Leaving comment here that actually applies to the `outliers.ipynb` notebook instead (but github won't let me comment there due to diff size):

Rename "How does cleanlab detect outliers?"
-> "How does cleanlab detect outliers from feature values?"
and move this section up to the end of the features-based OOD detection part of the tutorial, before you launch into pred_probs based OOD detection.

docs/source/tutorials/outliers.ipynb
```diff
@@ -318,35 +349,38 @@
...
    "cell_type": "code",
    "execution_count": null,
-   "id": "aef10ced",
+   "id": "feb0f519",
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Get outlier scores for each of our train_data feature embeddings\n",
-    "train_outlier_scores, knn = get_outlier_scores(features=train_feature_embeddings, return_estimator=True)\n",
+    "# Create OutOfDistribution object\n",
```
jwmueller, line 369: Such comments probably aren't necessary when the code is self-explanatory.  This notebook is on the longer side, so shortening it at any opportunity would be good!

docs/source/tutorials/outliers.ipynb
```diff
@@ -355,62 +389,56 @@
...
   {
    "cell_type": "markdown",
-   "id": "61020539",
+   "id": "2521aefb",
    "metadata": {},
    "source": [
     "#### Scoring outliers in additional test data\n",
     "\n",
-    "Now suppose we want to find outlier images in our additional test data, in particular images unlikely to stem from the same distribution as the training data. We can again use `get_outlier_scores()` this time passing in the previously-fitted estimator `knn` used to measure similarity between training examples."
+    "Now suppose we want to find outlier images in some never before seen test data, in particular images unlikely to stem from the same distribution as the training data. We can again use our fitted `OutOfDistribution` object to measure similarity between these new testing examples."
```
jwmueller, line 409: Instead of "OutOfDistribution object", what if we call it "OutOfDistribution estimator"?

jwmueller: "to measure similarity between these new testing examples"
->
"to score how typical each new test example would be under the the training data distribution"

docs/source/tutorials/outliers.ipynb
```diff
@@ -2,46 +2,75 @@
...
+    "Detecting outliers using `feature_embeddings`\n",
     "- Pre-process [cifar10](https://www.cs.toronto.edu/~kriz/cifar.html) to create a Pytorch dataset.\n",
     "- Create `train_data` and `test_data` such that `train_data` only contains images of animals and `test_data` contains images from all classes.\n",
     "- Load pretrained `timm` neural network model and extract `train_data` and `test_data` feature embeddings from its representations.\n",
     "- Use `cleanlab` to find naturally occurring outlier examples in the `train_data` (i.e. atypical images).\n",
     "- Find outlier examples in the `test_data` that do not stem from training data distribution (including out-of-distribution non-animal images).\n",
-    "- Explore threshold selection for determining which images are outliers vs not."
+    "- Explore threshold selection for determining which images are outliers vs not.\n",
+    "\n",
+    "Detecting outliers using `pred_probs`\n",
```
jwmueller, line 22: Detecting outliers using `pred_probs` from a trained classifier

docs/source/tutorials/outliers.ipynb
```diff
@@ -2,46 +2,75 @@
...
+    "- Train Logistic Regression classifier on the feature embeddings generated by `timm` model.\n",
```
jwmueller, line 23: Turn our `timm` model into a classifier by adding an output layer fit via Logistic Regression to the (in-distribution) training data.

docs/source/tutorials/outliers.ipynb
```diff
@@ -2,46 +2,75 @@
...
+    "\n",
+    "<div  class=markdown markdown=\"1\" style=\"background:white;margin:16px\">  \n",
+    "    \n",
+    "```python\n",
+    "\n",
+    "from cleanlab.outlier import OutOfDistribution\n",
+    "    \n",
+    "OOD = OutOfDistribution()\n",
+    "\n",
+    "# To get outlier scores for train_data using train prediction probabilities and example labels\n",
```
jwmueller, line 69: "train prediction probabilities and example labels"
->
"predicted class probabilities (from a trained classifier) and given class labels"

docs/source/tutorials/outliers.ipynb
```diff
@@ -2,46 +2,75 @@
...
+    "ood_train_predictions_scores = OOD.fit_score(pred_probs=train_pred_probs, labels=labels)\n",
+    "\n",
+    "# To get outlier scores for additional test_data using test prediction probabilities\n",
```
jwmueller, line 72: "test prediction probabilities" 
-> 
"predicted class probabilities"

docs/source/tutorials/outliers.ipynb
```diff
@@ -516,10 +544,105 @@
     "</div>"
```
jwmueller, line 544: This </div> appears to be extra

docs/source/tutorials/outliers.ipynb
```diff
@@ -516,10 +544,105 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "id": "937c7e97",
+   "metadata": {},
+   "source": [
+    "## 5. Use cleanlab and `pred_probs` to find outliers in the dataset\n",
+    "\n",
+    "\n",
+    "Another way we can detect unusual exampels in the data is using prediction probabilities for out-of-distribution detection. Instead of using intermediate feature representations, this type of outlier detection simply needs a model's `pred_probs` to work. These methods are faster than using `feature_embeddings` but as a tradeoff perform slightly worse. This is due to the intermediate feature embeddings containing more auxillary information for every example, only some of which is represented in the final predicted probabilites.\n",
```
jwmueller, line 555: typo: exampels

jwmueller: generally want to say "predicted probabilities" or "predicted class probabilities" instead of "prediction probabilities"

jwmueller: "as a tradeoff" -> "may"

docs/source/tutorials/outliers.ipynb
```diff
@@ -516,10 +544,105 @@
...
+    "\n",
+    "To get `pred_probs` for this example, a Logistic Regression Classifier is fit on the already generated `train_feature_embeddings` from the outlier example above."
```
jwmueller, line 557: a 4-class Logistic Regression classifier

jwmueller: add:  

and the given (in-distribution) label for each training example.  This mimics the standard practice of appending a fine-tuned  output layer at the top of a pretrained neural network to adapt it for a particular classification task (you could do that as well in this tutorial instead of our Logistic Regression, the fine-tuning just may run slower).

docs/source/tutorials/outliers.ipynb
```diff
@@ -516,10 +544,105 @@
...
+    "train_labels = np.array(train_data.dataset.targets)[train_data.indices]\n",
+    "train_labels = np.unique(train_labels, return_inverse=True)[1] # Zero index out train labels\n",
+    "test_labels = np.array(test_data.dataset.targets)[test_data.indices]\n",
+    "\n",
+    "scaler = preprocessing.StandardScaler().fit(train_feature_embeddings)\n",
+    "train_feature_embeddings_scaled = scaler.transform(train_feature_embeddings)\n",
+    "test_feature_embeddings_scaled = scaler.transform(test_feature_embeddings)\n",
+    "\n",
+    "# Create model and fit on scaled features and zero indexed train labels (may take a few minutes)\n",
+    "model = LogisticRegression(penalty='l2', C=1, solver='sag', max_iter=500)\n",
```
jwmueller, line 577: sorry I overlooked this before, but here we should wrap the LogisticRegression with `sklearn.model_selection.cross_val_predict` as done in our other tutorials:  https://docs.cleanlab.ai/v2.0.0/tutorials/tabular.html

and explain why it's important to use held-out pred_probs. For explanation, you can just link our [cross-validation tutorial](https://docs.cleanlab.ai/v2.0.0/tutorials/pred_probs_cross_val.html), but be sure to format a proper relative rather than absolute link since this needs work for all versions of cleanlab (hopefully you can find an example in existing codebase).

jwmueller: That is also important for the accuracy evaluation below to be meaningful at all.

jwmueller: To later get predictions for the test data, you can either: retrain another LogisticRegression model on the full training set, or average the pred_probs from each fold's LogisticRegression that was produced during cross-validation.

docs/source/tutorials/outliers.ipynb
```diff
@@ -516,10 +544,105 @@
...
+    "# top_subset = torch.utils.data.Subset(train_data, top_idxs)\n",
+    "# plot_images(top_subset)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "eb631419",
+   "metadata": {},
+   "source": [
+    "Out-of-distribution predictions scores can also be computed on the test data by first calculating `pred_probs` using `test_feature_embeddings`. To do this, an `OutOfDistribution` object fitted on `train_pred_probs` can be used to score `test_pred_probs`.\n",
```
jwmueller, line 617: The sentence "To do this, ..." is confusing, because it sounds like you are using OutOfDistribution to calculate pred_probs. 

I would edit these sentences to instead read something like this: 
"Out-of-distribution predictions scores can also be computed to estimate how likely additional test data are under the training data distribution. Here we first apply our trained classifier to obtain `pred_probs` for the test data and then pass these `pred_probs` into our already-fitted OutOfDistribution estimator."

jwmueller: Side-Note: we really want to minimize mention of the test_feature_embeddings throughout this section, as typically  these would not be relevant for computing pred_probs at all and our application of LogisticRegression to them here is nonstandard and mainly just for convenience.

docs/source/tutorials/outliers.ipynb
```diff
@@ -516,10 +544,105 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3f90d8a4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Get model predictions on test set\n",
+    "pred_probs = model.predict_proba(test_feature_embeddings_scaled)\n",
+    "pred_labels = pred_probs.argmax(1)\n",
```
jwmueller, line 631: I think you can delete this line since it's not used elsewhere?

docs/source/tutorials/outliers.ipynb
```diff
@@ -516,10 +544,105 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e110fc4b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Get model predictions and measure accuracy on trainset\n",
+    "train_pred_probs = model.predict_proba(train_feature_embeddings_scaled)\n",
+    "pred_labels = train_pred_probs.argmax(1)\n",
```
jwmueller, line 598: [nit] it's usually clearer to write: `argmax(axis=1)`

docs/source/tutorials/outliers.ipynb
```diff
@@ -528,9 +651,13 @@
...
     "# Verify the top identified test outliers data are mostly non-animal images\n",
-    "num_animals = len([i for i in range(len(top_outlier_subset)) if top_outlier_subset[i][1] in animal_classes])\n",
-    "if 1 - (num_animals / len(top_outlier_subset)) < 0.81:\n",
-    "    raise Exception(\"Not enough non-animal images amongst top-ranked outliers in test_data.\")"
+    "num_animals = len([i for i in range(len(top_ood_features_subset)) if top_ood_features_subset[i][1] in animal_classes])\n",
+    "if 1 - (num_animals / len(top_ood_features_subset)) < 0.81:\n",
+    "    raise Exception(\"Not enough non-animal images amongst top-ranked outliers in test_data.\")\n",
+    "    \n",
+    "num_animals = len([i for i in range(len(top_ood_predictions_subset)) if top_ood_predictions_subset[i][1] in animal_classes])\n",
+    "if 1 - (num_animals / len(top_ood_predictions_subset)) < 0.51:\n",
```
jwmueller, line 659: we should trying tweaking settings of the model until this 0.51 is at least 0.65 or so, to hopefully ensure somewhat reasonable performance. Could also try non-LogisticRegression sklearn models such as  `KNeighborsClassifier(n_neighbors=20, weights='distance', metric='cosine')` if just altering the LogisticRegression hyperparameters does not suffice to get good performance.

cleanlab/outlier.py
```diff
@@ -0,0 +1,534 @@
...
+
+class OutOfDistribution:
+    """
+    OutOfDistribution = Out of distribution detection for classification examples using intermediate feature
+    embeddings or predicted probabilities. Each passed in example is given an out-of-distribution score. Scores lie in [0,1] with smaller
+    values indicating examples that are less typical under the dataset distribution (values near 0 indicate outliers).
+
+    Parameters
+    ----------
+    params : dict, default = None
```
jwmueller, line 39: default is `{}` which is better than None

cleanlab/outlier.py
```diff
@@ -0,0 +1,534 @@
...
+        "adjust_pred_probs": True,  # ood pred_probs param
+        "method": "entropy",  # ood pred_probs param
+        "confident_thresholds": None,  # ood pred_probs param
+    }
+
+    def __init__(self, params: dict = {}):
+        self._assert_valid_params(params, self.DEFAULT_PARAM_DICT)
+        self.params = self.DEFAULT_PARAM_DICT
+        if params is not None:
+            self.params = {**self.params, **params}
```
jwmueller, line 101: Get rid of the `params is not None` check, since we expect empty dict `{}` as the default. We always want to execute the below update since it's fast and ensures we have safely copied the dict. 

We always want to be extremely cautious about assigning dicts to other variables without having created a deepcopy of them.

cleanlab/outlier.py
```diff
@@ -0,0 +1,534 @@
...
+          Feature array of shape ``(N, M)``, where N is the number of examples and M is the number of features used to represent each example.
+          For details, `features` in the same format expected by the :py:func:`fit <cleanlab.outlier.OutOfDistribution.fit>` function.
+
+        pred_probs : np.ndarray, optional
+          An array of shape ``(N, K)`` of model-predicted probabilities.
+          For details, `pred_probs` in the same format expected by the :py:func:`fit <cleanlab.outlier.OutOfDistribution.fit>` function.
+
+        labels : np.ndarray, optional
+          A discrete vector of noisy labels, i.e. some labels may be erroneous of shape ``(N,)``.
+          For details, `labels` in the same format expected by the :py:func:`fit <cleanlab.outlier.OutOfDistribution.fit>` function.
```
jwmueller, line 133: "noisy labels, i.e. some labels may be erroneous" -> "given labels for the data"

cleanlab/outlier.py
```diff
@@ -0,0 +1,534 @@
...
+
+          **Caution**: `pred_probs` from your model must be out-of-sample!
+          You should never provide predictions on the same examples used to train the model,
+          as these will be overfit and unsuitable for finding label-errors.
+          To obtain out-of-sample predicted probabilities for every datapoint in your dataset, you can use :ref:`cross-validation <pred_probs_cross_val>`.
+          Alternatively it is ok if your model was trained on a separate dataset and you are only evaluating
+          data that was previously held-out.
+
+        labels : np.ndarray, optional
+          A discrete vector of noisy labels, i.e. some labels may be erroneous of shape ``(N,)``.
```
jwmueller, line 202: "noisy labels, i.e. some labels may be erroneous" -> "given labels for the data"

cleanlab/outlier.py
```diff
@@ -0,0 +1,534 @@
...
+          *Format requirements*: for dataset with K classes, labels must be in 0, 1, ..., K-1.
+          All the classes (0, 1, ..., and K-1) MUST be present in ``labels``, such that: ``len(set(labels)) == pred_probs.shape[1]``
+          Note: multi-label classification is not supported by this method, each example must belong to a single class, e.g. format: ``labels = np.ndarray([1,0,2,1,1,0...])``.
+
```
jwmueller, line 206: Clarify in what situations `labels` must be specified.

cleanlab/outlier.py
```diff
@@ -0,0 +1,534 @@
...
+        """
+        # Assert valid inputs and params
+        self._assert_valid_inputs(features, pred_probs)
+        scores = None  # If none scores are returned, fit was skipped
+
+        if features is not None:
+            if self.params["knn"] is not None:
+                # No fitting twice if knn object already fit
+                warnings.warn(
+                    f"Skipping fit. Object already fit on knn estimator.",
```
jwmueller, line 335: "Object already fit on knn estimator"
->
"A KNN estimator has previously already been fit, call `score()` to use it, or create a new `OutOfDistribution` object to fit a different estimator."

cleanlab/outlier.py
```diff
@@ -0,0 +1,534 @@
...
+
+    confident_thresholds : np.ndarray, default = None
+      For details on 'confindent_thresholds` see :py:class:`OutOfDistribution <cleanlab.outlier.OutOfDistribution>`.
+
+    labels : np.ndarray, optional
+      `labels` in the same format expected by the :py:func:`fit <cleanlab.outlier.OutOfDistribution.fit>` function.
+
+    adjust_pred_probs : bool, True
+      Account for class imbalance in the label-quality scoring.
+      For details, see `adjust_pred_probs` in :py:func:`fit <cleanlab.outlier.OutOfDistribution.fit>` function.
```
jwmueller, line 476: For details, see key `adjust_pred_probs` in the `params` dict arg of  cleanlab.outlier.OutOfDistribution class

(I presume you can figure out how to properly link the spot in docs for this class where this documentation appears)

cleanlab/outlier.py
```diff
@@ -0,0 +1,534 @@
...
+    """
+    OutOfDistribution = Out of distribution detection for classification examples using intermediate feature
+    embeddings or predicted probabilities. Each passed in example is given an out-of-distribution score. Scores lie in [0,1] with smaller
+    values indicating examples that are less typical under the dataset distribution (values near 0 indicate outliers).
+
+    Parameters
+    ----------
+    params : dict, default = None
+     Optional keyword arguments to that change how estimator is fit. Effect of arguments passed in depends on if
+     `OutOfDistribution` object is fit on `features` or `pred_probs`.
```
jwmueller, line 41: "Effect of arguments passed in depends on if `OutOfDistribution` object is fit on `features` or `pred_probs`"
->
"Effect of arguments passed in here depend on if `OutOfDistribution` object will be fit to `features` or `pred_probs`"

cleanlab/outlier.py
```diff
@@ -0,0 +1,534 @@
...
+                )
+            else:
+                scores, _ = _get_ood_features_scores(
+                    features, **self._get_params(self.OUTLIER_PARAMS)
+                )
+
+        if pred_probs is not None:
+            if self.params["confident_thresholds"] is None and self.params["adjust_pred_probs"]:
+                raise ValueError(
+                    f"OOD Object needs to be fit on features first. Call fit() or fit_scores() before this function."
```
jwmueller, line 266: better error message:

OOD estimator needs to be fit on `pred_probs` first since `params["adjust_pred_probs"]=True`

cleanlab/outlier.py
```diff
@@ -0,0 +1,534 @@
...
+                scores, knn = _get_ood_features_scores(
+                    features, **self._get_params(self.OUTLIER_PARAMS)
+                )
+                self.params["knn"] = knn
+
+        if pred_probs is not None:
+            if self.params["confident_thresholds"] is not None:
+                # No fitting twice if confident_thresholds object already fit
+                warnings.warn(
+                    f"Skipping fit. Object already fit with confident thresholds.",
```
jwmueller, line 351: "Object already fit with confident thresholds"
->
"This estimator has previously already been fit, call score() to apply it to data, or create a new `OutOfDistribution` object to fit a different estimator with different confident thresholds."

docs/source/tutorials/outliers.ipynb
ulya-tkch: Just checking you meant a 6 class logistic regression classifier right? 6 animal classes in the trainset

docs/source/tutorials/outliers.ipynb
ulya-tkch: Decided to retrain for simplicity!

docs/source/tutorials/outliers.ipynb
ulya-tkch: Similar performance to logistic regression. Visually, all the examples look somewhat like outliers so not sure if 0.51 vs 0.65 (1 example) make a big difference.

docs/source/tutorials/outliers.ipynb
jwmueller: yep! sorry apparently I can only count on one hand ...

docs/source/tutorials/outliers.ipynb
jwmueller: I just think the evaluation is not great if we're only getting 0.51 (essentially random performance under our metric) yet we think the results are pretty good. The fact that we've decided to evaluate via animal vs non-animal means we believe this distinction to be more meaningful than finding strange animal images. If not, we should reconsider the setup I think. But I like this setup of having evaluation where true OOD examples = those whose class was not present in the training data.

docs/source/tutorials/outliers.ipynb
```diff
@@ -476,50 +491,140 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3f90d8a4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "model = MLPClassifier(solver='adam', hidden_layer_sizes=(512, 64, 6), random_state=1)\n",
```
jwmueller, line 614: instead of hardcoding the 6 here, it will be less confusing if at the top of the notebook, where you first defined `animal_classes`, you add line:
```
num_classes_indistribution = 6
```
and then use `num_classes_indistribution` in its place everywhere later.

docs/source/tutorials/outliers.ipynb
```diff
@@ -2,46 +2,75 @@
  "cells": [
```
jwmueller, line 2: Top of notebook is missing import for MLPClassifier from sklearn

jwmueller: import of LogisticRegression is no longer needed also

docs/source/tutorials/outliers.ipynb
```diff
@@ -2,46 +2,75 @@
...
+    "Already have `pred_probs` and `labels` for your data? Just run the code below to compute out-of-distribution predictions scores.\n",
+    "\n",
+    "\n",
+    "<div  class=markdown markdown=\"1\" style=\"background:white;margin:16px\">  \n",
+    "    \n",
+    "```python\n",
+    "\n",
+    "from cleanlab.outlier import OutOfDistribution\n",
+    "    \n",
+    "OOD = OutOfDistribution()\n",
```
jwmueller, line 67: I think using capitalized variable name is nonconventional?

https://peps.python.org/pep-0008/#:~:text=Prescriptive%3A%20Naming%20Conventions,-Names%20to%20Avoid&text=Never%20use%20the%20characters%20'l,the%20numerals%20one%20and%20zero.

cleanlab/outlier.py
```diff
@@ -0,0 +1,535 @@
...
+            f"score calculation.",
+            UserWarning,
+        )
+
+    if adjust_pred_probs:
+        if confident_thresholds is None:
+            if labels is None:
+                raise ValueError(
+                    f"Cannot calculate adjust_pred_probs without labels. Either pass in labels parameter or set "
+                    f"adjusted_pred_probs = False. "
```
jwmueller, line 507: should be `params["adjust_pred_probs"]` instead of `adjusted_pred_probs`.  This should be fixed throughout, you may be using that wrong name in other places as well

jwmueller: "labels parameter" 
->
"parameter \`labels\`"

docs/source/tutorials/outliers.ipynb
ulya-tkch: changing to ood

docs/source/tutorials/outliers.ipynb
```diff
@@ -476,50 +495,127 @@
...
+   "outputs": [],
+   "source": [
+    "train_pred_labels = train_pred_probs.argmax(1)\n",
+    "accuracy = np.mean(train_pred_labels == train_labels)\n",
+    "print(f\"Model accuracy on held-out train_data {accuracy}\")\n",
+    "\n",
+    "ood = OutOfDistribution()\n",
+    "train_ood_predictions_scores = ood.fit_score(pred_probs=train_pred_probs, labels=train_labels)\n",
+    "\n",
+    "model = BaggingClassifier(MLPClassifier(solver='adam', hidden_layer_sizes=(512, 64, num_classes_indistribution), random_state=1),random_state=1, n_jobs=-1)\n",
```
jwmueller, line 608: I don't think we need the `random_state=1` inside of the MLPClassfier?

From: 
https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.BaggingClassifier.html
> random_state Controls the random resampling of the original dataset (sample wise and feature wise). If the base estimator accepts a random_state attribute, a different seed is generated for each instance in the ensemble.

jwmueller: why not include: `early_stopping=True` in MLPClassifier?

docs/source/tutorials/outliers.ipynb
```diff
@@ -528,9 +624,15 @@
...
-    "if 1 - (num_animals / len(top_outlier_subset)) < 0.81:\n",
-    "    raise Exception(\"Not enough non-animal images amongst top-ranked outliers in test_data.\")"
+    "top_ood_features_subset = torch.utils.data.Subset(test_data, top_ood_features_idxs)\n",
+    "top_ood_predictions_subset = torch.utils.data.Subset(test_data, top_ood_predictions_idxs)\n",
+    "num_animals = len([i for i in range(len(top_ood_features_subset)) if top_ood_features_subset[i][1] in animal_classes])\n",
+    "if 1 - (num_animals / len(top_ood_features_subset)) < 0.81:\n",
+    "    raise Exception(\"Not enough non-animal images amongst top-ranked outliers in test_data.\")\n",
+    "    \n",
+    "num_animals = len([i for i in range(len(top_ood_predictions_subset)) if top_ood_predictions_subset[i][1] in animal_classes])\n",
+    "if 1 - (num_animals / len(top_ood_predictions_subset)) < 0.50:\n",
```
jwmueller, line 634: If only 40% of the test data is non-animal, maybe threshold like 0.45 is good enough for CI purposes

jwmueller (approved):
This PR adds: OutOfDistribution class with ability to score outliers based on either: feature values, or pred_probs and labels.

PR #379, opened 2022-08-25
Switch docs links from latest release to stable

elisno:
Find and replace 'docs.cleanlab.ai/v2.0.0/' with 'docs.cleanlab.ai/stable/'.

Only occurs in README.md.


jwmueller (approved):
Very happy to no longer have to update these links going forward!

PR #381, opened 2022-08-31
Add missing docs pages

jwmueller:
Some docs for newly added source code files were not being shown.  This PR ensures the documentation for some of these files is shown at docs.cleanlab.ai (not yet done for all the token classification files until its mypy typing and unit tests are more finalized).

In the future, we'll want CI to also check for potential issues like this when contributors add new source code files.


huiwengoh (approved):
LGTM!

PR #389, opened 2022-08-31
Change verbose functionality in multiannotator functions

huiwengoh:
- use `verbose` arg to specify whether or not to print warnings
- used the subset of examples with more than one annotation for `most_likely_class_accuracy` to maintain consistency with other accuracy computations
- clipped weights at 1e-6 to prevent division by zero issues


jwmueller (approved):
LGTM

Issue #384, opened 2022-08-31
Token Classification code release


jwmueller:
@- [x] Verify tutorials run well, and add any improvements you have for the main tutorial
- [x] Add Typing 
- [x] Add unit tests to improve code coverage
- [x] Add files to make the new code for token_classification module appear in docs.cleanlab.ai 
- [x] Improve documentation formatting
- [x] Replace `import *` from files to only import the necessary functions from a module. 
- [x] #388 
- [x] review example notebook and verify it runs well (on GPU machine):  https://github.com/cleanlab/examples/pull/7
- [x] #400
- [x] #414
- [x] #401

Relevant previous PRs:
https://github.com/cleanlab/cleanlab/pull/347
https://github.com/cleanlab/examples/pull/7

jwmueller:
Added one more small TODO:  Replace `import *` from files to only import the necessary functions from a module.

PR #390, opened 2022-08-31
Cleanup token classification utils

elisno:
As part of #384 , this PR adds the following to cleanlab/internal/token_classification_utils.py:

- Improved unit tests & Full test coverage
- Strict typing
- Docstrings

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,9 +50,6 @@ jobs:
           pip install -r requirements-dev.txt  # install development dependencies and type stubs
       - name: Type check
         run: mypy --install-types --non-interactive cleanlab
-      - name: Strict type check
-        run: |
-          mypy --install-types --non-interactive --strict cleanlab/internal/token_classification_utils.py
   fmt:
     name: Format
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,9 @@ jobs:
           pip install -r requirements-dev.txt  # install development dependencies and type stubs
       - name: Type check
         run: mypy --install-types --non-interactive cleanlab
+      - name: Strict type check
+        run: |
+          mypy --install-types --non-interactive --strict {cleanlab/internal/token_classification_utils.py,}
   fmt:
     name: Format
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         run: mypy --install-types --non-interactive cleanlab
       - name: Strict type check
         run: |
-          mypy --install-types --non-interactive --strict {cleanlab/internal/token_classification_utils.py,}
+          mypy --install-types --non-interactive --strict cleanlab/internal/token_classification_utils.py
   fmt:
     name: Format
     runs-on: ubuntu-latest
--- a/cleanlab/internal/token_classification_utils.py
+++ b/cleanlab/internal/token_classification_utils.py
@@ -1,6 +1,7 @@
 import re
 import string
 import numpy as np
+from numpy.typing import NDArray
 from termcolor import colored
 from typing import List, Union, Optional, Callable, Tuple
 
@@ -30,8 +31,10 @@ def get_sentence(words: List[str]) -> str:
 
 
 def filter_sentence(
-    sentences: List[str], condition: Optional[Callable] = None, return_mask: bool = True
-) -> Union[Tuple[list, list], list]:
+    sentences: List[str],
+    condition: Optional[Callable[[str], bool]] = None,
+    return_mask: bool = True,
+) -> Union[Tuple[List[str], List[bool]], List[str]]:
     """
     Filter sentence based on some condition, and returns filter mask
 
@@ -40,7 +43,7 @@ def filter_sentence(
         sentences: List[str]
             list of sentences
 
-        condition: Callable
+        condition: Optional[Callable[[str], bool]]
             sentence filtering condition
 
         return_mask: bool
@@ -94,21 +97,21 @@ def process_token(token: str, replace: List[Tuple[str, str]] = [("#", "")]) -> s
     return processed_token
 
 
-def mapping(entities: list, maps: list) -> list:
+def mapping(entities: List[int], maps: List[int]) -> List[int]:
     """
     Map a list of entities to its corresponding entities
 
     Parameters
     ----------
-        entities: list
+        entities: List[int]
             a list of given entities
 
-        maps:
+        maps: List[int]
             a list of mapped entities, such that the i'th indexed token should be mapped to `maps[i]`
 
     Returns
     ---------
-        mapped_entities:
+        mapped_entities: List[int]
             a list of mapped entities
 
     Examples
@@ -124,13 +127,13 @@ def mapping(entities: list, maps: list) -> list:
     return list(map(f, entities))
 
 
-def merge_probs(probs: np.ndarray, maps: List[int]) -> np.ndarray:
+def merge_probs(probs: NDArray[np.float64], maps: List[int]) -> NDArray[np.float64]:
     """
     Merges model-predictive probabilities with desired mapping
 
     Parameters
     ----------
-        probs: np.array
+        probs: NDArray[np.float64]
             np.array of shape `(N, K)`, where N is the number of tokens, and K is the number of classes for the model
 
         maps: List[int]
@@ -140,13 +143,14 @@ def merge_probs(probs: np.ndarray, maps: List[int]) -> np.ndarray:
 
     Returns
     ---------
-        probs_merged: np.array
+        probs_merged: NDArray[np.float64]
             np.array of shape `(N, K')`, where K' is the number of new classes. Probablities are merged and
             re-normalized if necessary.
 
     """
     old_classes = probs.shape[1]
-    probs_merged = np.zeros([len(probs), np.max(maps) + 1])
+    map_size = np.max(maps) + 1  # type: ignore
+    probs_merged = np.zeros([len(probs), map_size], dtype=probs.dtype.type)
 
     for i in range(old_classes):
         if maps[i] >= 0:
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -50,6 +50,9 @@ jobs:
           pip install -r requirements-dev.txt  # install development dependencies and type stubs
       - name: Type check
         run: mypy --install-types --non-interactive cleanlab
+      - name: Strict type check
+        run: |
+          mypy --install-types --non-interactive --strict {cleanlab/internal/token_classification_utils.py,}
```
elisno, line 55: The idea is to add files we know will pass the strict type-checker to this list.
It would be nice if we didn't have to write it like:
```yaml
- name: Strict type check
   run: |
     mypy --install-types --non-interactive --strict foo.py
     mypy --install-types --non-interactive --strict bar.py
     mypy --install-types --non-interactive --strict foobar.py
     ...
```
or
```yaml
- name: Strict type check
   run: |
     mypy --install-types --non-interactive --strict {foo.py,bar.py,foobar.py,...}
```

Any suggestions on how to format this, @anishathalye?

elisno: We can revisit this in a separate PR.

cleanlab/internal/token_classification_utils.py
```diff
@@ -73,64 +76,80 @@ def process_token(token: str, replace: list = [("#", "")]) -> str:
...
+        [0, 1, 1, 2, 2]  # ["O", "PER", "PER", "LOC", "LOC"]
+        >>> mapping([0, 0, 4, 4, 3, 4, 0, 2], maps)
+        [0, 0, 2, 2, 2, 2, 0, 1]  # ["O", "O", "LOC", "LOC", "LOC", "LOC", "O", "PER"]
     """
     f = lambda x: maps[x]
     return list(map(f, entities))
 
 
-def merge_probs(probs: np.ndarray, maps: list) -> np.ndarray:
+def merge_probs(probs: np.ndarray, maps: List[int]) -> np.ndarray:  # type: ignore[type-arg]
```
jwmueller, line 129: should we add description of this trick in mypy section of DEVELOPMENT.md?

jwmueller:
@anishathalye can you just give quick review of the type annotations to ensure best practices are being employed?  I'll be more confident in my evaluations of these once we have more files in codebase employing best type annotation practices :)

cleanlab/internal/token_classification_utils.py
```diff
@@ -29,25 +30,27 @@ def get_sentence(words: list) -> str:
 
 
 def filter_sentence(
-    sentences: list, condition: Optional[Callable] = None, return_mask: bool = True
-) -> Union[Tuple[list, list], list]:
+    sentences: List[str],
+    condition: Optional[Callable[[str], bool]] = None,
+    return_mask: bool = True,
+) -> Union[Tuple[List[str], List[bool]], List[str]]:
```
anishathalye, line 36: For external APIs, do we want to follow the convention that NumPy uses, using `typing.overload` (e.g., https://github.com/numpy/numpy/blob/54c52f13713f3d21795926ca4dbb27e16fada171/numpy/lib/arraysetops.pyi#L121)? I'd be on board with this plan for public APIs: seems like it'll be a common pattern in our library. If we do this, we should document this practice in the development guide, so we can standardize on it.

(See Ryan's message / this Slack thread: https://cleanlabinc.slack.com/archives/C03C42GDTR8/p1661824568503139?thread_ts=1661822614.989389&cid=C03C42GDTR8)

cleanlab/internal/token_classification_utils.py
```diff
@@ -73,64 +76,80 @@ def process_token(token: str, replace: list = [("#", "")]) -> str:
...
-        maps: list
+        maps: List[int]
             a list of mapped index, such that the probability of the token being in the i'th class is mapped to the
             `maps[i]` index. If `maps[i] == -1`, the i'th column of `probs` is ignored. If `np.any(maps == -1)`, the
             returned probability is re-normalized.
 
     Returns
     ---------
-        probs_merged: np.array
+        probs_merged: np.ndarray
```
anishathalye, line 145: Could we avoid duplicating types in annotations + docstrings? If we omit the types here, does Sphinx automatically fill them in based on the type annotations? If not, can we configure Sphinx in that way?

cleanlab/internal/token_classification_utils.py
anishathalye, line 129: The `type: ignore[type-arg]` isn't necessary with our current mypy configuration. And when we do switch to `--strict`, I think we'd want to figure out if we can globally ignore type arguments for `np.ndarray` or find some better solution, otherwise our code will be littered with `type: ignore[type-arg]`.

cleanlab/internal/token_classification_utils.py
```diff
@@ -73,64 +76,80 @@ def process_token(token: str, replace: list = [("#", "")]) -> str:
...
             np.array of shape `(N, K')`, where K' is the number of new classes. Probablities are merged and
             re-normalized if necessary.
 
     """
     old_classes = probs.shape[1]
-    probs_merged = np.zeros([len(probs), np.max(maps) + 1])
+    map_size = np.max(maps) + 1  # type: ignore
```
anishathalye, line 151: If you use `max(maps)` (the builtin max), its type is inferred as `int`, so you don't need the `type: ignore`. In any case, mypy is warning that the `type: ignore` is unused, and `map_size` is being inferred as an `Any`. You could explicitly type `map_size: int = ...` if you want, though I think `max(maps)` is a cleaner solution.

cleanlab/internal/token_classification_utils.py
```diff
@@ -1,21 +1,22 @@
+import re
 import string
 import numpy as np
 from termcolor import colored
-from typing import Union, Optional, Callable, Tuple
+from typing import Any, List, Union, Optional, Callable, Tuple
```
anishathalye, line 5: `Any` is unused

anishathalye (commented):
Overall looks good, left a couple comments. Only reviewed the static types, not the code changes.

Btw, the CI issue can be fixed by checking this single file with `--strict` mode plus [`--follow-imports=skip`](https://mypy.readthedocs.io/en/latest/running_mypy.html#follow-imports).

cleanlab/internal/token_classification_utils.py
anishathalye: Even for internal APIs: on the callee side, (untagged) unions are not a very useful type, and we'll need type assertions (see Ulyana's PR mentioned in that Slack thread); refactoring is an alternative to `typing.overload`.

elisno: return_mask flag will be removed and the function always return a tuple

cleanlab/internal/token_classification_utils.py
elisno: Removing ignore-comment.

cleanlab/internal/token_classification_utils.py
elisno: Removing hard-coded tag. Should be auto-tagged with sphinx-autodoc-typehints in a separate PR.

cleanlab/internal/token_classification_utils.py
elisno: Actually, np.max is untyped which didn't pass the `--strict` check. This is not an issue with the inference of `map_size`

```python
# From numpy/__init__.pyi
def max(a, axis=..., out=..., keepdims=..., initial=..., where=...): ...
```

elisno: However, the ignore-comment will be removed.

cleanlab/internal/token_classification_utils.py
elisno: 🔥

cleanlab/internal/token_classification_utils.py
```diff
@@ -29,42 +30,38 @@ def get_sentence(words: list) -> str:
...
 
     Parameters
     ----------
-        sentences: list
+        sentences: List[str]
             list of sentences
 
-        condition: Callable
+        condition: Optional[Callable[[str], bool]]
             sentence filtering condition
```
jwmueller, line 45: Save for future PR where you clean up the docstrings and their formatting (eg. capitalization), but each of these docstrings should have some example of what could possibly be passed in here for this argument, otherwise it remains too vague for users.

cleanlab/internal/token_classification_utils.py
```diff
@@ -29,42 +30,38 @@ def get_sentence(words: list) -> str:
...
     ---------
-        sentences: list
+        sentences: List[str]
             list of sentences filtered
 
-        mask: if `return_mask`, also returns a mask such that `mask[i] == True` if the i'th sentence is included in the
-        filtered sentence, otherwise `mask[i] == False`
+        mask: List[bool]
+            boolean mask such that `mask[i] == True` if the i'th sentence is included in the
+            filtered sentence, otherwise `mask[i] == False`
```
jwmueller, line 54: Save for future PR where you clean up the docstrings and their formatting (eg. capitalization), but each of these docstrings should have some example of what could possibly be passed in here for this argument, otherwise it remains too vague for users.

cleanlab/internal/token_classification_utils.py
```diff
@@ -73,64 +70,80 @@ def process_token(token: str, replace: list = [("#", "")]) -> str:
         token: str
             token which potentially contains special characters
 
-        replace: list
+        replace: List[Tuple[str, str]]
             list of tuples `(s1, s2)`, where all occurances of s1 are replaced by s2
```
jwmueller, line 74: Save for future PR where you clean up the docstrings and their formatting (eg. capitalization), but each of these docstrings should have some example of what could possibly be passed in here for this argument, otherwise it remains too vague for users.

jwmueller (approved):
Great stuff!  

Remaining TODO for future PR cleaning up the token-classification docstrings and their formatting (eg. capitalization):  each of these docstrings should have some example of what could possibly be passed in for some of the more complicated-looking arguments, otherwise they remain too vague for users.

PR #391, opened 2022-09-01
Fix typing for find_label_issues

elisno:
In relation to #384 where code introduced in cleanlab/token_classification/filter.py is reviewed/modified for release.

This PR only addresses typing of one function, `find_label_issues`.

### DIFF
--- a/cleanlab/token_classification/filter.py
+++ b/cleanlab/token_classification/filter.py
@@ -1,22 +1,24 @@
 import numpy as np
 from cleanlab.filter import find_label_issues as find_label_issues_main
-from typing import Optional, Union, Tuple
+from typing import List, Tuple
 
 
 def find_label_issues(
-    labels: list, pred_probs: list, return_indices_ranked_by: str = "self_confidence"
-) -> list:
+    labels: List[List[int]],
+    pred_probs: List[np.ndarray],  # type: ignore[type-arg]
+    return_indices_ranked_by: str = "self_confidence",
+) -> List[Tuple[int, int]]:
     """Returns issues identified by cleanlab
 
     This is a function to find the label issues for token classification datasets
 
     Parameters
     ----------
-    labels: list
+    labels: List[List[int]]
         noisy token labels in nested list format, such that `labels[i]` is a list of token labels of the i'th
         sentence. For datasets with `K` classes, each label must be in 0, 1, ..., K-1. All classes must be present.
 
-    pred_probs: list
+    pred_probs: List[np.ndarray]
         list of np.arrays, such that `pred_probs[i]` is the model-predicted probabilities for the tokens in
         the i'th sentence, and has shape `(N, K)`. Each row of the matrix corresponds to a token `t` and contains
         the model-predicted probabilities that `t` belongs to each possible class, for each of the K classes. The
@@ -28,20 +30,20 @@ def find_label_issues(
 
     Returns
     ----------
-    issues: list
+    issues: List[Tuple[int, int]]
         a list containing all potential issues identified by cleanlab, such that each element is a tuple (i, j), which
         corresponds to the j'th token of the i'th sentence.
     """
     labels_flatten = [l for label in labels for l in label]
     pred_probs_flatten = np.array([pred for pred_prob in pred_probs for pred in pred_prob])
 
-    issues = find_label_issues_main(
+    issues_main = find_label_issues_main(
         labels_flatten, pred_probs_flatten, return_indices_ranked_by=return_indices_ranked_by
-    )
+    )  # type: ignore[no-untyped-call]
 
     lengths = [len(label) for label in labels]
     mapping = [[(i, j) for j in range(length)] for i, length in enumerate(lengths)]
     mapping_flatten = [index for indicies in mapping for index in indicies]
 
-    issues = [mapping_flatten[issue] for issue in issues]
+    issues = [mapping_flatten[issue] for issue in issues_main]
     return issues
--- a/cleanlab/token_classification/filter.py
+++ b/cleanlab/token_classification/filter.py
@@ -4,8 +4,8 @@ from typing import List, Tuple
 
 
 def find_label_issues(
-    labels: List[List[int]],
-    pred_probs: List[np.ndarray],
+    labels: list,
+    pred_probs: list,
     return_indices_ranked_by: str = "self_confidence",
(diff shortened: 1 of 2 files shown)


cleanlab/token_classification/filter.py
```diff
@@ -1,22 +1,24 @@
...
 
     Parameters
     ----------
-    labels: list
+    labels: List[List[int]]
         noisy token labels in nested list format, such that `labels[i]` is a list of token labels of the i'th
         sentence. For datasets with `K` classes, each label must be in 0, 1, ..., K-1. All classes must be present.
 
-    pred_probs: list
+    pred_probs: List[np.ndarray]
```
anishathalye, line 21: Redundant with type annotations, can we auto-populate these in Sphinx instead?

cleanlab/token_classification/filter.py
```diff
@@ -1,22 +1,24 @@
...
 from cleanlab.filter import find_label_issues as find_label_issues_main
-from typing import Optional, Union, Tuple
+from typing import List, Tuple
 
 
 def find_label_issues(
-    labels: list, pred_probs: list, return_indices_ranked_by: str = "self_confidence"
-) -> list:
+    labels: List[List[int]],
+    pred_probs: List[np.ndarray],  # type: ignore[type-arg]
```
anishathalye, line 8: `type: ignore[type-arg]` is unnecessary for our current mypy configuration. And when we do switch to strict mode, I think we might want some other way of handling this, e.g., explicit `np.ndarray[Any]` everywhere we allow any type.

cleanlab/token_classification/filter.py
```diff
@@ -28,20 +30,20 @@ def find_label_issues(
...
         corresponds to the j'th token of the i'th sentence.
     """
     labels_flatten = [l for label in labels for l in label]
     pred_probs_flatten = np.array([pred for pred_prob in pred_probs for pred in pred_prob])
 
-    issues = find_label_issues_main(
+    issues_main = find_label_issues_main(
         labels_flatten, pred_probs_flatten, return_indices_ranked_by=return_indices_ranked_by
-    )
+    )  # type: ignore[no-untyped-call]
```
anishathalye, line 42: Would omit this as well, unnecessary with our current mypy configuration. I'd be worried that we'd need to review all the `type: ignore[...]` once we switch to strict mode, to make sure that we don't have any unnecessary ones left over.

cleanlab/token_classification/filter.py
elisno: Removing ignore-comment.

cleanlab/token_classification/filter.py
elisno: Removing hard-coded tag. We'll let sphinx-autodoc-typehints handle this in a separate PR.

cleanlab/token_classification/filter.py
elisno: Removed ignore-comment.

cleanlab/token_classification/filter.py
```diff
@@ -1,22 +1,24 @@
...
 from cleanlab.filter import find_label_issues as find_label_issues_main
-from typing import Optional, Union, Tuple
+from typing import List, Tuple
 
 
 def find_label_issues(
-    labels: list, pred_probs: list, return_indices_ranked_by: str = "self_confidence"
-) -> list:
+    labels: List[List[int]],
+    pred_probs: List[np.ndarray],
```
anishathalye, line 8: Are these types overly restrictive? Could the user, for example, pass in a 2d np.array of ints for `labels`, or a 3d np.array of floats for `pred_probs`?

anishathalye: Slightly misaligned with [`find_label_issues`](https://docs.cleanlab.ai/v2.0.0/cleanlab/filter.html#cleanlab.filter.find_label_issues) types, where we ask for a np.array for labels.

elisno: > Are these types overly restrictive? Could the user, for example, pass in a 2d np.array of ints for `labels`, or a 3d np.array of floats for `pred_probs`?

The only use case I can think of is if the user wants to truncate their inputs or add padding tokens to fit a batch of sentences/documents into a rectangular array.

I'm not sure if it was originally intended to support truncated/padded inputs. Thoughts on this, @ericwang1997?

> Slightly misaligned with [`find_label_issues`](https://docs.cleanlab.ai/v2.0.0/cleanlab/filter.html#cleanlab.filter.find_label_issues) types, where we ask for a np.array for labels.

Looks like both of these should use `LabelLike`.
It should be `List[LabelLike]` for token classification, but there are no tests that cover this.

jwmueller: I think the two supported types should be: `List[List[int]]`, `List[np.ndarray of ints]`
@elisno Can you verify if the second type works and if not, just convert each of the arrays inside the list to a List?  

@anishathalye This does not currently work for 2d array of labels because of the variable-length nature of sentences. And dealing with padding would be annoying IMO and too much unnecessary extra work. The user anyway likely has to process their raw predictions/labels from neural net format into cleanlab's format, so we might as well ask for the simplest and most intuitive format.

elisno: Both types work, but maybe it was a bit excessive of me to tracking the types of the nested sequences at this stage. Reverting these changes for now.

cleanlab/token_classification/filter.py
```diff
@@ -1,22 +1,24 @@
...
+    return_indices_ranked_by: str = "self_confidence",
+) -> List[Tuple[int, int]]:
     """Returns issues identified by cleanlab
 
     This is a function to find the label issues for token classification datasets
 
     Parameters
     ----------
-    labels: list
+    labels:
```
jwmueller, line 17: @anishathalye Is it preferable like this vs. the old way: `labels: list` ?

jwmueller (approved):
New unit test LGTM. Just have quick question for Anish regarding ideal docstring format, whether we prefer to have: 
```
labels:
  a list .... 
```   
VS
```
labels: list
  a list .... 
```

cleanlab/token_classification/filter.py
anishathalye: Yes, we'll generate the types based on Python annotations, so we don't have duplicates here.

anishathalye (approved):

PR #393, opened 2022-09-01
Cleanup token_classification.rank module

elisno:
More work related to #384, specifically for cleanlab/token_classification/rank.py.

## This PR adds:
- Unit tests and specialized edge cases for `softmin_sentence_score`.
- Make a test fixture for `get_label_quality_score`.

### Some notes

- Calling `get_label_quality_scores(..., sentence_score_method="min", ...)` should give similar results to calling `get_label_quality_scores(..., sentence_score_method="softmin", sentence_score_kwargs={"temperature": 0}`.
  - It might be useful when testing the "softmin" function for both extremes in `temperature`.
- `issues_from_scores` should be split into two separate functions for each return option.
  - I'd say it's outside the scope of this PR.

### DIFF
--- a/cleanlab/token_classification/rank.py
+++ b/cleanlab/token_classification/rank.py
@@ -125,7 +125,7 @@ def get_label_quality_scores(
 
 
 def issues_from_scores(
-    sentence_scores: np.ndarray, token_scores: list, threshold: float = 0.1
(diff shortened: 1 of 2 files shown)


jwmueller (approved):
Really great cleanup work, code is a lot better now!

PR #412, opened 2022-09-08
Add autodoc-typehints extension for sphinx

### LINKED ISSUE #398: CI: Suppress complex mypy typing in docs.cleanlab.ai
If docs.cleanlab.ai shows the mypy type annotations, the current signatures look a bit intimidating, eg. for: `cleanlab.internal.validation`
https://docs.cleanlab.ai/master/cleanlab/internal/validation.html 

Plan is to use https://github.com/tox-dev/sphinx-autodoc-typehints to auto-tag parameters in docstrings based on the function signatures.

Consider the docstring for `cleanlab.internal.validation.labels_to_array`:
https://docs.cleanlab.ai/master/cleanlab/internal/validation.html#cleanlab.internal.validation.labels_to_array

The docstring would not change as the tags have been hard-coded. But the signature at the top would read:
```
cleanlab.internal.validation.labels_to_array(y)
```
instead of
```
cleanlab.internal.validation.labels_to_array(y: Union[list, ndarray, Series, DataFrame, generic]) → ndarray
```

This plan requires ironing out details with unrolling/expanding custom types like `LabelLike`. 
It will take some effort to review all hard-coded types/tags in docstrings, because they won't be replaced with the auto-tag.

elisno:
With the extension: 
Function parameters are auto-tagged in docs, based on the function signatures.

Resolves #398

---

## Example
Consider this definition of `cleanlab.internal.labels_to_array`:

```python
def labels_to_array(y: Union[LabelLike, np.generic]) -> np.ndarray:
    """Converts different types of label objects to 1D numpy array and checks their validity.

    Parameters
    ----------
    y:
        Labels to convert to 1D numpy array. Can be a list, numpy array, pandas Series, or pandas DataFrame.

    Returns
    -------
    np.ndarray
        1D numpy array of labels.
    """
    ...
```

([`LabelLike` is defined in cleanlab.typing](https://github.com/cleanlab/cleanlab/blob/f6d9fceddec4c34bf1a50c90bb040a36879d66ed/cleanlab/typing.py#L5))

## Before PR
The documentation for this function would look like this:

before

## After PR

The function signature is much cleaner, and the parameters are typed instead:

after

Hard-coded types in the docstrings should be unaffected.


elisno:
Still have to iron out the proper formatting for examples like

```python
"""
def foo(x: int) -> str
...

Returns
-------
    Description of return value
"""
```
so that the extension correctly infers the return type (here it should be `str`).

docs/source/conf.py
```diff
@@ -79,7 +79,7 @@
 napoleon_use_admonition_for_references = False
 napoleon_use_ivar = False
 napoleon_use_param = True
-napoleon_use_rtype = True
+napoleon_use_rtype = False
```
elisno, line 82: This example code:
```python
# cleanlab/internal/token_classification_utils.py
def get_sentence(words: List[str]) -> str:
    """
    Get sentence formed by a list of words with minor processing for readability

    Parameters
    ----------
    words:
        list of word-level tokens

    Returns
    ----------
    sentence formed by list of word-level tokens

    Examples
    ----------
    >>> from cleanlab.internal.token_classification_utils import get_sentence
    >>> words = ["This", "is", "a", "sentence", "."]
    >>> get_sentence(words)
    'This is a sentence.'
    """
    ...
```

would output the following in the docs.

image

elisno: I don't see a clean way to handle the ` `` ` around the description of the return value.

jwmueller (approved):
LGTM. I'll wait on @anishathalye to also approve before merging.  After this is merged, my plan will be to update all existing docstrings to have a named return variable, such that they look good under the new system.

That will match what numpy docs are doing:   https://numpy.org/doc/stable/reference/routines.array-creation.html

anishathalye (approved):
LGTM!

PR #415, opened 2022-09-08
Add termcolor dependency

huiwengoh:
Add missing dependency `termcolor` used in [token_classification_utils.py ](https://github.com/cleanlab/cleanlab/blob/23067aa9c6d4cdf479cc997d3b38fbaf9b87ebf3/cleanlab/internal/token_classification_utils.py#L24)


jwmueller (approved):
Great find, still wondering how CI didn't catch this...

PR #423, opened 2022-09-09
Update argument given_labels -> labels in token classification module

### LINKED ISSUE #418: Token Classification: given_labels -> labels
Rename argument `given_labels` to just `labels` throughout the module to be more consistent with rest of package.

Example:  https://github.com/cleanlab/cleanlab/blob/master/cleanlab/token_classification/summary.py#L33

Need to correspondingly update tutorial notebook as well, possibly also example notebook but I don't think so.

elisno:
The only inconsistent use of the (keyword) argument `given_labels` instead of `labels` is in `cleanlab.token_classification.display_issues`.

This PR fixes this argument name and updates any calls to `display_issues` that would be affected by this.

Fixes #418

### DIFF
--- a/cleanlab/token_classification/summary.py
+++ b/cleanlab/token_classification/summary.py
@@ -9,7 +9,7 @@ def display_issues(
     given_words: List[List[str]],
     *,
     pred_probs: Optional[list] = None,
-    given_labels: Optional[list] = None,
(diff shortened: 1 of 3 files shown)


jwmueller (approved):
LGTM!

PR #425, opened 2022-09-09
Update DEVELOPMENT.md with Relative Linking

ulya-tkch:
Added instructions for relative linking between docs and tutorials


ulya-tkch:
No disclaimer that linking to another notebook only works if its in the same directory but I specified that linking is between tutorial notebooks. Feel free to change to explicit disclaimer instead

jwmueller (approved):

PR #426, opened 2022-09-09
Updated labels to allow array_like 

ulya-tkch:
labels param can be either np.ndarray or list now.

- [x] Allow pd.Series???

### DIFF
--- a/cleanlab/outlier.py
+++ b/cleanlab/outlier.py
@@ -29,7 +29,7 @@ from cleanlab.internal.label_quality_utils import (
     get_normalized_entropy,
 )
 from cleanlab.internal.validation import assert_valid_inputs, labels_to_array
-
+from cleanlab.typing import LabelLike
 
 class OutOfDistribution:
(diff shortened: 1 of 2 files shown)


cleanlab/outlier.py
```diff
@@ -504,6 +504,8 @@ def _get_ood_predictions_scores(
                     f"params['adjusted_pred_probs'] = False. "
                 )
             else:
+                if isinstance(labels, list):
+                    labels = np.array(labels)
```
jwmueller, line 508: You can instead use this function to convert to array: 
https://github.com/cleanlab/cleanlab/blob/master/cleanlab/internal/validation.py#L180
without type-checking the input. Will make it automatically work for pd.Series and list

jwmueller (approved):
Note I changed the type annotation of `labels` to be our custom `LabelLike` type. Even though it's hardly used in the current package, let's try to use this where it is appropriate in the future

Issue #433, opened 2022-09-13
Extend KerasWrapper to Functional API


jwmueller:
- [x] Add new class to experimental/keras.py called: KerasWrapperModel.
 * KerasWrapperModel should support arbitrary Keras Models (their Functional API) and be sklearn-compatible, and work with tensorflow Dataset objects.
 * Instantiation of KerasWrapperModel can be based on existing keras.Model object, or based on Callable to create such an object (only one these options is needed, whichever is easier).
 * Make sure cross-validation of KerasWrapperModel does not keep training the same set of weights over all folds, but rather re-initializes between folds.
 * Rename existing KerasWrapper -> KerasWrapperSequential.

- [x] Add new unit tests for KerasWrapperModel to: https://github.com/cleanlab/cleanlab/blob/master/tests/test_frameworks.py
- [x] Update the stuff outlined in: https://github.com/cleanlab/cleanlab/issues/372 and the HuggingfaceKerasClassifier to utilize  the new KerasWrapperModel. This can be our primary showcase of the new KerasWrapperModel.

PR #434, opened 2022-09-13
Extend KerasWrapper to Functional API

huiwengoh:
Add functionality for KerasWrapper to support Keras functional API (in addition to sequential API).

Partially addresses #433


tests/test_frameworks.py
```diff
@@ -153,13 +153,59 @@ def test_tensorflow(batch_size, shuffle_config, data=DATA, hidden_units=128):
...
+    preds_base = model.predict_proba(dataset_tf)
+
+    # Test CleanLearning performs well:
+    cl = CleanLearning(model)
+    cl.fit(dataset_tf, data["y"], clf_kwargs={"epochs": 10}, clf_final_kwargs={"epochs": 15})
+
+    preds = cl.predict(dataset_og_order)
+    err = np.sum(preds != data["y_og"]) / len(data["y_og"])
+    issue_indices = list(cl.label_issues_df[cl.label_issues_df["is_label_issue"]].index.values)
+    assert set(issue_indices).issubset(set(data["error_indices"]))
```
jwmueller, line 198: might this check start to fail if Keras releases new version with eg. different weight-initializations / optimizer (ie. how robust do you think it is)?  would it be more future proof if we instead check their intersection is nonempty or Jaccard similarity is at least `2 / len(data["error_indices"]` or something like that which doesn't require strict subset?

cleanlab/experimental/keras.py
```diff
@@ -32,28 +32,86 @@
...
+from typing import Callable
 
 
-class KerasWrapper:
+class KerasWrapperModel:
     """
-    KerasWrapper is instantiated in the same way as a ``tf.keras.models.Sequential`` object,
+    KerasWrapperModel takes in a callable function to instantiate a Keras model (using Keras functional API)
+    that is compatible with :py:meth:`CleanLearning.fit()<cleanlab.classification.CleanLearning.fit>`.
+    """
```
jwmueller, line 42:
```suggestion
    For using Keras sequential API, see `KerasWrapperSequential` class defined below.
    """
```

cleanlab/experimental/keras.py
```diff
@@ -32,28 +32,86 @@
...
 import tensorflow as tf
 import numpy as np
+from typing import Callable
 
 
-class KerasWrapper:
+class KerasWrapperModel:
     """
-    KerasWrapper is instantiated in the same way as a ``tf.keras.models.Sequential`` object,
+    KerasWrapperModel takes in a callable function to instantiate a Keras model (using Keras functional API)
```
jwmueller, line 40:
```suggestion
    KerasWrapperModel takes in a callable function to instantiate a Keras Model (using Keras functional API)
```

cleanlab/experimental/keras.py
```diff
@@ -32,28 +32,86 @@
...
+    that is compatible with :py:meth:`CleanLearning.fit()<cleanlab.classification.CleanLearning.fit>`.
+    """
+
```
jwmueller, line 43: Add a short docstring to describe some of the args below, especially the ones that are not standard Keras Model() init args. Also, a one-line real code example of what `model` might look like could be helpful

jwmueller (commented):
Overall looks great! Just suggested a few minor changes

tests/test_frameworks.py
huiwengoh: Updated to check for non-empty intersection

jwmueller (approved):
Great stuff!!

PR #435, opened 2022-09-13
Fixed relative linking for outlier example

ulya-tkch:
- renamed outlier example in cleanlab/examples
- fixed all relative linking

Ref cleanlab/examples#18


jwmueller (approved):
Approving for now, but I think we should consider:

Is putting the dataset name first actually a good idea?  I feel like that conveys very little info re what the example is actually about, so I actually prefer the current name (where the dataset name appears last, if at all).  Eg. what if in the future we have 10 examples all using cifar10 and all these examples start with cifar10_* prefix, it will be less easy to distinguish them and I don't think somebody would want to go through them all simply because they're interested in cifar10.

jwmueller (approved):
(moving comment to other PR)

ulya-tkch:
Closed due to larger renaming

PR #446, opened 2022-09-15
Ignore links to pradyunsg when validating HTML

elisno:


anishathalye (approved):
For future reference, this URL appears in the footer of the theme that we are using. We don't care if the link is broken.

Screen Shot 2022-09-15 at 3 02 22 PM

elisno:
Failed checks are unrelated. This can be merged.

jwmueller (approved):

PR #452, opened 2022-09-16
More improvements to token classification code and documentation

jwmueller:
More important code changes include:

- arg rename: `given_words` -> `tokens`
- setting some optional args to be keyword-only args.


elisno (approved):
LGTM!

Issue #440, opened 2022-09-14
docs: mis-rendered summary/details


anishathalye:
The collapsible summary/details in the notebooks look a bit broken. Looking at the HTML in the rendered result, looks like the `<summary>` tag is being replaced with a `<p>` tag. Also, I think we should standardize our summary text across the docs.

Some examples of broken behavior:

Image tutorial: says "Details" instead of "Below is the code used for plotting":

Collapsed:

<img width="825" alt="Screen Shot 2022-09-14 at 10 10 03 AM" src="https://user-images.githubusercontent.com/3526486/190177634-3b102b98-66bc-4092-a209-4d47f0be8700.png">

Expanded:

<img width="867" alt="Screen Shot 2022-09-14 at 10 10 27 AM" src="https://user-images.githubusercontent.com/3526486/190177670-384c2b46-8d37-4012-bd2b-4786f046d264.png">

Token classification tutorial: has the right text but an unnecessary newline:

<img width="866" alt="Screen Shot 2022-09-14 at 10 11 01 AM" src="https://user-images.githubusercontent.com/3526486/190177796-da98d2b7-1d32-4629-87d9-978f6cd83441.png">

I like the way we've formatted it on our [blog](https://cleanlab.ai/blog/cleanlab-vizzy/), with the text "(click to expand)" included there:

<img width="501" alt="Screen Shot 2022-09-14 at 10 11 50 AM" src="https://user-images.githubusercontent.com/3526486/190178086-7a72409d-9c20-4e64-accf-bf166a4501e5.png">

Another nit, we should have `cursor: pointer` for the `<summary>`.

Also, it could be helpful to delineate the block with a border-top and border-bottom, so after it's expanded, the reader knows where the details end.

jwmueller:
FYI if trying to show the “summary” field in [docs.cleanlab.ai](http://docs.cleanlab.ai/), you’ll need to update all the text in those fields. The current text “Below is …” is written with the assumption the summary is not being shown on [docs.cleanlab.ai](http://docs.cleanlab.ai/).  It would not make sense for docs.cleanlab.ai to say "Below is ..." with all the stuff below actually hidden.

Issue #355, opened 2022-08-23
Juypter notebook tutorials: Better support for code drop-downs


jwmueller:
Currently the drop-down cells are implemented via Markdown with special html tags. 
This means the code to be displayed in these cells is duplicated: 
* once to be displayed in the Markdown drop-down. 
* once to be actually executed in a hidden code cell of the Jupyter notebook.

Would be good to avoid this code duplication if a better solution can be found.

PR #456, opened 2022-09-16
Fix details disclosure elements in docs

anishathalye:
This patch makes the following fixes:

- Make summary text appear on the same line as the ">" arrow marker; prior to this patch, it was being pushed down to the next line because there were nested block elements inside the `<summary>` element
- Style details element to visually delineate collapsible content with dashed lines above and below the content
- Add syntax highlighting for code blocks in details disclosure elements (some were missing syntax highlighting)
- Hide a cell that was meant to be hidden in the audio tutorial

See https://github.com/cleanlab/cleanlab/issues/440 for some screenshots of the details disclosure elements prior to this patch.

Here's an example of what these elements look like now.

Collapsed:

Screen Shot 2022-09-16 at 2 58 13 PM

Expanded:

Screen Shot 2022-09-16 at 2 58 19 PM

The summary element looked ugly due to a block element being present inside the `<summary>`. This was present in the rendered docs even though it was not explicitly in the source. Even though our source (in a Markdown cell in a Jupyter notebook) looked like:

```html
<summary>Example text</summary>
```

The rendered docs had:

```html
<summary><p>Example text</p></summary>
```

The easy fix was to leave this as-is and fix the styling with CSS, making all elements inside the summary to be displayed as inline elements.

Where was the extra `<p>` tag coming from? Different Markdown parsers handle mixed Markdown/HTML differently. We are using nbsphinx, which does something especially weird [[1]]: converting from Markdown -> reStructuredText -> HTML.

For example, consider the following Markdown:

```markdown
<details><summary>Summary</summary>

Body.

</details>
```

nbsphinx was first converting it to this reST:

```rst
.. raw:: html

   <details>

.. raw:: html

   <summary>

Summary

.. raw:: html

   </summary>

Body.

.. raw:: html

   </details>
```

And then this reST was being converted to HTML. This is where the extra `<p>` was coming from.

One workaround for this behavior is to switch to raw HTML cells in the notebook, rather than Markdown cells, and write pure HTML there. This has the disadvantage that it's inconvenient when the content inside the `<details>` has extra markup: Markdown is very convenient for that (e.g., for styling text), and having the contents being parsed as Markdown also allows us to easily use syntax highlighting inside the body.

Another workaround is to use raw reST cells [[2]] and write the proper reST directly (avoiding the bad Markdown -> reST conversion). This preserves the ability to use convenient formatting (though with reST instead of Markdown) as well as syntax-highlighted code blocks. For example:

```rst
.. raw:: html

    <details><summary>Summary</summary>

.. code-block:: python

    def meaning_of_life():
        return 42

.. raw:: html

    </details>
```

The above snippet produces the expected HTML, with syntax-highlighted code in the body, and without the extra `<p>` inside the `<summary>`.

Rather than either of these workarounds, the most convenient one for developers is to fix the issue using CSS, so that the Markdown cells in the tutorials can be left as-is and no special care is required when writing details disclosure elements, which is what this patch does.

Fixes https://github.com/cleanlab/cleanlab/issues/440.

[1]: https://github.com/spatialaudio/nbsphinx/blob/fc79fd1ac5ab6df225b256ef4ca203c58c5ad92b/src/nbsphinx.py#L1309
[2]: https://nbsphinx.readthedocs.io/en/0.8.9/raw-cells.html#reST

### DIFF
--- a/docs/source/_static/css/custom.css
+++ b/docs/source/_static/css/custom.css
@@ -1,0 +1,21 @@
+details {
+    margin-bottom: 0.75rem;
+    margin-top: 0.5rem;
+}
+
+details summary {
+    cursor: pointer;
+}
+
+details summary > * {
+    display: inline;
+}
+
+details[open] summary {
+    padding-bottom: 0.75rem;
+    border-bottom: 2px dashed #ccc;
+}
+
+details[open] {
+    border-bottom: 2px dashed #ccc;
+}
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -230,6 +230,8 @@ html_theme_options = {
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ["_static"]
 
+html_css_files = ["css/custom.css"]
+
 html_sidebars = {
     "**": [
         "brand.html",
--- a/docs/source/tutorials/audio.ipynb
+++ b/docs/source/tutorials/audio.ipynb
@@ -246,10 +246,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "<details><summary>Below is the code used to make audio files playable.</summary>\n",
+    "<details><summary>See the implementation of `display_example` **(click to expand)**</summary>\n",
     "\n",
-    "```\n",
-    "\n",
+    "```python\n",
     "# Note: This pulldown content is for docs.cleanlab.ai, if running on local Jupyter or Colab, please ignore it.\n",
     "\n",
     "import tensorflow_io as tfio\n",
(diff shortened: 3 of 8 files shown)


anishathalye:
(please preserve my commit message when merging this, don't squash)

jwmueller (approved):
Amazing stuff!  Your method of fixing it seems very good!

I'm all ears if you have thoughts in the future how we can avoid writing the actual code itself twice, in markdown and in hidden code cell :) 

https://github.com/cleanlab/cleanlab/issues/355

anishathalye:
Sounds good, will look into it as a low-priority thing, post 2.1 release.

PR #458, opened 2022-09-16
Added find_top_outliers functionality

ulya-tkch:
code + test + docstring


ulya-tkch:
I think renaming method from `find_top_outliers` to `find_outliers` actually makes sense. Thoughts?

ulya-tkch:
Solved by #459

PR #463, opened 2022-09-17
Add helper function to reduce code duplication

huiwengoh:
- Added helper function to check for missing classes in consensus labels to eliminate code duplication
- Moved `convert_long_to_wide_dataset` higher in the file to group public functions at the top


jwmueller (approved):
Thanks for the quick PR!

PR #471, opened 2022-09-19
multi label fix

aditya1503:


cleanlab/filter.py
```diff
@@ -398,6 +412,38 @@ class 0, 1, ..., K-1.
...
+    frac_noise=1.0,
+    num_to_remove_per_class=None,
+    min_examples_per_class=1,
+    confident_joint=None,
+    n_jobs=None,
+    verbose=False,
+):
+    ranked_label_issues_list = []
+    for i in range(0, len(pred_probs)):
+        y_labels = labels[i]
```
aditya1503, line 431: make it binary 0/1 manually here

cleanlab/filter.py
```diff
@@ -398,6 +412,38 @@ class 0, 1, ..., K-1.
...
+        pred_probabilitites = pred_probs[i]
```
aditya1503, line 432: Pred_probs, 1-pred_probs

cleanlab/filter.py
```diff
@@ -398,6 +412,38 @@ class 0, 1, ..., K-1.
...
+        if confident_joint is None:
+            conf = None
+        else:
+            conf = confident_joint[i]
+        ranks = find_label_issues(
```
aditya1503, line 437: rename to index_issues

cleanlab/filter.py
```diff
@@ -398,6 +412,38 @@ class 0, 1, ..., K-1.
...
+            y_labels.astype(np.int32),
+            pred_probabilitites,
+            return_indices_ranked_by="self_confidence",
```
aditya1503, line 440: keep as original
if it's None, and not None
if it's None, it returns a different an array of all true and false values (union1D to np.or ) , read code

cleanlab/filter.py
```diff
@@ -398,6 +411,80 @@ class 0, 1, ..., K-1.
...
+    return_indices_ranked_by=None,
+    rank_by_kwargs={},
+    filter_by="prune_by_noise_rate",
+    frac_noise=1.0,
+    num_to_remove_per_class=None,
+    min_examples_per_class=1,
+    confident_joint=None,
+    n_jobs=None,
+    verbose=False,
+):
```
jwmueller, line 426: note we are using type annotations in the package going forward, so all function signatures  should declare type of each arg as well as the return type. See for example:  https://github.com/cleanlab/cleanlab/blob/master/cleanlab/token_classification/filter.py

(unfortunately we haven't updated every file to comply with the type annotations yet, but all files in token_classification are compliant and serve as good references)

cleanlab/filter.py
```diff
@@ -398,6 +411,80 @@ class 0, 1, ..., K-1.
...
+                filter_by=filter_by,
+                multi_label=False,
+                frac_noise=frac_noise,
+                num_to_remove_per_class=num_to_remove_per_class,
+                min_examples_per_class=min_examples_per_class,
+                confident_joint=conf,
+                n_jobs=n_jobs,
+                verbose=verbose,
+            )
+            label_issues_list.append(index_issues)
```
jwmueller, line 484: there is a lot of code duplication here that I think could be reduced if you only had a single one of these for loops over `class_num` but did two separate checks `if return_indices_ranked_by is None`. One before the for loop to decide what kind of object to initialize for storing for loop results, and one inside the for loop for updating that object with the output of `find_label_issues`. Eg:

``` 
if return_indices_ranked_by is None:
    bissues = np.zeros(y_one.shape).astype(bool)
else:
    label_issues_list = []

for class_num in ...
    ...
    binary_label_issues = find_label_issues(...)
    if return_indices_ranked_by is None:
        bissues[:, class_num] = binary_label_issues
    else:
        label_issues_list.append(binary_label_issues)

if return_indices_ranked_by is None:
    return bissues  # Note I'm just copying your current code here but it seems wrong, should be row-wise Or operation applied here right? 
else:
    return reduce(np.union1d, label_issues_list).astype(np.int32)
```

cleanlab/filter.py
```diff
@@ -398,6 +411,80 @@ class 0, 1, ..., K-1.
...
+                rank_by_kwargs=rank_by_kwargs,
+                filter_by=filter_by,
+                multi_label=False,
+                num_to_remove_per_class=num_to_remove_per_class,
+                min_examples_per_class=min_examples_per_class,
+                confident_joint=conf,
+                n_jobs=n_jobs,
+                verbose=verbose,
+            )
+        return bissues
```
jwmueller, line 458: should be row-wise Or operation applied here right?

cleanlab/filter.py
```diff
@@ -398,6 +411,80 @@ class 0, 1, ..., K-1.
...
+    filter_by="prune_by_noise_rate",
+    frac_noise=1.0,
+    num_to_remove_per_class=None,
+    min_examples_per_class=1,
+    confident_joint=None,
+    n_jobs=None,
+    verbose=False,
+):
+    """
+    NOT
```
jwmueller, line 428: you'll want to add documentation here, especially because some of the args now have particular shapes they require

cleanlab/filter.py
```diff
@@ -235,7 +236,19 @@ class 0, 1, ..., K-1.
 
     # Number of examples in each class of labels
     if multi_label:
-        label_counts = value_counts([i for lst in labels for i in lst])
+        return _find_label_issues_multilabel(
```
jwmueller, line 239: To remain backwards compatible, we also have to additionally support the current case where `multi_label=True` and `confident_joint` is k x k matrix. My suggestion is to explicitly check for this shape of `confident_joint` inside `_find_label_issues_multilabel()` and if it is detected then:

-  set `confident_joint = None` internally at the start of the function (so it is completely ignored subsequently).
- raise Warning that says: "The new recommended format for `confident_joint` in multi_label settings is .... (as output by `compute_confident_joint(...,multi_label=True)`).  Your k x k `confident_joint` in the old format is being ignored.

cleanlab/filter.py
```diff
@@ -398,6 +411,80 @@ class 0, 1, ..., K-1.
...
+    verbose=False,
+):
+    """
+    NOT
+    """
+    num_classes = pred_probs.shape[1]
+    y_one = np.zeros((len(labels), num_classes))
+    for class_num in range(0, len(labels)):
+        for j in range(0, len(labels[class_num])):
+            y_one[class_num][j] = 1
```
jwmueller, line 434: seems like this should be a numpy operation?

jwmueller:
Closing in favor of: https://github.com/cleanlab/cleanlab/pull/483

PR #475, opened 2022-09-20
updated label_quality_utils.py and rebuilt the doc

ethanotran:
Updated label_quality_utils.html after PR: 
Screenshot (687)

PR #478, opened 2022-09-20
Fixed label_quality_utils.py and rebuilt the html doc

ethanotran:
Label_quality_utils.html after PR:
Screenshot (688)


jwmueller:
Why did you open another PR when you already have opened: https://github.com/cleanlab/cleanlab/pull/475/

I have already approved that one, so there is no more you need to do for this file.

PR #483, opened 2022-09-23
supporting multilabel via one-vs-rest reductions

aditya1503:
Adding Multi-Label fix for find label issues, 
This method now solves it using one-vs-rest reductions

```
pred_probs = np.array(
         [
             [0.9, 0.1, 0.0, 0.4],
             [0.7, 0.8, 0.2, 0.3],
             [0.9, 0.8, 0.4, 0.2],
             [0.1, 0.1, 0.8, 0.3],
             [0.4, 0.5, 0.1, 0.1],
         ]
     )
labels = [[0], [0, 1], [0, 1], [2], [0, 2, 3]]
noise_idx = filter.find_label_issues(
         labels=labels,
         pred_probs=pred_probs,
         multi_label=True,
     )
```

### DIFF
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -176,7 +176,7 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) -> 
       The multi-label setting supports classification tasks where an example has 1 or more labels.
       Example of a multi-labeled `labels` input: ``[[0,1], [1], [0,2], [0,1,2], [0], [1], ...]``.
       The major difference in how this is calibrated versus single-label is that
-      confident/calibrated joint arrays have shape ``(K, 2, 2)`` formatted in a one-vs-rest fashion such that they contain a 2x2 matrix for each class that counts examples which are correctly/incorrectly labeled as belonging to that class. 
+      confident/calibrated joint arrays have shape ``(K, 2, 2)`` formatted in a one-vs-rest fashion such that they contain a 2x2 matrix for each class that counts examples which are correctly/incorrectly labeled as belonging to that class.
       After calibration, the entries in each class-specific 2x2 matrix will sum to the number of examples.
 
     Returns
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -338,7 +338,7 @@ def _estimate_joint_multilabel(labels, pred_probs, *, confident_joint=None) -> n
       class 0, 1, ..., K-1. They need not sum to 1.0
 
     confident_joint : np.ndarray, optional
-       An array of shape ``(num_classes,2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
+       An array of shape ``(K, 2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
        estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
        Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
        The `confident_joint` can be computed using :py:func:`count.compute_confident_joint <cleanlab.count.compute_confident_joint>`.
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -160,6 +160,7 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) -> 
       Entry ``(j, k)`` in the matrix is the number of examples confidently counted into the pair of ``(noisy label=j, true label=k)`` classes.
       The `confident_joint` can be computed using :py:func:`count.compute_confident_joint <cleanlab.count.compute_confident_joint>`.
       If not provided, it is computed from the given (noisy) `labels` and `pred_probs`.
+      If multi_label is True, then the confident should be an array of shape ``(K, 2, 2)``.
 
     labels : np.ndarray
       A discrete vector of noisy labels, i.e. some labels may be erroneous.
@@ -228,7 +229,12 @@ def _calibrate_confident_joint_multilabel(confident_joint: np.ndarray, labels: l
     calibrated_cj : np.ndarray
       An array of shape ``(K, 2, 2)`` of type float representing a valid
       estimate of the joint *counts* of noisy and true labels in a one-vs-rest setting."""
-    y_one = int2onehot(labels)
+    try:
+        y_one = int2onehot(labels)
+    except TypeError:
+        raise ValueError(
+            "wrong format for labels, should be a list of list[indices], please check the documentation in find_label_issues for further information"
+        )
     num_classes = len(confident_joint)
     calibrate_confident_joint_list: np.ndarray = np.ndarray(
         shape=(num_classes, 2, 2), dtype=np.int64
@@ -273,6 +279,7 @@ def estimate_joint(labels, pred_probs, *, confident_joint=None, multi_label=Fals
       Entry ``(j, k)`` in the matrix is the number of examples confidently counted into the pair of ``(noisy label=j, true label=k)`` classes.
       The `confident_joint` can be computed using :py:func:`count.compute_confident_joint <cleanlab.count.compute_confident_joint>`.
       If not provided, it is computed from the given (noisy) `labels` and `pred_probs`.
+      If multi_label is True, then the confident should be an array of shape ``(K, 2, 2)``.
 
     multi_label : bool, optional
       If ``True``, labels should be an iterable (e.g. list) of iterables, containing a
@@ -285,6 +292,11 @@ def estimate_joint(labels, pred_probs, *, confident_joint=None, multi_label=Fals
     confident_joint_distribution : np.ndarray
       An array of shape ``(K, K)`` representing an
       estimate of the true joint distribution of noisy and true labels.
+      If multi_label is True,
+      An array of shape ``(num_classes,2, 2)`` representing an
+      estimate of the true joint distribution of noisy and true labels,
+      Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
+
     """
 
     if confident_joint is None:
@@ -306,9 +318,46 @@ def estimate_joint(labels, pred_probs, *, confident_joint=None, multi_label=Fals
         return calibrated_cj / float(np.sum(calibrated_cj))
 
 
-def _estimate_joint_multilabel(labels, pred_probs, *, confident_joint) -> np.ndarray:
+def _estimate_joint_multilabel(labels, pred_probs, *, confident_joint=None) -> np.ndarray:
+    """Parameters
+     ----------
+     labels : np.ndarray
+       An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
+       Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
+       All the classes (0, 1, ..., and K-1) MUST be present in ``labels``, such that:
+       ``len(set(labels)) == pred_probs.shape[1]`` for standard multi-class classification with single-labeled data (e.g. ``labels =  [1,0,2,1,1,0...]``).
+       For multi-label classification where each example can belong to multiple classes(e.g. ``labels = [[1,2],[1],[0],..]``),
+       your labels should instead satisfy: ``len(set(k for l in labels for k in l)) == pred_probs.shape[1])``.
+
+     pred_probs : np.ndarray
+       An array of shape ``(N, K)`` of model-predicted probabilities,
+       ``P(label=k|x)``. Each row of this matrix corresponds
+       to an example `x` and contains the model-predicted probabilities that
+       `x` belongs to each possible class, for each of the K classes. The
+       columns must be ordered such that these probabilities correspond to
+       class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
+       higher) fold cross-validation.
+
+    confident_joint : np.ndarray, optional
+       An array of shape ``(num_classes,2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
+       estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
+       Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
+       The `confident_joint` can be computed using :py:func:`count.compute_confident_joint <cleanlab.count.compute_confident_joint>`.
+       If not provided, it is computed from the given (noisy) `labels` and `pred_probs`.
+
+     Returns
+     -------
+     confident_joint_distribution : np.ndarray
+       An array of shape ``(K, 2, 2)`` representing an
+       estimate of the true joint distribution of noisy and true labels in a multi-label setting.
+    """
     num_classes = get_num_classes(labels=labels, pred_probs=pred_probs)
-    y_one = int2onehot(labels)
+    try:
+        y_one = int2onehot(labels)
+    except TypeError:
+        raise ValueError(
+            "wrong format for labels, should be a list of list[indices], please check the documentation in find_label_issues for further information"
+        )
     if confident_joint is None:
         calibrated_cj = compute_confident_joint(
             labels,
@@ -407,6 +456,10 @@ def compute_confident_joint(
     confident_joint_counts : np.ndarray
       An array of shape ``(K, K)`` representing counts of examples
       for which we are confident about their given and true label.
+      If multi_label is True,
+      An array of shape ``(num_classes,2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
+      estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
+      Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
 
 
       Note
@@ -1331,9 +1384,34 @@ def _get_confident_thresholds_multilabel(
     labels: np.ndarray,
     pred_probs: np.ndarray,
 ):
+    """Returns expected (average) "self-confidence" for each class.
+
+    The confident class threshold for a class j is the expected (average) "self-confidence" for class j in a one-vs-rest setting.
+
+    Parameters
+    ----------
+    labels: list
+       List of noisy labels for multi-label classification where each example can belong to multiple classes (e.g. ``labels = [[1,2],[1],[0],..]`` indicates the first example in dataset belongs to both class 1 and class 2.
+       For multi-label settings, your `labels` should instead satisfy: ``len(set(k for l in labels for k in l)) == pred_probs.shape[1])``.
+
+    pred_probs : np.ndarray
+       Predicted-probabilities in the same format expected by the :py:func:`get_confident_thresholds <cleanlab.count.get_confident_thresholds>` function.
+
+
+
+    Returns
+    -------
+    confident_thresholds : np.ndarray
+      An array of shape ``(K, 2)`` where K is the number of classes.
+    """
     num_classes = get_num_classes(labels=labels, pred_probs=pred_probs)
+    try:
+        y_one = int2onehot(labels)
+    except TypeError:
+        raise ValueError(
+            "wrong format for labels, should be a list of list[indices], please check the documentation in find_label_issues for further information"
+        )
     confident_thresholds: np.ndarray = np.ndarray((num_classes, 2))
-    y_one = int2onehot(labels)
     for class_num in range(num_classes):
         pred_probabilitites = _binarize_pred_probs_slice(pred_probs, class_num)
         confident_thresholds[class_num] = get_confident_thresholds(
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -439,16 +439,18 @@ def compute_confident_joint(
     confident_joint_counts : np.ndarray
       An array of shape ``(K, K)`` representing counts of examples
       for which we are confident about their given and true label (if `multi_label` is False).
-      If multi_label is True,
-      An array of shape ``(K, 2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
-      estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
-      Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
+      If `multi_label` is True,
+      this array instead has shape ``(K, 2, 2)`` representing a one-vs-rest format for the  confident joint, where for each class `c`:
+      Entry ``(c, 0, 0)`` in this one-vs-rest array is the number of examples whose noisy label contains `c` confidently identified as truly belonging to class `c` as well.
+      Entry ``(c, 1, 0)`` in this one-vs-rest array is the number of examples whose noisy label contains `c` confidently identified as not actually belonging to class `c`.
+      Entry ``(c, 0, 1)`` in this one-vs-rest array is the number of examples whose noisy label does not contain `c` confidently identified as truly belonging to class `c`.
+      Entry ``(c, 1, 1)`` in this one-vs-rest array is the number of examples whose noisy label does not contain `c` confidently identified as actually not belonging to class `c` as well.
 
 
       Note
       ----
-      if `return_indices_of_off_diagonals` is set as True, this function instead returns a tuple `(confident_joint, indices_off_diagonal)`
-      where `indices_off_diagonal` is an array and each array contains the indices of examples counted in off-diagonals of confident joint.
+      If `return_indices_of_off_diagonals` is set as True, this function instead returns a tuple `(confident_joint, indices_off_diagonal)`
+      where `indices_off_diagonal` is a list of arrays and each array contains the indices of examples counted in off-diagonals of confident joint.
 
     Note
     ----
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -438,7 +438,7 @@ def compute_confident_joint(
     -------
     confident_joint_counts : np.ndarray
       An array of shape ``(K, K)`` representing counts of examples
-      for which we are confident about their given and true label.
+      for which we are confident about their given and true label (if `multi_label` is False).
       If multi_label is True,
       An array of shape ``(K, 2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
       estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -1333,10 +1333,11 @@ def _get_confident_thresholds_multilabel(
     pred_probs: np.ndarray,
 ):
     num_classes = get_num_classes(labels=labels, pred_probs=pred_probs)
-    confident_thresholds = np.array((num_classes, pred_probs.shape[1]))
+    confident_thresholds = np.ndarray((num_classes, 2))
     y_one = int2onehot(labels)
     for class_num in range(num_classes):
         pred_probabilitites = _binarize_pred_probs_slice(pred_probs, class_num)
         confident_thresholds[class_num] = get_confident_thresholds(
             pred_probs=pred_probabilitites, labels=y_one[:, class_num]
         )
+    return confident_thresholds
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -332,7 +332,7 @@ def _estimate_joint_multilabel(labels, pred_probs, *, confident_joint=None) -> n
      -------
      confident_joint_distribution : np.ndarray
        An array of shape ``(K, 2, 2)`` representing an
-       estimate of the true joint distribution of noisy and true labels in a multi-label setting.
+       estimate of the true joint distribution of noisy and true labels for each class, in a one-vs-rest format employed for multi-label settings.
     """
     num_classes = get_num_classes(labels=labels, pred_probs=pred_probs)
     try:
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -576,8 +576,8 @@ def _compute_confident_joint_multi_label(
     Parameters
     ----------
     labels : list of list/iterable (length N)
-        These are multiclass labels. Each list in the list contains
-        all the labels for that example. This method will fail if labels
+        List of noisy labels for multi-label classification. Each list in the list contains
+        all the class assignments for that example. This method will fail if labels
         is not a list of lists (or a list of np.ndarrays or iterable).
 
     pred_probs : np.ndarray (shape (N, K))
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -321,28 +321,14 @@ def estimate_joint(labels, pred_probs, *, confident_joint=None, multi_label=Fals
 def _estimate_joint_multilabel(labels, pred_probs, *, confident_joint=None) -> np.ndarray:
     """Parameters
      ----------
-     labels : np.ndarray
-       An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
-       Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
-       All the classes (0, 1, ..., and K-1) MUST be present in ``labels``, such that:
-       ``len(set(labels)) == pred_probs.shape[1]`` for standard multi-class classification with single-labeled data (e.g. ``labels =  [1,0,2,1,1,0...]``).
-       For multi-label classification where each example can belong to multiple classes(e.g. ``labels = [[1,2],[1],[0],..]``),
-       your labels should instead satisfy: ``len(set(k for l in labels for k in l)) == pred_probs.shape[1])``.
+     labels : list
+      Refer to documentation for this argument in filter.find_label_issues() for details.
 
      pred_probs : np.ndarray
-       An array of shape ``(N, K)`` of model-predicted probabilities,
-      ``P(label=k|x)``. Each row of this matrix corresponds
-      to an example `x` and contains the model-predicted probabilities that
-      `x` belongs to each possible class, for each of the K classes. The
-      columns must be ordered such that these probabilities correspond to
-      class 0, 1, ..., K-1. They need not sum to 1.0
+       Refer to documentation for this argument in count.estimate_joint() for details.
 
     confident_joint : np.ndarray, optional
-       An array of shape ``(K, 2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
-       estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
-       Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
-       The `confident_joint` can be computed using :py:func:`count.compute_confident_joint <cleanlab.count.compute_confident_joint>`.
-       If not provided, it is computed from the given (noisy) `labels` and `pred_probs`.
+       Refer to documentation for this argument in filter.find_label_issues() with multi_label=True for details.
 
      Returns
      -------
@@ -690,7 +676,12 @@ def estimate_latent(
     Returns
     ------
     tuple
-      A tuple containing (py, noise_matrix, inv_noise_matrix)."""
+      A tuple containing (py, noise_matrix, inv_noise_matrix).
+
+    Note
+    ----
+    Multi-label classification is not supported in this method.
+    """
 
     # 'ps' is p(labels=k)
     ps = value_counts(labels) / float(len(labels))
@@ -795,7 +786,12 @@ def estimate_py_and_noise_matrices_from_probabilities(
     Returns
     ------
     estimates : tuple
-        A tuple of arrays: (`py`, `noise_matrix`, `inverse_noise_matrix`, `confident_joint`)."""
+        A tuple of arrays: (`py`, `noise_matrix`, `inverse_noise_matrix`, `confident_joint`).
+
+    Note
+    ----
+    Multi-label classification is not supported in this method.
+    """
 
     confident_joint = compute_confident_joint(
         labels=labels,
@@ -1390,13 +1386,9 @@ def _get_confident_thresholds_multilabel(
     Parameters
     ----------
     labels: list
-       List of noisy labels for multi-label classification where each example can belong to multiple classes (e.g. ``labels = [[1,2],[1],[0],..]`` indicates the first example in dataset belongs to both class 1 and class 2.
-       For multi-label settings, your `labels` should instead satisfy: ``len(set(k for l in labels for k in l)) == pred_probs.shape[1])``.
-
+       Refer to documentation for this argument in count.calibrate_confident_joint() with multi_label=True for details.
     pred_probs : np.ndarray
        Predicted-probabilities in the same format expected by the :py:func:`get_confident_thresholds <cleanlab.count.get_confident_thresholds>` function.
-
-
 
     Returns
     -------
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -175,9 +175,8 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) -> 
       The multi-label setting supports classification tasks where an example has 1 or more labels.
       Example of a multi-labeled `labels` input: ``[[0,1], [1], [0,2], [0,1,2], [0], [1], ...]``.
       The major difference in how this is calibrated versus single-label is that
-      the total number of errors considered is based on the number of labels,
-      not the number of examples. So, the calibrated `confident_joint` will sum
-      to the number of total labels.
+      the calibrated joint is calculated in a one-vs-rest setting, and will return an array of shape ``(K, 2, 2)``
+      each entry k in ``(k,i,j)`` will sum to the number of total labels.
 
     Returns
     -------
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -297,7 +297,27 @@ def estimate_joint(labels, pred_probs, *, confident_joint=None, multi_label=Fals
         calibrated_cj = calibrate_confident_joint(confident_joint, labels, multi_label=multi_label)
 
     assert isinstance(calibrated_cj, np.ndarray)
-    return calibrated_cj / float(np.sum(calibrated_cj))
+    if multi_label:
+        return _estimate_joint_multilabel(
+            labels=labels, pred_probs=pred_probs, confident_joint=confident_joint
+        )
+    else:
+        return calibrated_cj / float(np.sum(calibrated_cj))
+
+
+def _estimate_joint_multilabel(labels, pred_probs, *, confident_joint) -> np.ndarray:
+    num_classes = get_num_classes(labels=labels, pred_probs=pred_probs)
+    y_one = int2onehot(labels)
+    calibrated_cf = np.ndarray((num_classes, 2, 2))
+    for class_num in range(num_classes):
+        pred_probabilitites = _binarize_pred_probs_slice(pred_probs, class_num)
+        calibrated_cf[class_num] = estimate_joint(
+            labels=y_one[:, class_num],
+            pred_probs=pred_probabilitites,
+            confident_joint=confident_joint[class_num],
+        )
+
+    return calibrated_cf
 
 
 def compute_confident_joint(
@@ -378,9 +398,12 @@ def compute_confident_joint(
       An array of shape ``(K, K)`` representing counts of examples
       for which we are confident about their given and true label.
 
-    Note: if `return_indices_of_off_diagonals` is set as True, this function instead returns a tuple `(confident_joint, indices_off_diagonal)`
-    where `indices_off_diagonal` is an array and each array contains the indices of examples counted in off-diagonals of confident joint.
-    
+
+      Note
+      ----
+      if `return_indices_of_off_diagonals` is set as True, this function instead returns a tuple `(confident_joint, indices_off_diagonal)`
+      where `indices_off_diagonal` is an array and each array contains the indices of examples counted in off-diagonals of confident joint.
+
     Note
     ----
 
@@ -1286,14 +1309,23 @@ def get_confident_thresholds(
         get_num_classes(labels=labels, pred_probs=pred_probs, multi_label=multi_label)
     )
     if multi_label:
-        # Compute thresholds = p(label=k | k in set of given labels)
-        k_in_l = np.array([[k in lst for lst in labels] for k in unique_classes])
-        # The avg probability of class given that the label is represented.
-        confident_thresholds = np.array(
-            [np.mean(pred_probs[:, k][k_in_l[k]]) for k in unique_classes]
-        )
+        return _get_confident_thresholds_multilabel(labels=labels, pred_probs=pred_probs)
     else:
         confident_thresholds = np.array(
             [np.mean(pred_probs[:, k][labels == k]) for k in unique_classes]
         )
     return confident_thresholds
+
+
+def _get_confident_thresholds_multilabel(
+    labels: np.ndarray,
+    pred_probs: np.ndarray,
+):
+    num_classes = get_num_classes(labels=labels, pred_probs=pred_probs)
+    confident_thresholds = np.array((num_classes, pred_probs.shape[1]))
+    y_one = int2onehot(labels)
+    for class_num in range(num_classes):
+        pred_probabilitites = _binarize_pred_probs_slice(pred_probs, class_num)
+        confident_thresholds[class_num] = get_confident_thresholds(
+            pred_probs=pred_probabilitites, labels=y_one[:, class_num]
+        )
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -476,7 +476,7 @@ def _compute_confident_joint_multi_label(
     thresholds=None,
     calibrate=True,
     return_indices_of_off_diagonals=False,
-) -> Union[np.ndarray, Tuple[np.ndarray, list]]:  # type: ignore
+) -> Union[np.ndarray, Tuple[np.ndarray, list]]:
     """Computes the confident joint for multi_labeled data. Thus,
     input `labels` is a list of lists (or list of iterable).
     This is intended as a helper function. You should probably
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -1384,7 +1384,7 @@ def _get_confident_thresholds_multilabel(
     Parameters
     ----------
     labels: list
-       Refer to documentation for this argument in count.calibrate_confident_joint() with multi_label=True for details.
+       Refer to documentation for this argument in ``count.calibrate_confident_joint()`` with `multi_label=True` for details.
     pred_probs : np.ndarray
        Predicted-probabilities in the same format expected by the :py:func:`get_confident_thresholds <cleanlab.count.get_confident_thresholds>` function.
 
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -377,8 +377,6 @@ def compute_confident_joint(
     confident_joint_counts : np.ndarray
       An array of shape ``(K, K)`` representing counts of examples
       for which we are confident about their given and true label.
-      If `return_indices_of_off_diagonals` is ``True``, `confident_joint_counts` is the first element of returned tuple
-      and second element is another array of indices counted in off-diagonals of confident joint.
 
     Note: if `return_indices_of_off_diagonals` is set as True, this function instead returns a tuple `(confident_joint, indices_off_diagonal)`
     where `indices_off_diagonal` is an array and each array contains the indices of examples counted in off-diagonals of confident joint.
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -182,6 +182,9 @@ def calibrate_confident_joint(
     calibrated_cj : np.ndarray
       An array of shape ``(K, K)`` of type float representing a valid
       estimate of the joint *counts* of noisy and true labels.
+      If multi_label is true, it returns an array of shape ``(K, 2, 2)``,
+      where entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
+
     """
 
     if multi_label:
@@ -200,6 +203,30 @@ def calibrate_confident_joint(
 def _calibrate_confident_joint_multilabel(
     confident_joint: np.ndarray, labels: list
 ) -> List[np.ndarray]:
+    """Calibrates the confident joint for multi_labeled data. Thus,
+        input `labels` is a list of lists (or list of iterable).
+        This is intended as a helper function. You should probably
+        be using `calibrate_confident_joint(multi_label=True)` instead.
+
+
+        See `calibrate_confident_joint` docstring for more info.
+
+    Parameters
+    ----------
+    confident_joint : np.ndarray
+        Refer to documentation for this argument in count.calibrate_confident_joint() for details.
+
+    labels : np.ndarray
+        Refer to documentation for this argument in count.calibrate_confident_joint() for details.
+
+    multi_label : bool, optional
+        Refer to documentation for this argument in count.calibrate_confident_joint() for details.
+
+    Returns
+    -------
+    calibrated_cj : np.ndarray
+      An array of shape ``(K, 2, 2)`` of type float representing a valid
+      estimate of the joint *counts* of noisy and true labels in a one-vs-rest setting."""
     y_one = int2onehot(labels)
     num_classes = len(confident_joint)
     calibrate_confident_joint_list = []
@@ -486,12 +513,28 @@ def _compute_confident_joint_multi_label(
     return_indices_of_off_diagonals: bool, default = False
         If true returns indices of examples that were counted in off-diagonals
         of confident joint as a baseline proxy for the label issues. This
-        sometimes works as well as filter.find_label_issues(confident_joint)."""
+        sometimes works as well as filter.find_label_issues(confident_joint).
+
+    Returns
+    -------
+    confident_joint_counts : np.ndarray
+      An array of shape ``(num_classes,2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
+      estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
+      Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
+
+    Note: if return_indices_of_off_diagonals is True, this function returns confident_joint_counts, indices_off_diagonal
+    where indices_off_diagonal is a list of array of indices counted in off-diagonals of confident joint for each class.
+    """
 
     num_classes = get_num_classes(labels=labels, pred_probs=pred_probs)
-    y_one = int2onehot(labels)
+    try:
+        y_one = int2onehot(labels)
+    except TypeError:
+        raise ValueError(
+            "wrong format for labels, should be a list of list[indices], please check the documentation in find_label_issues for further information"
+        )
     confident_joint_list = np.ndarray(shape=(num_classes, 2, 2), dtype=np.int64)  # type: ignore
-    indices_of_issues_list = []
+    indices_off_diagonal = []
     for class_num in range(0, num_classes):
         pred_probabilitites = _binarize_pred_probs_slice(pred_probs, class_num)
         if return_indices_of_off_diagonals:
@@ -503,7 +546,7 @@ def _compute_confident_joint_multi_label(
                 calibrate=calibrate,
                 return_indices_of_off_diagonals=return_indices_of_off_diagonals,
             )
-            indices_of_issues_list.append(ind)
+            indices_off_diagonal.append(ind)
         else:
             cj = compute_confident_joint(
                 labels=y_one[:, class_num],
@@ -516,7 +559,7 @@ def _compute_confident_joint_multi_label(
         confident_joint_list[class_num] = cj
 
     if return_indices_of_off_diagonals:
-        return confident_joint_list, indices_of_issues_list
+        return confident_joint_list, indices_off_diagonal
 
     return np.array(confident_joint_list)
 
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -115,7 +115,7 @@ def num_label_issues(
 
     if confident_joint is None:
         # Original non-calibrated counts of confidently correctly and incorrectly labeled examples.
-        confident_joint = compute_confident_joint(
+        computed_confident_joint = compute_confident_joint(
             labels=labels, pred_probs=pred_probs, calibrate=False
         )  # type: ignore
     else:
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -456,7 +456,7 @@ def compute_confident_joint(
       An array of shape ``(K, K)`` representing counts of examples
       for which we are confident about their given and true label.
       If multi_label is True,
-      An array of shape ``(num_classes,2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
+      An array of shape ``(K, 2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
       estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
       Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
 
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -160,7 +160,7 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) -> 
       Entry ``(j, k)`` in the matrix is the number of examples confidently counted into the pair of ``(noisy label=j, true label=k)`` classes.
       The `confident_joint` can be computed using :py:func:`count.compute_confident_joint <cleanlab.count.compute_confident_joint>`.
       If not provided, it is computed from the given (noisy) `labels` and `pred_probs`.
-      If multi_label is True, then the confident should be an array of shape ``(K, 2, 2)``.
+      If `multi_label` is True, then the `confident_joint` should be a one-vs-rest array of shape ``(K, 2, 2)``, and an array of the same shape will be returned.
 
     labels : np.ndarray
       A discrete vector of noisy labels, i.e. some labels may be erroneous.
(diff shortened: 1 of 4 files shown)


cleanlab/filter.py
```diff
@@ -208,7 +209,13 @@ class 0, 1, ..., K-1.
...
         "predicted_neq_given",
     ]  # TODO: change default to confident_learning ?
-    assert_valid_inputs(X=None, y=labels, pred_probs=pred_probs, multi_label=multi_label)
+    assert_valid_inputs(
+        X=None,
+        y=labels,
+        pred_probs=pred_probs,
+        multi_label=multi_label,
+        allow_missing_classes=allow_missing_classes,
+    )
```
aditya1503, line 218: just for multi-label

tests/test_filter_count.py
```diff
@@ -341,11 +341,15 @@ def test_find_label_issues_multi_label(multi_label, filter_by):
...
         filter_by=filter_by,
         multi_label=multi_label,
+        allow_missing_classes=multi_label,
     )
     acc = np.mean((data["labels"] != data["true_labels_train"]) == noise_idx)
     # Make sure cleanlab does reasonably well finding the errors.
     # acc is the accuracy of detecting a label error.
-    assert acc > 0.85
+    if multi_label:
+        assert acc > 0.78
```
aditya1503, line 350: lower accuracy for multi_label :(

cleanlab/filter.py
```diff
@@ -398,6 +423,79 @@ class 0, 1, ..., K-1.
...
+
+    Returns
+    -------
+
+    """
+    num_classes = pred_probs.shape[1]
+    y_one = np.zeros((len(labels), num_classes)).astype(np.int32)
+    for class_num in range(0, len(labels)):
+        for j in range(0, len(labels[class_num])):
+            y_one[class_num][labels[class_num][j]] = 1
```
aditya1503, line 462: couldn't get Numpy to apply across rows, figure this out

cleanlab/filter.py
```diff
@@ -66,6 +66,7 @@ def find_label_issues(
     confident_joint=None,
     n_jobs=None,
     verbose=False,
+    allow_missing_classes=False,
```
jwmueller, line 69: I don't think we should add this argument to our public API (which we can never change in the future once we add an argument). Let's just not have multilabel support missing classes right now. We be supporting missing classes shortly in the package FYI, so you don't have to worry about this. Just make sure your multilabel test does not have missing classes in the meantime.

jwmueller: That said if you really need this allow missing functionality for multilabel to work at all, then you can just declare it internally inside your _find_label_issues_multilabel() helper function. The user should not have to know about this though

tests/test_filter_count.py
```diff
@@ -341,6 +341,7 @@ def test_find_label_issues_multi_label(multi_label, filter_by):
         pred_probs=data["pred_probs"],
         filter_by=filter_by,
         multi_label=multi_label,
+        allow_missing_classes=multi_label,
     )
     acc = np.mean((data["labels"] != data["true_labels_train"]) == noise_idx)
```
jwmueller, line 346: Are you sure this line is working properly in multilabel settings? Just want to confirm because I thought `data[labels]` is a list of lists in multilabel settings for which I'm not sure this would work properly?

cleanlab/filter.py
aditya1503: yes I'll fix the test case so it doesn't run into this error

tests/test_filter_count.py
aditya1503: it creates a list of labels with all 1, which throws an error

jwmueller: Right you should specially create properly formatted data for testing multilabel. The given & true labels need to be a list of lists, so I don't think you can simply assess the accuracy in this way.

aditya1503: Got it, with this I can remove allow_missing_classes

cleanlab/filter.py
```diff
@@ -398,6 +423,165 @@ class 0, 1, ..., K-1.
...
+    if return_indices_ranked_by is None:
+        bissues = np.zeros(y_one.shape).astype(bool)
+    else:
+        label_issues_list = []
+    for class_num in range(0, num_classes):
+        pred_probabilitites = np.stack([1 - pred_probs[:, class_num], pred_probs[:, class_num]]).T
+        if confident_joint is None:
+            conf = None
+        else:
+            if len(confident_joint.shape()) != 3:
```
jwmueller, line 552: Can we make this condition more specific?
I think we should check that confident_joint is either (num_classes,k,k) or (k,k) and otherwise throw ValueError that it is not of an acceptable shape.

cleanlab/filter.py
```diff
@@ -398,6 +423,165 @@ class 0, 1, ..., K-1.
...
+                conf = None
+                warnings.warn(
+                    f"The new recommended format for confident_joint in multi_label settings is (num_classes,k,k) (as output by compute_confident_joint(...,multi_label=True)). Your k x k confident_joint in the old format is being ignored."
```
jwmueller, line 555: won't this warning be printed in every iteration of the loop? 
We don't want that, so I think you should move this check before the loop

jwmueller:
Note your code is not passing the type check: https://github.com/cleanlab/cleanlab/actions/runs/3114924202/jobs/5051293689

Sync with @elisno who can walk you through the steps to take to get it to pass, and give you a general overview of our type checking.

cleanlab/filter.py
elisno, line 555: Regardless of whether you move this warning message, you can turn this into a multiline string for readability and you don't have to use an f-string here.

```suggestion
                    "The new recommended format for confident_joint in multi_label settings is (num_classes,k,k) "
                    "(as output by compute_confident_joint(...,multi_label=True)). "
                    "Your k x k confident_joint in the old format is being ignored."
```

cleanlab/filter.py
```diff
@@ -398,6 +423,165 @@ class 0, 1, ..., K-1.
...
+    return_indices_ranked_by: Union[None, str] = None,
+    rank_by_kwargs={},
+    filter_by: str = "prune_by_noise_rate",
+    frac_noise: float = 1.0,
+    num_to_remove_per_class: int = None,
+    min_examples_per_class=1,
+    confident_joint: Union[None, np.array] = None,
+    n_jobs: int = None,
+    verbose: bool = False,
+) -> List:
```
elisno, line 438: Most of the mypy issues are resolved by using `np.ndarray` instead of `np.array`/`List`.

```suggestion
def _find_label_issues_multilabel(
    labels: list,
    pred_probs: np.ndarray,
    return_indices_ranked_by: Union[None, str] = None,
    rank_by_kwargs={},
    filter_by: str = "prune_by_noise_rate",
    frac_noise: float = 1.0,
    num_to_remove_per_class: int = None,
    min_examples_per_class=1,
    confident_joint: Union[None, np.ndarray] = None,
    n_jobs: int = None,
    verbose: bool = False,
) -> np.ndarray:
```

cleanlab/filter.py
elisno: I'm fine with this condition for now, but I agree that we should validate the tuple elements if we're going to throw ValueErrors.

Nevertheless, the shape property is not callable:
```suggestion
            if len(confident_joint.shape) != 3:
```

elisno (commented):
I made some suggestions that address the mypy errors in the CI, and one for formatting.

You may need to remove some unused imports afterwards, like `typing.List`.

---

### One general comment
Please use `Optional[T]` ([docs](https://docs.python.org/3/library/typing.html#typing.Optional)) intead of `Union[None, T]` when explicitly allowing the use of `None` in the function signatures.

cleanlab/filter.py
```diff
@@ -398,6 +421,171 @@ class 0, 1, ..., K-1.
...
+      Note
+      ----
+      Obtain the *indices* of label issues in your dataset by setting
+      `return_indices_ranked_by`.
+
+    """
+    k = get_num_classes(
+        labels=labels, pred_probs=pred_probs, label_matrix=confident_joint, multi_label=True
+    )
+    num_classes = pred_probs.shape[1]
```
aditya1503, line 539: isn't k same as num_classes, we're expecting pred_probs here, 
should multi_label work if pred_probs are None?

aditya1503: add assert pred_probs is not None?

cleanlab/filter.py
```diff
@@ -398,6 +415,185 @@ class 0, 1, ..., K-1.
...
+    min_examples_per_class : int, default=1
+      Minimum number of examples per class to avoid flagging as label issues.
+      This is useful to avoid deleting too much data from one class
+      when pruning noisy examples in datasets with rare classes.
+
+    confident_joint : np.ndarray, optional
+      An array of shape ``(num_classes,K, K)`` representing the confident joint, the matrix used for identifying label issues, which
+      estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
+      Entry ``(j, k)`` in the matrix is the number of examples confidently counted into the pair of ``(noisy label=j, true label=k)`` classes.
+      The `confident_joint` can be computed using :py:func:`count.compute_confident_joint <cleanlab.count.compute_confident_joint>`.
```
aditya1503, line 503: Recheck docs

cleanlab/filter.py
```diff
@@ -398,6 +403,194 @@ class 0, 1, ..., K-1.
...
+        if return_indices_ranked_by is None:
+            bissues[:, class_num] = binary_label_issues
+        else:
+            label_issues_list.append(binary_label_issues)
+            labels_list.append(y_one[:, class_num])
+            pred_probs_list.append(pred_probabilitites)
+    if return_indices_ranked_by is None:
+        return bissues.sum(axis=1) >= 1
+    else:
+        label_issues_idx = reduce(np.union1d, label_issues_list).astype(np.int32)
```
jwmueller, line 574: When I run your test code from your PR description, but I add in: `return_indices_ranked_by='self_confidence', then I get back: `array([4], dtype=int32)`, presumably because of this line.

However that doesn't match output of `find_label_issues(..., multi_label=False, return_indices_ranked_by='self_confidence')` which would typically just look like this: `array([4])`

Can we just make the multi-label case match the existing return type from the regular non-multilabel case? I don't think returning two different types is a good idea

tests/test_filter_count.py
```diff
@@ -332,40 +337,87 @@ def test_pruning_order_method():
...
         pred_probs=data["pred_probs"],
         filter_by=filter_by,
         multi_label=multi_label,
+        return_indices_ranked_by=return_indices_ranked_by,
     )
+    if return_indices_ranked_by is not None:
+        noise_bool = np.zeros(len(labels)).astype(bool)
+        noise_bool[noise_idx] = True
+        noise_idx = noise_bool
     acc = np.mean((data["labels"] != data["true_labels_train"]) == noise_idx)
```
jwmueller, line 358: I don't think this `acc` computation is being properly done to find which indices have label errors in the multi-label case?

tests/test_filter_count.py
```diff
@@ -332,40 +337,87 @@ def test_pruning_order_method():
...
     # Make sure cleanlab does reasonably well finding the errors.
     # acc is the accuracy of detecting a label error.
     assert acc > 0.85
 
 
+@pytest.mark.parametrize(
```
jwmueller, line 364: TODO: replace this test entirely.
Replace all code with multilabel settings and KxK confident joints, they all need to be Kx2x2 anytime multi_label=True in any cleanlab function

tests/test_filter_count.py
aditya1503: This was the same behavior as the previous function, 
I'll rewrite multi_label data generation

tests/test_filter_count.py
```diff
@@ -332,40 +337,87 @@ def test_pruning_order_method():
 @pytest.mark.parametrize(
     "filter_by", ["prune_by_noise_rate", "prune_by_class", "both", "confident_learning"]
 )
-def test_find_label_issues_multi_label(multi_label, filter_by):
+@pytest.mark.parametrize(
```
jwmueller, line 340: We may want extra checks to ensure informative error message gets printed if the user wrongly passes in regular labels with `multi_label=True` (or vice versa).   Currently I get:
```
>>> filter.find_label_issues([0,0,0,1,1,1],pred_probs=np.array([[0.1,0.9],[0.1,0.9],[0.1,0.9],[0.1,0.9],[0.1,0.9],[0.1,0.9]]), return_indices_ranked_by='self_confidence', multi_label=True)

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jonas/Dropbox/Cleanlab/opensource/pr_review/cleanlab/cleanlab/filter.py", line 242, in find_label_issues
    return _find_label_issues_multilabel(
  File "/Users/jonas/Dropbox/Cleanlab/opensource/pr_review/cleanlab/cleanlab/filter.py", line 528, in _find_label_issues_multilabel
    y_one = int2onehot(labels)
  File "/Users/jonas/Dropbox/Cleanlab/opensource/pr_review/cleanlab/cleanlab/internal/util.py", line 243, in int2onehot
    return mlb.fit_transform(labels)
  File "/Users/jonas/virtual/cldev/lib/python3.9/site-packages/sklearn/preprocessing/_label.py", line 794, in fit_transform
    yt = self._transform(y, class_mapping)
  File "/Users/jonas/virtual/cldev/lib/python3.9/site-packages/sklearn/preprocessing/_label.py", line 867, in _transform
    for label in labels:
TypeError: 'int' object is not iterable
```

tests/test_filter_count.py
aditya1503: ok

cleanlab/filter.py
```diff
@@ -138,6 +139,9 @@ class 0, 1, ..., K-1.
       list of labels for each example, instead of just a single label.
       The multi-label setting supports classification tasks where an example has 1 or more labels.
```
jwmueller, line 140:
```suggestion
      The multi-label setting supports classification tasks where an example can belong to more than 1 class (rather than exactly one class as in standard multi-class classification).
```

cleanlab/filter.py
```diff
@@ -138,6 +139,9 @@ class 0, 1, ..., K-1.
...
       Example of a multi-labeled `labels` input: ``[[0,1], [1], [0,2], [0,1,2], [0], [1], ...]``.
+      Confident joint should be an array of shape ``(N, K, K)``
```
jwmueller, line 142:
```suggestion
      Confident joint should be an array of shape ``(K, 2, 2)`` in multi-label settings.
```

jwmueller: Note throughout this package, we reserve `N` for number of examples, `K` for number of classes.

cleanlab/filter.py
```diff
@@ -138,6 +139,9 @@ class 0, 1, ..., K-1.
...
+      The pred_probs need not sum to 1.0
```
jwmueller, line 143:
```suggestion
      Each row of `pred_probs` no longer needs to sum to 1 in multi-label settings, since one example can now belong to multiple classes simultaneously.
```

cleanlab/filter.py
```diff
@@ -138,6 +139,9 @@ class 0, 1, ..., K-1.
       list of labels for each example, instead of just a single label.
       The multi-label setting supports classification tasks where an example has 1 or more labels.
       Example of a multi-labeled `labels` input: ``[[0,1], [1], [0,2], [0,1,2], [0], [1], ...]``.
```
jwmueller, line 141:
```suggestion
      Example of a multi-labeled `labels` input: ``[[0,1], [1], [0,2], [0,1,2], [0], [1], ...]``. This says the first example in dataset belongs to both class 0 and class 1, according to its given label.
```

cleanlab/filter.py
```diff
@@ -398,6 +403,194 @@ class 0, 1, ..., K-1.
...
+    confident_joint: Optional[np.ndarray] = None,
+    n_jobs: int = None,
+    verbose: bool = False,
+) -> np.ndarray:
+    """
+    Parameters
+    ----------
+    labels : list
+      List of noisy labels for multi-label classification where each example can belong to multiple classes(e.g. ``labels = [[1,2],[1],[0],..]``),
+      your labels should instead satisfy: ``len(set(k for l in labels for k in l)) == pred_probs.shape[1])``.
```
jwmueller, line 424:
```suggestion
      List of noisy labels for multi-label classification where each example can belong to multiple classes (e.g. ``labels = [[1,2],[1],[0],..]`` indicates the first example in dataset belongs to both class 1 and class 2.  
      For multi-label settings, your `labels` should instead satisfy: ``len(set(k for l in labels for k in l)) == pred_probs.shape[1])``.
```

cleanlab/filter.py
```diff
@@ -398,6 +403,194 @@ class 0, 1, ..., K-1.
...
+      `x` belongs to each possible class, for each of the K classes. The
+      columns must be ordered such that these probabilities correspond to
+      class 0, 1, ..., K-1. They need not sum to 1.0
+
+      **Caution**: `pred_probs` from your model must be out-of-sample!
+      You should never provide predictions on the same examples used to train the model,
+      as these will be overfit and unsuitable for finding label-errors.
+      To obtain out-of-sample predicted probabilities for every datapoint in your dataset, you can use :ref:`cross-validation <pred_probs_cross_val>`.
+      Alternatively it is ok if your model was trained on a separate dataset and you are only evaluating
+      data that was previously held-out.
```
jwmueller, line 440:
```suggestion
```

jwmueller: can remove this for helper function documentation

cleanlab/filter.py
```diff
@@ -398,6 +403,194 @@ class 0, 1, ..., K-1.
...
+    filter_by: str = "prune_by_noise_rate",
+    frac_noise: float = 1.0,
+    num_to_remove_per_class: int = None,
+    min_examples_per_class=1,
+    confident_joint: Optional[np.ndarray] = None,
+    n_jobs: int = None,
+    verbose: bool = False,
+) -> np.ndarray:
+    """
+    Parameters
```
jwmueller, line 420: In the docstring for this helper method, for any arg where you just copied its docstring from the main `find_label_issues` and this arg is no different in multi-label vs regular classification settings, then let's just have the arg's docstring say:

Refer to documentation for this argument in `filter.find_label_issues()` for details.

cleanlab/filter.py
```diff
@@ -398,6 +403,194 @@ class 0, 1, ..., K-1.
...
+        bissues = np.zeros(y_one.shape).astype(bool)
+    else:
+        label_issues_list = []
+    labels_list = []
+    pred_probs_list = []
+    if confident_joint is not None:
+        confident_joint_shape = confident_joint.shape
+        if confident_joint_shape == (num_classes, num_classes):
+            warnings.warn(
+                f"The new recommended format for confident_joint in multi_label settings is (num_classes,2,2) (as output by compute_confident_joint(...,multi_label=True)). Your k x k confident_joint in the old format is being ignored."
```
jwmueller, line 539:
```suggestion
                f"The new recommended format for `confident_joint` in multi_label settings is (num_classes,2,2) as output by compute_confident_joint(...,multi_label=True). Your K x K  confident_joint in the old format is being ignored."
```

cleanlab/filter.py
aditya1503: ok

jwmueller:
source code for`find_label_issues(..., multi_label=True)` all LGTM!  Just need to update confident joint everywhere it's involved in multilabel and fix up the unit tests.

cleanlab/internal/util.py
```diff
@@ -687,6 +687,12 @@ def num_unique_classes(labels, multi_label=None) -> int:
         return len(set(labels))
 
 
+def _binarize_pred_probs_slice(pred_probs: np.ndarray, class_num: int) -> np.ndarray:
+    pred_probs_class = pred_probs[:, class_num]
+    pred_probabilitites = np.stack([1 - pred_probs_class, pred_probs_class]).T
```
jwmueller, line 692: try to avoid typos in variable names `probabilitites`

cleanlab/count.py
```diff
@@ -281,40 +282,31 @@ def _compute_confident_joint_multi_label(
...
-            np.any(np.delete(pred_probs_bool, k, axis=1)[k_in_l[k]], axis=1) for k in unique_classes
-        ]
-        # Map boolean mask to the indices of the examples, for each class k
-        indices_of_issues_list = [indices_k_in_l[k][issues_mask[k]] for k in unique_classes]
-        # Combine error indices for each class k into a single set of all the indices of errors
-        indices_of_issues = np.asarray(
-            list(set([z for lst in indices_of_issues_list for z in lst]))
+    unique_classes = get_num_classes(labels=labels, pred_probs=pred_probs)
+    y_one = int2onehot(labels)
+    conf_list = []
```
jwmueller, line 287: [nit] would be better to instead pre-allocate this to be a np.array of the right shape ahead of time and fill its slices in with the `cj` from each iteration below

cleanlab/count.py
```diff
@@ -43,6 +42,8 @@
     get_num_classes,
     is_torch_dataset,
     is_tensorflow_dataset,
+    int2onehot,
+    _binarize_pred_probs_slice,
```
jwmueller, line 46: Need to update documentation to describe what shape `_compute_confident_joint_multi_label()` returns, especially with `return_indices_of_off_diagonals`

tests/test_filter_count.py
```diff
@@ -104,11 +105,108 @@ def make_data(
...
+    noisy_labels_idx = generate_noisy_labels(labels_idx, noise_matrix)
+    noisy_labels = [eval(inv_d[i]) for i in noisy_labels_idx]
+    ps = np.bincount(labels_idx) / float(len(labels_idx))
+    inv = compute_inv_noise_matrix(py, noise_matrix, ps=ps)
+
+    clf = MultiOutputClassifier(LogisticRegression())
+    y_train = int2onehot(noisy_labels)
+    clf.fit(X_train, y_train)
+    pyi = clf.predict_proba(X_train)
+    pred_probs = -1 * np.ones(y_train.shape)
```
jwmueller, line 186: could you explain why it is initialized like this, sorry don't exactly follow it

tests/test_filter_count.py
```diff
@@ -180,39 +278,74 @@ def test_prune_on_small_data(filter_by):
     assert not any(noise_idx)
 
 
-def test_calibrate_joint():
+@pytest.mark.parametrize(
+    "multi_label",
+    [True, False],
+)
+def test_calibrate_joint(multi_label):
```
jwmueller, line 285: can we just split this into two separate tests one for multi_label, one for regular classification. The two tests share hardly anything in common, so should not be one function IMO.

tests/test_filter_count.py
```diff
@@ -180,39 +278,74 @@ def test_prune_on_small_data(filter_by):
...
 
 @pytest.mark.parametrize("use_confident_joint", [True, False])
-def test_estimate_joint(use_confident_joint):
-    joint = count.estimate_joint(
-        labels=data["labels"],
-        pred_probs=data["pred_probs"],
-        confident_joint=data["cj"] if use_confident_joint else None,
-    )
+@pytest.mark.parametrize("multi_label", [True, False])
+def test_estimate_joint(use_confident_joint, multi_label):
```
jwmueller, line 329: same here, let's split this into two separate functions, one for multi_label one for regular classification. There is little purpose to have one function that encompasses two branches of logic when the two branches share little common code.

tests/test_filter_count.py
jwmueller: Can you point to where you are resolving comments like these before you mark them resolved? Otherwise it is not so easy for reviewers to figure it out

tests/test_filter_count.py
```diff
@@ -332,42 +465,110 @@ def test_pruning_order_method():
...
+        None,
+        [[[1, 0], [0, 4]], [[3, 0], [0, 2]], [[3, 0], [1, 1]], [[3, 1], [0, 1]]],
+        [[1, 1, 0, 2], [0, 1, 0, 1], [0, 0, 1, 1], [0, 0, 0, 1]],
+    ],
+)
+@pytest.mark.parametrize(
+    "return_indices_ranked_by",
+    [None, "self_confidence", "normalized_margin", "confidence_weighted_entropy"],
+)
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
jwmueller, line 511: what warning are you filtering here?

cleanlab/internal/util.py
```diff
@@ -688,6 +688,12 @@ def num_unique_classes(labels, multi_label=None) -> int:
         return len(set(labels))
 
 
+def _binarize_pred_probs_slice(pred_probs: np.ndarray, class_num: int) -> np.ndarray:
```
jwmueller, line 691: would be nice to add 1-line descriptive comment, eg:
""" This function does XYZ... "

cleanlab/filter.py
```diff
@@ -398,6 +404,182 @@ class 0, 1, ..., K-1.
...
+      An array of shape ``(N, K)`` of model-predicted probabilities,
+      ``P(label=k|x)``. Each row of this matrix corresponds
+      to an example `x` and contains the model-predicted probabilities that
+      `x` belongs to each possible class, for each of the K classes. The
+      columns must be ordered such that these probabilities correspond to
+      class 0, 1, ..., K-1. They need not sum to 1.0
+
+
+    return_indices_ranked_by : {None, 'self_confidence', 'normalized_margin', 'confidence_weighted_entropy'}, default=None
+      Refer to documentation for this argument in filter.find_label_issues() for details.
```
jwmueller, line 483: FYI it is ok here because this is a helper method not displayed in our official documentation, but this is not the way to reference another function properly in the docstrings. You can refer to Development.md guide for an example of how to properly refer to other functions in docstrings, which will be important for any user-facing function.

This one can be left as is though

cleanlab/count.py
```diff
@@ -115,10 +117,10 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
         # Original non-calibrated counts of confidently correctly and incorrectly labeled examples.
         confident_joint = compute_confident_joint(
             labels=labels, pred_probs=pred_probs, calibrate=False
-        )
+        )  # type: ignore
 
     if estimation_method is "off_diagonal":
-        num_issues = np.sum(confident_joint) - np.trace(confident_joint)
+        num_issues = np.sum(confident_joint) - np.trace(confident_joint)  # type: ignore
```
jwmueller, line 123: why do we have to ignore this type? Isn't it clearly a scalar?

jwmueller: consider asking for help if you're unsure how to get mypy to allow this to pass type-checker. If you've already asked Elias and Anish about this and this was the recommended strategy, then 👍 from me too!

tests/test_filter_count.py
aditya1503: it's from PoHe's code, 
I can initialize this with zeros,

tests/test_filter_count.py
aditya1503: Resolved here: https://github.com/aditya1503/cleanlab/blob/682bd91957f54191add1c892e39714fb6a05af6e/cleanlab/filter.py#L532

tests/test_filter_count.py
jwmueller: oh ok so there is no reason to initialize it this way? It's just being overwritten right?

tests/test_filter_count.py
aditya1503: This warning that we print when confident joint is of the old formatting shape


https://github.com/aditya1503/cleanlab/blob/682bd91957f54191add1c892e39714fb6a05af6e/cleanlab/filter.py#L544

tests/test_filter_count.py
aditya1503: yes it's being overridden, can be initialized with zeros, 
just faster than assigning a list and then converting to array

cleanlab/count.py
aditya1503: Fixed: https://github.com/aditya1503/cleanlab/blob/b5e45d622f48959292bc9ec1cc5842acfcee53e9/cleanlab/count.py#L120

cleanlab/count.py
```diff
@@ -510,6 +470,104 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+
+    Returns
+    -------
+    confident_joint_counts : np.ndarray
+      An array of shape ``(num_classes,2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
+      estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
+      Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
+
+    Note: if return_indices_of_off_diagonals is True, this function returns confident_joint_counts, indices_off_diagonal
+    where indices_off_diagonal is a list of array of indices counted in off-diagonals of confident joint for each class.
```
jwmueller, line 530:
```suggestion
    Note: if `return_indices_of_off_diagonals` is set as True, this function instead returns a tuple `(confident_joint_counts, indices_off_diagonal)`
    where `indices_off_diagonal` is a list of arrays (one per class) and each array contains the indices of examples counted in off-diagonals of confident joint for that class.
```

jwmueller: note how we should always use special formatting when referring to variables in docstrings. And for referring to general code blocks, should use double-backticks. Our packages uses the numpy documentation style format, so please look through the relevant links in Development.md to learn how that works.

jwmueller: can you also copy this improved part of the docstring (the part that is not specific to multilabel) to the docstring return-type description above for `compute_confident_joint()` ?

cleanlab/internal/util.py
aditya1503: Fixed: https://github.com/aditya1503/cleanlab/blob/940558449f0dd07d69b775bc20dfbeb41fa311c6/cleanlab/internal/util.py#L692

cleanlab/count.py
```diff
@@ -423,6 +380,8 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
       If `return_indices_of_off_diagonals` is ``True``, `confident_joint_counts` is the first element of returned tuple
       and second element is another array of indices counted in off-diagonals of confident joint.
 
+    Note: if `return_indices_of_off_diagonals` is set as True, this function instead returns a tuple `(confident_joint, indices_off_diagonal)`
+    where `indices_off_diagonal` is an array and each array contains the indices of examples counted in off-diagonals of confident joint.
     Note
```
jwmueller, line 385: Don't you need a space between these?  Have you built the docs based on your PR and verified they look good?

jwmueller:
```suggestion
    where `indices_off_diagonal` is an array and each array contains the indices of examples counted in off-diagonals of confident joint.
    
    Note
```

cleanlab/count.py
```diff
@@ -423,6 +380,8 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
       If `return_indices_of_off_diagonals` is ``True``, `confident_joint_counts` is the first element of returned tuple
       and second element is another array of indices counted in off-diagonals of confident joint.
```
jwmueller, line 381:
```suggestion

```

cleanlab/count.py
aditya1503: I'll double check

cleanlab/count.py
```diff
@@ -199,6 +202,44 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) ->
     return round_preserving_row_totals(calibrated_cj)
 
 
+def _calibrate_confident_joint_multilabel(confident_joint: np.ndarray, labels: list) -> np.ndarray:
```
jwmueller, line 205: Note the docstring for public `calibrate_confident_joint` function seems a little off now that we've updated multilabel. 
https://github.com/cleanlab/cleanlab/blob/9f52e346ea1cf65c8a7425d276b0e29b2076fb4a/cleanlab/count.py#L144

Specifically this part:
```
multi_label : bool, optional
      If ``True``, labels should be an iterable (e.g. list) of iterables, containing a
      list of labels for each example, instead of just a single label.
      The multi-label setting supports classification tasks where an example has 1 or more labels.
      Example of a multi-labeled `labels` input: ``[[0,1], [1], [0,2], [0,1,2], [0], [1], ...]``.
      The major difference in how this is calibrated versus single-label is that
      the total number of errors considered is based on the number of labels,
      not the number of examples. So, the calibrated `confident_joint` will sum
      to the number of total labels.
```

Also the docstring for `confident_joint` arg does not mention the new multilabel shape at all.

jwmueller, line 205: I presume for all the args of this function, only their multilabel versions are supported?  That should be clarified for each arg below, so you should say: 

"See `calibrate_confident_joint` documentation for more information (only the multilabel format of this argument is supported here)"

Note above I am also using more formal language (not statements like "docstring")

jwmueller: Same issue for `confident_joint` arg in `estimate_joint()` documentation. Nothing is mentioned about how it should look for multilabel setting.

tests/test_filter_count.py
```diff
@@ -181,43 +270,99 @@ def test_prune_on_small_data(filter_by):
...
+    else:
+
+        joint = count.estimate_joint(
+            labels=dataset["labels"],
+            pred_probs=dataset["pred_probs"],
+            confident_joint=dataset["cj"] if use_confident_joint else None,
+            multi_label=True,
+        )
+    # Check that joint sums to 1.
+    assert abs(np.sum(joint) - 1.0) < 1e-6
```
jwmueller, line 363: Should't the joint now be K x 2 x 2 in multilabel settings?  I think you forgot to update the `estimate_joint` function?  Please make sure every multilabel occurrence of `confident_joint` in count.py, filter.py is always K x 2 x 2 now.

jwmueller: in fact your test should explicitly check that the shape returned by all new multilabel confident joint functions is always: K x 2 x 2

cleanlab/count.py
```diff
@@ -184,10 +184,13 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) ->
     calibrated_cj : np.ndarray
       An array of shape ``(K, K)`` of type float representing a valid
       estimate of the joint *counts* of noisy and true labels.
+      If multi_label is true, it returns an array of shape ``(K, 2, 2)``,
+      where entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
```
jwmueller, line 188: This notation does not make sense to me. What does ``(class c, noisy label=j, true label=k)`` mean?

jwmueller, line 188:
```suggestion
      An array of shape ``(K, K)`` representing a valid estimate of the joint *counts* of noisy and true labels (if `multi_label` is False).
      If `multi_label` is True, the returned `calibrated_cj` is instead an one-vs-rest array of shape ``(K, 2, 2)``, 
      where for class `c`: entry ``(c, 0, 0)`` in this one-vs-rest  array is the number of examples whose noisy label contains `c` confidently identified as truly belonging to class `c` as well. 
      Entry ``(c, 1, 0)`` in this one-vs-rest  array is the number of examples whose noisy label contains `c` confidently identified as not actually belonging to class `c`. 
      Entry ``(c, 0, 1)`` in this one-vs-rest array is the number of examples whose noisy label does not contain `c` confidently identified as truly belonging to class `c`.
      Entry ``(c, 1, 1)`` in this one-vs-rest array is the number of examples whose noisy label does not contain `c` confidently identified as actually not belonging to class `c` as well.
```

jwmueller: Let's only define the one-vs-rest confident joint with the detailed description in a single place in the code (I'll let you figure out the best place). Then everywhere else, we should just say: 

If `multi_label` is True, the returned <confident-joint-like-array> is instead an one-vs-rest array of shape ``(K, 2, 2)``, see the documentation of <XYZ-function, linked properly in the docstring as described in Development.md> for a detailed description of its entries.

cleanlab/count.py
aditya1503: Fixed : https://github.com/aditya1503/cleanlab/blob/8b088504e89eb7c055de1b2ae9e9405f9ff28329/cleanlab/count.py#L179

cleanlab/count.py
```diff
@@ -510,6 +553,104 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+            )
+        confident_joint_list[class_num] = cj
+
+    if return_indices_of_off_diagonals:
+        return confident_joint_list, indices_off_diagonal
+
+    return confident_joint_list
+
+
 def estimate_latent(
```
aditya1503, line 654: Should this support multi-label as well?
find label issues doesn't call this with multilabal

aditya1503: same Doubt for estimate_confident_joint_and_cv_pred_proba

jwmueller: If these are never ever called by any configuration of `find_label_issues()`, then let's not support multilabel for these.  That said, can you add to the docstrings (for every such function) that:  

Multi-label classification is not supported in this method.

cleanlab/count.py
```diff
@@ -256,92 +310,72 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+
+
+def _estimate_joint_multilabel(labels, pred_probs, *, confident_joint=None) -> np.ndarray:
+    """Parameters
+     ----------
+     labels : np.ndarray
+       An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
+       Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
+       All the classes (0, 1, ..., and K-1) MUST be present in ``labels``, such that:
+       ``len(set(labels)) == pred_probs.shape[1]`` for standard multi-class classification with single-labeled data (e.g. ``labels =  [1,0,2,1,1,0...]``).
```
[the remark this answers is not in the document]
aditya1503: Yes, changing docs to refer to that

cleanlab/filter.py
```diff
@@ -398,6 +411,168 @@ class 0, 1, ..., K-1.
...
+      label issues to return is dependent on the `filter_by` method used. It works by reducing the
+      size of the off-diagonals of the `joint` distribution of given labels and true labels
+      proportionally by `frac_noise` prior to estimating label issues with each method.
+      This parameter only applies for `filter_by=both`, `filter_by=prune_by_class`, and
+      `filter_by=prune_by_noise_rate` methods and currently is unused by other methods.
+      When ``frac_noise=1.0``, return all "confident" estimated noise indices (recommended).
+
+      frac_noise * number_of_mislabeled_examples_in_class_k.
+
+    num_to_remove_per_class : array_like
```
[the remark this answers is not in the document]
aditya1503: Not yet supported in multi_label setting

cleanlab/count.py
aditya1503: Added "Multi-label classification is not supported in this method." to a couple of function docstrings

cleanlab/filter.py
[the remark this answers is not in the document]
aditya1503: https://github.com/aditya1503/cleanlab/blob/159e61259bb70edebd079aca6501a197afcc69bd/cleanlab/filter.py#L159

cleanlab/count.py
```diff
@@ -256,92 +310,72 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+            labels=labels, pred_probs=pred_probs, confident_joint=confident_joint
+        )
+    else:
+        return calibrated_cj / float(np.sum(calibrated_cj))
+
+
+def _estimate_joint_multilabel(labels, pred_probs, *, confident_joint=None) -> np.ndarray:
+    """Parameters
+     ----------
+     labels : np.ndarray
```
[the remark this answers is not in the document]
aditya1503: fixed: https://github.com/aditya1503/cleanlab/blob/159e61259bb70edebd079aca6501a197afcc69bd/cleanlab/count.py#L325

cleanlab/count.py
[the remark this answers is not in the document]
aditya1503: Fixed: https://github.com/aditya1503/cleanlab/blob/159e61259bb70edebd079aca6501a197afcc69bd/cleanlab/count.py#L325

tests/test_filter_count.py
aditya1503: added all tests.

cleanlab/count.py
aditya1503: I've added these statements in the required places.

cleanlab/count.py
```diff
@@ -199,6 +204,49 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) ->
...
+        Refer to documentation for this argument in count.calibrate_confident_joint() for details.
+
+    multi_label : bool, optional
+        Refer to documentation for this argument in count.calibrate_confident_joint() for details.
+
+    Returns
+    -------
+    calibrated_cj : np.ndarray
+      An array of shape ``(K, 2, 2)`` of type float representing a valid
+      estimate of the joint *counts* of noisy and true labels in a one-vs-rest setting."""
```
jwmueller, line 231:
```suggestion
      estimate of the joint *counts* of noisy and true labels in a one-vs-rest fashion."""
```

cleanlab/count.py
```diff
@@ -160,6 +160,7 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) ->
       Entry ``(j, k)`` in the matrix is the number of examples confidently counted into the pair of ``(noisy label=j, true label=k)`` classes.
       The `confident_joint` can be computed using :py:func:`count.compute_confident_joint <cleanlab.count.compute_confident_joint>`.
       If not provided, it is computed from the given (noisy) `labels` and `pred_probs`.
+      If multi_label is True, then the confident should be an array of shape ``(K, 2, 2)``.
```
jwmueller, line 163:
```suggestion
      If `multi_label` is True, then the `confident_joint` should be a one-vs-rest array of shape ``(K, 2, 2)``, and an array of the same shape will be returned.
```

cleanlab/count.py
```diff
@@ -175,19 +176,23 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) ->
       The multi-label setting supports classification tasks where an example has 1 or more labels.
       Example of a multi-labeled `labels` input: ``[[0,1], [1], [0,2], [0,1,2], [0], [1], ...]``.
       The major difference in how this is calibrated versus single-label is that
-      the total number of errors considered is based on the number of labels,
-      not the number of examples. So, the calibrated `confident_joint` will sum
-      to the number of total labels.
+      the calibrated joint is calculated in a one-vs-rest setting, and will return an array of shape ``(K, 2, 2)``
+      each entry k in ``(k,i,j)`` will sum to the number of total labels.
```
jwmueller, line 180: > each entry k in ``(k,i,j)`` will sum to the number of total labels

Not super clear what this means.

cleanlab/count.py
```diff
@@ -199,6 +204,49 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) ->
     return round_preserving_row_totals(calibrated_cj)
 
 
+def _calibrate_confident_joint_multilabel(confident_joint: np.ndarray, labels: list) -> np.ndarray:
+    """Calibrates the confident joint for multi_labeled data. Thus,
```
jwmueller, line 208:
```suggestion
    """Calibrates the confident joint for muli-label classification data. Here
```

cleanlab/count.py
```diff
@@ -199,6 +204,49 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) ->
     return round_preserving_row_totals(calibrated_cj)
 
 
+def _calibrate_confident_joint_multilabel(confident_joint: np.ndarray, labels: list) -> np.ndarray:
+    """Calibrates the confident joint for muli-label classification data. Here
```
jwmueller, line 208:
```suggestion
    """Calibrates the confident joint for multi-label classification data. Here
```

cleanlab/count.py
```diff
@@ -231,6 +279,7 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
       Entry ``(j, k)`` in the matrix is the number of examples confidently counted into the pair of ``(noisy label=j, true label=k)`` classes.
       The `confident_joint` can be computed using :py:func:`count.compute_confident_joint <cleanlab.count.compute_confident_joint>`.
       If not provided, it is computed from the given (noisy) `labels` and `pred_probs`.
+      If multi_label is True, then the confident should be an array of shape ``(K, 2, 2)``.
```
jwmueller, line 282:
```suggestion
      If `multi_label` is True, then the `confident_joint` should instead be a one-vs-rest array of shape ``(K, 2, 2)``.
```

cleanlab/count.py
```diff
@@ -243,6 +292,11 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
     confident_joint_distribution : np.ndarray
       An array of shape ``(K, K)`` representing an
       estimate of the true joint distribution of noisy and true labels.
+      If multi_label is True,
+      An array of shape ``(num_classes,2, 2)`` representing an
+      estimate of the true joint distribution of noisy and true labels,
+      Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
+
```
jwmueller, line 299:
```suggestion
      An array of shape ``(K, K)`` representing an
      estimate of the true joint distribution of noisy and true labels (if `multi_label` is False).
      If `multi_label` is True, an array of shape ``(K, 2, 2)`` representing an
      estimate of the true joint distribution of noisy and true labels for each class in a one-vs-rest fashion.
      Entry ``(c, i, j)`` in this array is the number of examples confidently counted into a``(class c, noisy label=i, true label=j)`` bin, where `i,j` are either 0 or 1 to denote whether this example belongs to class `c` or not (recall examples can belong to multiple classes in multi-label classification).
```

cleanlab/count.py
jwmueller, line 180:
```suggestion
      confident/calibrated joint arrays have shape ``(K, 2, 2)`` formatted in a one-vs-rest fashion such that they contain a 2x2 matrix for each class that counts examples which are correctly/incorrectly labeled as belonging to that class. 
      After calibration, the entries in each class-specific 2x2 matrix will sum to the number of examples.
```

cleanlab/count.py
```diff
@@ -256,92 +308,58 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
-        # Map boolean mask to the indices of the examples, for each class k
-        indices_of_issues_list = [indices_k_in_l[k][issues_mask[k]] for k in unique_classes]
-        # Combine error indices for each class k into a single set of all the indices of errors
-        indices_of_issues = np.asarray(
-            list(set([z for lst in indices_of_issues_list for z in lst]))
+     Returns
+     -------
+     confident_joint_distribution : np.ndarray
+       An array of shape ``(K, 2, 2)`` representing an
+       estimate of the true joint distribution of noisy and true labels in a multi-label setting.
```
jwmueller, line 335:
```suggestion
       estimate of the true joint distribution of noisy and true labels for each class, in a one-vs-rest format employed for multi-label settings.
```

cleanlab/count.py
```diff
@@ -415,13 +433,22 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
       of confident joint as a baseline proxy for the label issues. This
       sometimes works as well as ``filter.find_label_issues(confident_joint)``.
 
+
     Returns
     -------
     confident_joint_counts : np.ndarray
       An array of shape ``(K, K)`` representing counts of examples
       for which we are confident about their given and true label.
```
jwmueller, line 441:
```suggestion
      for which we are confident about their given and true label (if `multi_label` is False).
```

cleanlab/count.py
```diff
@@ -415,13 +433,22 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+      If multi_label is True,
+      An array of shape ``(K, 2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
+      estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
+      Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
+
+
+      Note
+      ----
+      if `return_indices_of_off_diagonals` is set as True, this function instead returns a tuple `(confident_joint, indices_off_diagonal)`
+      where `indices_off_diagonal` is an array and each array contains the indices of examples counted in off-diagonals of confident joint.
```
jwmueller, line 451:
```suggestion
      If `multi_label` is True,
      this array instead has shape ``(K, 2, 2)`` representing a one-vs-rest format for the  confident joint, where for each class `c`:
      Entry ``(c, 0, 0)`` in this one-vs-rest array is the number of examples whose noisy label contains `c` confidently identified as truly belonging to class `c` as well.
      Entry ``(c, 1, 0)`` in this one-vs-rest array is the number of examples whose noisy label contains `c` confidently identified as not actually belonging to class `c`.
      Entry ``(c, 0, 1)`` in this one-vs-rest array is the number of examples whose noisy label does not contain `c` confidently identified as truly belonging to class `c`.
      Entry ``(c, 1, 1)`` in this one-vs-rest array is the number of examples whose noisy label does not contain `c` confidently identified as actually not belonging to class `c` as well.


      Note
      ----
      if `return_indices_of_off_diagonals` is set as True, this function instead returns a tuple `(confident_joint, indices_off_diagonal)`
      where `indices_off_diagonal` is a list of arrays and each array contains the indices of examples counted in off-diagonals of confident joint.
```

cleanlab/count.py
```diff
@@ -415,13 +433,24 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+      Entry ``(c, 1, 1)`` in this one-vs-rest array is the number of examples whose noisy label does not contain `c` confidently identified as actually not belonging to class `c` as well.
+
+
+      Note
+      ----
+      If `return_indices_of_off_diagonals` is set as True, this function instead returns a tuple `(confident_joint, indices_off_diagonal)`
+      where `indices_off_diagonal` is a list of arrays and each array contains the indices of examples counted in off-diagonals of confident joint.
 
     Note
     ----
```
jwmueller, line 456: Can you confirm these two "Note"s are properly formatted in the new docs?

cleanlab/count.py
```diff
@@ -510,6 +539,104 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+
+    thresholds : iterable (list or np.ndarray) of shape (K, 1)  or (K,)
+        P(label^=k|label=k). If an example has a predicted probability "greater" than
+        this threshold, it is counted as having true_label = k. This is
+        not used for filtering/pruning, only for estimating the noise rates using
+        confident counts. This value should be between 0 and 1. Default is None.
+
+    calibrate : bool, default = True
+        Calibrates confident joint estimate P(label=i, true_label=j) such that
+        np.sum(cj) == len(labels) and np.sum(cj, axis = 1) == np.bincount(labels).
```
jwmueller, line 584:
```suggestion
        ``np.sum(cj) == len(labels) and np.sum(cj, axis = 1) == np.bincount(labels)``.
```

cleanlab/count.py
```diff
@@ -510,6 +539,104 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+    will have larger values.
+
+    See `compute_confident_joint` docstring for more info.
+
+    Parameters
+    ----------
+    labels : list of list/iterable (length N)
+        List of noisy labels for multi-label classification. Each list in the list contains
+        all the class assignments for that example. This method will fail if labels
+        is not a list of lists (or a list of np.ndarrays or iterable).
```
jwmueller, line 567:
```suggestion
        Given noisy labels for multi-label classification.
        Must be a list of lists (or a list of np.ndarrays or iterable).
        The i-th list in the list contains the classes the i-th example belongs to.
```

cleanlab/count.py
```diff
@@ -510,6 +539,104 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+        If true returns indices of examples that were counted in off-diagonals
+        of confident joint as a baseline proxy for the label issues. This
+        sometimes works as well as filter.find_label_issues(confident_joint).
+
+    Returns
+    -------
+    confident_joint_counts : np.ndarray
+      An array of shape ``(K, 2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
+      estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
+      Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
```
jwmueller, line 596:
```suggestion
      An array of shape ``(K, 2, 2)`` representing the confident joint of noisy and true labels for each class, in a one-vs-rest format employed for multi-label settings.
```

cleanlab/count.py
```diff
@@ -1230,14 +1365,45 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+    pred_probs: np.ndarray,
+):
+    """Returns expected (average) "self-confidence" for each class.
+
+    The confident class threshold for a class j is the expected (average) "self-confidence" for class j in a one-vs-rest setting.
+
+    Parameters
+    ----------
+    labels: list
+       Refer to documentation for this argument in count.calibrate_confident_joint() with multi_label=True for details.
```
jwmueller, line 1387:
```suggestion
       Refer to documentation for this argument in ``count.calibrate_confident_joint()`` with `multi_label=True` for details.
```

cleanlab/count.py
```diff
@@ -1230,14 +1365,45 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
+    ----------
+    labels: list
+       Refer to documentation for this argument in ``count.calibrate_confident_joint()`` with `multi_label=True` for details.
+    pred_probs : np.ndarray
+       Predicted-probabilities in the same format expected by the :py:func:`get_confident_thresholds <cleanlab.count.get_confident_thresholds>` function.
+
+    Returns
+    -------
+    confident_thresholds : np.ndarray
+      An array of shape ``(K, 2)`` where K is the number of classes.
```
jwmueller, line 1394:
```suggestion
      An array of shape ``(K, 2, 2)`` where `K` is the number of classes, in a one-vs-rest format.
```

cleanlab/filter.py
```diff
@@ -149,6 +154,9 @@ class 0, 1, ..., K-1.
       When ``frac_noise=1.0``, return all "confident" estimated noise indices (recommended).
 
       frac_noise * number_of_mislabeled_examples_in_class_k.
+      Note
+      ----
+      This is not yet supported in multilabel setting.
```
jwmueller, line 159:
```suggestion
      Note: specifying `frac_noise` is not yet supported if `multi_label` is True.
```

cleanlab/filter.py
```diff
@@ -368,18 +384,9 @@ class 0, 1, ..., K-1.
...
-        pred = pred_probs.argmax(axis=1)
+    pred = pred_probs.argmax(axis=1)
     for i, pred_label in enumerate(pred):
-        if (
-            multi_label
-            and np.all(pred_label == labels[i])
-            or not multi_label
-            and pred_label == labels[i]
-        ):
+        if not multi_label and pred_label == labels[i]:
```
jwmueller, line 389:
```suggestion
        if pred_label == labels[i]:
```

jwmueller: You don't have to check for `multi_label` anymore right? Isn't all the multi_label code now moved to the helper `_find_label_issues_multilabel` function?

cleanlab/filter.py
```diff
@@ -20,22 +20,24 @@
```
jwmueller, line 20: Please add this sentence to end of docstring for `confident_joint` arg in `find_label_issues`:
https://github.com/cleanlab/cleanlab/blob/21c1730a80a248732951543bf126b203c903db81/cleanlab/filter.py#L186


If \`multi_label\` is True, confident_joint should instead be a one-vs-rest array with shape \`\`(K, 2, 2)\`\`.

cleanlab/filter.py
```diff
@@ -398,6 +405,189 @@ class 0, 1, ..., K-1.
...
+    frac_noise: float = 1.0,
+    num_to_remove_per_class: int = None,
+    min_examples_per_class=1,
+    confident_joint: Optional[np.ndarray] = None,
+    n_jobs: int = None,
+    verbose: bool = False,
+) -> np.ndarray:
+    """
+
+    A pooling function for _find_multilabel_issues_per_class
```
jwmueller, line 423:
```suggestion
    Finds label issues in multi-label classification data where each example can belong to more than one class. This is done via a one-vs-rest reduction for each class and the  results are subsequently aggregated across all classes.
```

cleanlab/filter.py
```diff
@@ -398,6 +405,189 @@ class 0, 1, ..., K-1.
...
+    num_to_remove_per_class : array_like
+      Refer to documentation for this argument in filter.find_label_issues() for details.
+
+    min_examples_per_class : int, default=1
+      Refer to documentation for this argument in filter.find_label_issues() for details.
+
+    confident_joint : np.ndarray, optional
+      An array of shape ``(num_classes,2, 2)`` representing the confident joint, the matrix used for identifying label issues, which
+      estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
+      Entry ``(c, j, k)`` in the matrix is the number of examples in a one-vs-rest class confidently counted into the pair of ``(class c, noisy label=j, true label=k)`` classes.
```
jwmueller, line 511:
```suggestion
    confident_joint : np.ndarray, optional
      An array of shape ``(K, 2, 2)`` representing a one-vs-rest formatted confident joint.
      Entry ``(c, i, j)`` in this array is the number of examples confidently counted into a ``(class c, noisy label=i, true label=j)`` bin,
      where `i, j` are either 0 or 1 to denote whether this example belongs to class `c` or not
      (recall examples can belong to multiple classes in multi-label classification).
```

cleanlab/count.py
```diff
@@ -242,7 +291,10 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
     -------
     confident_joint_distribution : np.ndarray
       An array of shape ``(K, K)`` representing an
-      estimate of the true joint distribution of noisy and true labels.
+      estimate of the true joint distribution of noisy and true labels (if `multi_label` is False).
+      If `multi_label` is True, an array of shape ``(K, 2, 2)`` representing an
+      estimate of the true joint distribution of noisy and true labels for each class in a one-vs-rest fashion.
+      Entry ``(c, i, j)`` in this array is the number of examples confidently counted into a``(class c, noisy label=i, true label=j)`` bin, where `i,j` are either 0 or 1 to denote whether this example belongs to class `c` or not (recall examples can belong to multiple classes in multi-label classification).
```
jwmueller, line 297:
```suggestion
      Entry ``(c, i, j)`` in this array is the number of examples confidently counted into a ``(class c, noisy label=i, true label=j)`` bin,
      where `i, j` are either 0 or 1 to denote whether this example belongs to class `c` or not
      (recall examples can belong to multiple classes in multi-label classification).
```

cleanlab/filter.py
```diff
@@ -398,6 +405,190 @@ class 0, 1, ..., K-1.
...
+    pred_probs_list = []
+    if confident_joint is not None:
+        confident_joint_shape = confident_joint.shape
+        if confident_joint_shape == (num_classes, num_classes):
+            warnings.warn(
+                f"The new recommended format for `confident_joint` in multi_label settings is (num_classes,2,2) as output by compute_confident_joint(...,multi_label=True). Your K x K confident_joint in the old format is being ignored."
+            )
+            confident_joint = None
+        elif confident_joint_shape != (num_classes, 2, 2):
+            raise ValueError("confident_joint should be of shape (num_classes,2, 2)")
```
jwmueller, line 558:
```suggestion
            raise ValueError("confident_joint should be of shape (num_classes, 2, 2)")
```

tests/test_filter_count.py
```diff
@@ -181,43 +271,115 @@ def test_prune_on_small_data(filter_by):
...
     assert abs(np.sum(joint) - 1.0) < 1e-6
 
 
+@pytest.mark.parametrize("use_confident_joint", [True, False])
+def test_estimate_joint_multilabel(use_confident_joint):
+    dataset = multilabel_data
+    if use_confident_joint:
+        cj = count.compute_confident_joint(
+            labels=dataset["labels"], pred_probs=dataset["pred_probs"], multi_label=True
+        )
```
jwmueller, line 348: assert that `cj` has the right shape

tests/test_filter_count.py
```diff
@@ -181,43 +271,115 @@ def test_prune_on_small_data(filter_by):
...
+
+    else:
+
+        joint = count.estimate_joint(
+            labels=dataset["labels"],
+            pred_probs=dataset["pred_probs"],
+            confident_joint=dataset["cj"] if use_confident_joint else None,
+            multi_label=True,
+        )
+    # Check that each joint sums to 1.
```
jwmueller, line 364: assert that `joint` has the right shape

tests/test_filter_count.py
```diff
@@ -181,43 +271,115 @@ def test_prune_on_small_data(filter_by):
...
+    for class_num in range(0, len(calibrated_cj)):
+        label_counts = np.bincount(y_one[:, class_num])
+        assert all(calibrated_cj[class_num].sum(axis=1).round().astype(int) == label_counts)
+        assert len(dataset["labels"]) == int(round(np.sum(calibrated_cj[class_num])))
+    calibrated_cj2 = count.compute_confident_joint(
+        labels=dataset["labels"],
+        pred_probs=dataset["pred_probs"],
+        calibrate=True,
+        multi_label=True,
+    )
```
jwmueller, line 324: assert `calibrated_cj` and `calibrated_cj2` have the right shapes.

jwmueller (changes requested):
Made bunch of commits with docstring improvements so please look over those to see how future docstrings should look.  Otherwise also had a couple more minor changes to recommend.

tests/test_filter_count.py
```diff
@@ -332,44 +494,131 @@ def test_pruning_order_method():
...
+            [0.9, 0.1, 0.0, 0.4],
+            [0.7, 0.8, 0.2, 0.3],
+            [0.9, 0.8, 0.4, 0.2],
+            [0.1, 0.1, 0.8, 0.3],
+            [0.4, 0.5, 0.1, 0.1],
+        ]
+    )
+    labels = [[0], [0, 1], [0, 1], [2], [0, 2, 3]]
+    cj = count.compute_confident_joint(labels=labels, pred_probs=pred_probs, multi_label=True)
+    assert cj.shape == (4, 2, 2)
```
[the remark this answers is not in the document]
jwmueller: It'd also be nice to add a sanity check that `confident_joint` arg is not simply ignored in `find_label_issues` by adding an extra call to this function that passes in a completely different confident_joint, and verifying the output no longer matches the original output.

tests/test_filter_count.py
```diff
@@ -181,43 +271,115 @@ def test_prune_on_small_data(filter_by):
...
+        joint = count.estimate_joint(
+            labels=dataset["labels"],
+            pred_probs=dataset["pred_probs"],
+            confident_joint=dataset["cj"] if use_confident_joint else None,
+            multi_label=True,
+        )
+    # Check that each joint sums to 1.
+    for j in joint:
+        assert abs(np.sum(j) - 1.0) < 1e-6
+
```
jwmueller, line 367: We should also be checking that output of `estimate_joint` is the same whether or not `confident_joint` was passed in as precomputed object or computed internally in this function. The current test does not appear to check the values of the output beyond that they sum to one.  

I am not sure you can implement my suggestion via `parametrize()`, may have to explicitly call `estimate_joint` multiple times to compare the outputs with each other.

aditya1503: OK

aditya1503: I've added a separate test case, the one that call it multiple times

tests/test_filter_count.py
aditya1503: https://github.com/aditya1503/cleanlab/blob/84b4f960e23c1562dbf254aa8c37dc6544f93d80/tests/test_filter_count.py#L350

tests/test_filter_count.py
aditya1503: https://github.com/aditya1503/cleanlab/blob/84b4f960e23c1562dbf254aa8c37dc6544f93d80/tests/test_filter_count.py#L344

tests/test_filter_count.py
```diff
@@ -181,49 +271,127 @@ def test_prune_on_small_data(filter_by):
...
+        pred_probs=dataset["pred_probs"],
+        confident_joint=dataset["cj"] if use_confident_joint else None,
     )
 
     # Check that joint sums to 1.
     assert abs(np.sum(joint) - 1.0) < 1e-6
 
 
+@pytest.mark.parametrize("use_confident_joint", [True, False])
+def test_estimate_joint_multilabel(use_confident_joint):
```
jwmueller, line 344: I don't think you need the `paramterize` anymore? 
Isn't the test now redundant when `use_confident_joint=False` because you are now already computing `joint_2`. So you should just move all the `assert` statements about shape/sum up and delete the `else` logic.

cleanlab/filter.py
```diff
@@ -176,6 +182,7 @@ class 0, 1, ..., K-1.
       Entry ``(j, k)`` in the matrix is the number of examples confidently counted into the pair of ``(noisy label=j, true label=k)`` classes.
       The `confident_joint` can be computed using :py:func:`count.compute_confident_joint <cleanlab.count.compute_confident_joint>`.
       If not provided, it is computed from the given (noisy) `labels` and `pred_probs`.
+      If `multi_label` is True, confident_joint should instead be a one-vs-rest array with shape ``(K, 2, 2)``.
```
jwmueller, line 185:
```suggestion
      If `multi_label` is True, `confident_joint` should instead be a one-vs-rest array with shape ``(K, 2, 2)``.
```

tests/test_filter_count.py
jwmueller: appears addressed now, thanks!

tests/test_filter_count.py
```diff
@@ -332,44 +500,156 @@ def test_pruning_order_method():
...
+
+    if return_indices_ranked_by is not None:
+        noise_idx = _idx_to_bool(noise_idx)
+        noise_idx2 = _idx_to_bool(noise_idx2)
+        if confident_joint is not None:
+            noise_idx3 = _idx_to_bool(noise_idx3)
+    expected_output = [False, False, False, False, True]
+    assert noise_idx.tolist() == noise_idx2.tolist() == expected_output
+    if confident_joint is not None:
+        assert noise_idx3.tolist() != [False, False, False, False, True]
```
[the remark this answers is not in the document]
aditya1503: https://github.com/aditya1503/cleanlab/blob/bba4234cee1853f60a8fb1336dc0d896cd85d5a5/tests/test_filter_count.py#L587

jwmueller (approved):
Great work, the functionality for multi-label classification is significantly improved now!

Issue #453, opened 2022-09-16
docs: parameter and return types


jwmueller:
Fix types in docstrings under:

Parameters
------------

Returns
--------

to utilize our new auto-typing where possible:  https://github.com/cleanlab/cleanlab/pull/412 

Sphinx can now automatically insert the type for us in the docstring (based on type annotation in the function signature), so you should not specify the type of Parameter or Return-variable *unless* it is a custom/nonstandard type that looks ugly when automatically inserted by Sphinx.  The best strategy would be to remove all the types (only in docstrings, not function signatures!) in a file and rebuild the docs to see how it looks. Then add back the subset of types that were not properly auto-inserted by Sphinx or simply don't look good.  

This has already been done for files token_classification module, which you can reference.
Remaining files to do/verify this for:

- [ ] classification.py
- [ ] count.py
- [ ] dataset.py
- [ ] filter.py
- [ ] multiannotator.py
- [ ] outlier.py
- [ ] rank.py
- [ ] experimental/keras.py
- [ ] experimental/fasttext.py
- [ ] experimental/coteaching.py
- [ ] experimental/mnist_pytorch.py
- [ ] experimental/cifar_cnn.py
- [ ] internal/label_quality_utils.py
- [ ] internal/latent_algebra.py
- [ ] internal/token_classification_utils.py
- [ ] internal/util.py
- [ ] internal/validation.py
- [ ] benchmarking/noise_generation.py

ethanotran:
Can I work on this? Looking for my first open source contribution.

jwmueller:
@ethanotran Yes definitely, thanks for your help!!  I'd just start with a single file. And make sure build the docs before/after your changes following our guide, so when you make the PR you can include a before/after screenshot as comments for easier review.

PR #484, opened 2022-09-27
Cleanlab doc edits

huiwengoh:
- Specified custom h1/h2 tag sizes, the font sizes from the template were very big
- Moved around some code to try to supress tf logs + eliminate error messages

TODO: some docs inconsistencies, eg. an arg has a default value but it is not shown in docs (left as todo as I think #453 might resolve it)

Thanks @calebchiam for the feedback!


docs/source/tutorials/audio.ipynb
```diff
@@ -85,6 +85,11 @@
    "source": [
     "# Package installation (hidden on docs website).\n",
     "# Package versions used: tensorflow==2.9.1 speechbrain==0.5.12 tensorflow-io==0.26.0 torch==1.11.0 torchaudio==0.11.0\n",
+    "\n",
+    "# Supress outputs that may appear if tensorflow happens to be improperly installed: \n",
+    "import os \n",
+    "os.environ[\"TF_CPP_MIN_LOG_LEVEL\"] = \"3\" \n",
+    "\n",
```
jwmueller, line 92: Can we not move this above `dependencies`, which is a variable all tutorial authors must declare and thus should be very close to top of the notebooks.

docs/source/tutorials/text.ipynb
```diff
@@ -107,6 +107,13 @@
     "# Package installation (hidden on docs.cleanlab.ai).\n",
     "# If running on Colab, may want to use GPU (select: Runtime > Change runtime type > Hardware accelerator > GPU)\n",
     "# Package versions we used: tensorflow==2.9.1 scikeras==0.6.1 tensorflow_datasets==4.5.2\n",
+    "\n",
+    "# Supress outputs that may appear if tensorflow happens to be improperly installed: \n",
+    "import os \n",
+    "import logging \n",
+    "os.environ[\"TF_CPP_MIN_LOG_LEVEL\"] = \"3\"  # suppress tensorflow log output \n",
+    "logging.getLogger('tensorflow').setLevel(logging.FATAL) \n",
+    "\n",
```
jwmueller, line 116: Can we not move this above `dependencies`, which is a variable all tutorial authors must declare and thus should be very close to top of the notebooks.

jwmueller (commented):
LGTM, just requesting that we minimize the amount of code in each notebook before the `dependencies` variable is declared.

docs/source/tutorials/audio.ipynb
huiwengoh: I saw [here](https://www.kaggle.com/c/ventilator-pressure-prediction/discussion/274717) that we should set the tf log level before importing tensorflow to suppress the warnings, hence I move it up. 
Changed it so that the `dependencies` variable is above this log but imports are still done below.

jwmueller (approved):
LGTM!

PR #488, opened 2022-09-29
Make type checking pass with mypy 0.981

anishathalye:
Mypy 0.981 adds a check [[1]] to ensure that functions that return a type variable receive at least one argument containing that type variable.

For situations where an instance method returns an instance of the class, this patch annotates the `self` argument with the generic type as well (otherwise, it's inferred to be the class type, which triggers the check added in mypy 0.981). This is the pattern recommended by the mypy documentation [[2]].

This patch also fixes an unrelated type error (missing `type: ignore` annotation).

[1]: https://github.com/python/mypy/commit/6a4c2c843e5f3c743a855022ca9cdce1089c8559
[2]: https://mypy.readthedocs.io/en/stable/generics.html#generic-methods-and-generic-self

CC @elisno @ulya-tkch

### DIFF
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -265,7 +265,7 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
         self.clf_final_kwargs = None
 
     def fit(
-        self,
+        self: TCleanLearning,
         X,
         labels=None,
         *,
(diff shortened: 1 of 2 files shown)


ulya-tkch:
Thanks Anish, makes sense!

jwmueller (approved):

PR #489, opened 2022-09-29
Switch to typing.Self

anishathalye:
PEP 673 [[1]] introduces `typing.Self` as a simple way to annotate methods that return an instance of their class. This patch adopts `typing.Self`.

Because this is only included in Python 3.11, for backwards compatibility, this patch adds typing-extensions [[2]], an officially-supported module that is designed for enabling use of new type system features on older Python versions.

We don't want to make typing-extensions a runtime dependency, so we add a guard and only import it when we're `TYPE_CHECKING`.

[1]: https://peps.python.org/pep-0673/
[2]: https://pypi.org/project/typing-extensions/


anishathalye:
mypy 1.0 has been released, and it includes support for the Self type (https://github.com/python/mypy/pull/14041).

cleanlab/classification.py
```diff
@@ -120,7 +120,10 @@ def score(self, X, y, sample_weight=None):
 import pandas as pd
 import inspect
 import warnings
-from typing import TypeVar, Optional
+from typing import Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
```
elisno, line 125: refactor(classification): 💡 exclude non-executable type-checking code from test coverage report

Code within if TYPE_CHECKING is not executed during runtime, so it won't be covered by tests.

```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

anishathalye: Good call, thanks!

elisno (approved):
LGTM!

PR #490, opened 2022-09-29
Update __init__.py to include major files

ulya-tkch:


elisno (approved):

cleanlab/__init__.py
```diff
@@ -6,3 +6,4 @@
 from . import benchmarking
 from . import dataset
 from . import multiannotator
+from . import outlier
```
jwmueller, line 9:
```suggestion
from . import outlier
from . import token_classification
```

PR #499, opened 2022-10-07
Add basic utilities for handling quality scores for multilabel data

elisno:
This PR adds helper classes that allow us to compute label quality scores for multi-label examples. It also sets up a toy dataset generator for testing the utilities for scoring multi-label examples.

## New internal module

The new functionality is in `cleanlab.internal.multilabel_utils.py` and includes:

- `ClassLabelScorer`: Handles the quality scores for each separate class assignment.
  - This is an enum that only allows quality score functions defined in `cleanlab.rank`([link](https://github.com/cleanlab/cleanlab/blob/master/cleanlab/rank.py))
- `MultilabelScorer`: Uses a base scorer (`ClassLabelScorer`) and pools the scores for each example with a provided aggregation method.
  - By default, uses "self_confidence" for the base score for each class assignment.
  - By default, takes the mean of the scores for each class assignment.
- `get_label_quality_scores`: The multi-label implementation of `cleanlab.rank.get_label_quality_scores`.
  - This uses a `MultilabelScorer` as its "scoring" method, instead of a string. 
- Some helper functions for getting cross-validated `pred_probs` in the multi-label classification setting.
  - We implement our own splitter to support `StratifiedKFold` cross-validation in this setting.

### Usage example

```python
from cleanlab.internal.multilabel_rank_utils import MultilabelScorer, ClassLabelScorer
import numpy as np
scorer = MultilabelScorer(
    base_scorer = ClassLabelScorer.NORMALIZED_MARGIN,
    aggregator = np.min,
)
labels = np.array([[0, 1, 0], [1, 0, 1]])
pred_probs = np.array([[0.1, 0.9, 0.3], [0.4, 0.1, 0.9]])
scores = scorer(labels, pred_probs)
scores
# array([0.7, 0.4])
```

The `get_label_quality_scores` for multi-label quality scores should have its own namespace for that business logic. 

```python
import cleanlab.internal.multilabel_rank_utils as mlrank

def foo_multilabel(...):
    # ...
    scores = mlrank.get_label_quality_scores(...)
    # ...
```

## Testing

I've added a conftest.py with some common testing functionality for multi-label data (currently focused on quality scores).

We manage to create "out-of-sample" `pred_probs` by manipulating the label format and getting `StratifiedKFold` cross-validation with scikit-learn to work for multi-label examples.


elisno:
There might be some overlapping functionality in your PR @aditya1503.

None of the code here should conflict with yours, but we might want to explore ways we can merge these functionalities later.

tests/test_multilabel_rank.py
```diff
@@ -0,0 +1,92 @@
+# Copyright (C) 2017-2022  Cleanlab Inc.
```
jwmueller, line 1: Could we rename this test file to: `test_multilabel_classification.py` and try to move all the easily moveable multi-label related tests in here eventually?  @aditya1503 what do you think?

cleanlab/internal/multilabel_rank_utils.py
```diff
@@ -0,0 +1,186 @@
...
+# cleanlab is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
+
+"""
+Helper classes and functions used internally for computing multilabel label quality scores.
```
jwmueller, line 18:
```suggestion
Helper classes and functions used internally to compute label quality scores in multi-label  classification.
```

jwmueller: Let's try to stick with naming convention "multi-label  classification" throughout. @aditya1503 

Just referring to it as "multi-label" has potential confusion with multi-annotator (as well as other settings we will introduce in the future)

cleanlab/internal/multilabel_rank_utils.py
```diff
@@ -0,0 +1,186 @@
...
+        self.f = f
+
+    def __call__(self, *args, **kwargs):
+        return self.f(*args, **kwargs)
+
+    def __repr__(self):
+        return self.f.__name__
+
+
+class BaseQualityScorer(Enum):
```
jwmueller, line 53: maybe call this something more specific like: `ClassLabelScorer` or `ClassificationLabelScorer` ?  The name is currently a bit vague, especially now that we're starting to branch beyond standard multiclass classification tasks, as well as scoring other issues beyond label errors.

tests/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+    n = y.shape[0]
+    multilabel_counts = [
+        np.sum(np.prod([label[i] == y[:, i] for i in range(y.shape[1])], axis=0))
+        for label in unique_labels
+    ]
+    return np.array(multilabel_counts) / float(n)
+
+
+DATASET_KWARGS = {
+    "n_samples": 500,
```
jwmueller, line 46: seems overly large? We need unit tests to run really fast. 

Code used for performance estimation or benchmarking can go in examples/ repo instead. For now can just be a comment in the PR description too if it's useful to facilitate reviewing.

tests/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+def train_fold(X, labels, *, clf, pred_probs, cv_train_idx, cv_holdout_idx):
+    clf_copy = sklearn.base.clone(clf)
+    X_train_cv, X_holdout_cv, s_train_cv, _ = train_val_split(
+        X, labels, cv_train_idx, cv_holdout_idx
+    )
+    clf_copy.fit(X_train_cv, s_train_cv)
+    pred_probs[cv_holdout_idx] = clf_copy.predict_proba(X_holdout_cv)
+
+
+def get_cross_validated_multilabel_pred_probs(X, labels, *, clf, cv):
```
jwmueller, line 75: seems like useful function, consider moving to cleanlab internal multilabel helper methods file?

cleanlab/internal/multilabel_rank_utils.py
elisno:
```suggestion
class ClassLabelScorer(Enum):
```

tests/conftest.py
elisno: Yup, it's too large for unit tests, but better for benchmarking.
The test runtime isn't affected too drastically right now, but the tests should pass with far fewer samples.

jwmueller: Let's just plan to have your benchmark code go into cleanlab/examples/ eventually. So here let's focus on minimizing runtime of the unit tests. It's a bit cumbersome to revisit individual unit tests in the future to speed them up, so we should always aim to ensure each test runs fast when we first introduce them.

tests/conftest.py
jwmueller: Let's just do this in future PR though, given it seems this requires interfacing with the basic find_label_issues_multilabel unit tests. So let's first merge this PR and @aditya1503's one, and then you can open a new PR to move this function to live in cleanlab.internal.multilabel_utils.

At that time let's also move @aditya1503's utils function: `_binarize_pred_probs_slice()` into internal/multilabel_utils.py as well, so that all multilabel helper functionality is contained in this file.

cleanlab/internal/multilabel_rank_utils.py
```diff
@@ -0,0 +1,186 @@
+# Copyright (C) 2017-2022  Cleanlab Inc.
```
jwmueller, line 1: Let's rename this file: `internal/multilabel_utils.py` and ensure it contains all helper functionality that is specific only to multilabel settings.

tests/test_multilabel_rank.py
aditya1503: I agree

tests/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+    for cv_train_idx, cv_holdout_idx in split_generator:
+        train_fold(
+            X,
+            labels,
+            clf=clf,
+            pred_probs=pred_probs,
+            cv_train_idx=cv_train_idx,
+            cv_holdout_idx=cv_holdout_idx,
+        )
+    return pred_probs
```
aditya1503, line 87: is this functionality similar to : 
`from sklearn.model_selection import cross_val_predict
pred_probs = cross_val_predict(clf, X_train, y_train, method="predict_proba")`

jwmueller: @aditya1503  Does the `sklearn.cross_val_predict()` work well for multilabel?  I've never tried it.  For instance how does it do a stratified split in the multilabel setting?

 If it does it well, then let's just have this function internally convert the labels to proper format for `sklearn.cross_val_predict()` and then call that.

aditya1503: It worked well for me, 
It gave consistent results, I need to see how it does the splitting, will run some tests
I used int2onehot(labels), 

I'll also see if sampling works as well with encoding labels to unique class names

jwmueller: I'd try it with rare classes and check that it is actually stratifying the split properly

elisno: `sklearn.model_selection.StratifiedKFold` ([link](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.StratifiedKFold.html)) does not support (binarized) multi-labels.

This is reflected in the docstring for `cv` in `cross_val_predict` ([link](https://github.com/scikit-learn/scikit-learn/blob/36958fb240fbe435673a9e3c52e769f01f36bec0/sklearn/model_selection/_validation.py#L847-L859)).

---

@aditya1503 It's not clear to me what part of the function you think `cross_val_predict` should replace.
Did you have the following in mind?

```python
def get_cross_validated_multilabel_pred_probs(X, labels, *, clf, cv):
    split_generator = get_split_generator(labels, cv)
    pred_probs = cross_val_predict(clf, X, labels, cv=split_generator, method="predict_proba") 
    return pred_probs
```

cleanlab/internal/multilabel_utils.py
```diff
@@ -0,0 +1,255 @@
...
+Helper classes and functions used internally to compute label quality scores in multi-label classification.
+"""
+
+from enum import Enum
+import itertools
+from typing import Callable, Optional
+
+import numpy as np
+from sklearn.model_selection import cross_val_predict
+from sklearn.utils.multiclass import is_multilabel
```
jwmueller, line 27: FYI depending on sklearn.utils is not ideal:
https://scikit-learn.org/stable/developers/utilities.html#developers-utils

> Warning These utilities are meant to be used internally within the scikit-learn package. They are not guaranteed to be stable between versions of scikit-learn. Backports, in particular, will be removed as the scikit-learn dependencies evolve.

jwmueller: so consider just implementing our own version of this basic function as needed in this file.

cleanlab/internal/multilabel_utils.py
```diff
@@ -0,0 +1,255 @@
...
+    NORMALIZED_MARGIN = _Wrapper(get_normalized_margin_for_each_label)
+    CONFIDENCE_WEIGHTED_ENTROPY = _Wrapper(get_confidence_weighted_entropy_for_each_label)
+
+    def __call__(self, labels: np.ndarray, pred_probs: np.ndarray, **kwargs) -> np.ndarray:
+        """Returns the label-quality scores for each datapoint based on the given labels and predicted probabilities."""
+        return self.value(labels, pred_probs, **kwargs)
+
+
+class MultilabelScorer:
+    """A class for aggregating label quality scores for multi-label classification problems."""
```
jwmueller, line 68:
```suggestion
    """Aggregates label quality scores across different classes to produce one score per example in multi-label classification tasks."""
```

tests/conftest.py
jwmueller: @elisno What is the plan with this file?

jwmueller: I think we should only have tests in: tests/test_multilabel_classification.py

Anything that involves sample-size 500 or does not fit well in that file is probably not a unit test and instead an overall performance benchmark. Those we can put in the examples/ repo as I mentioned above.

jwmueller (changes requested):
Overall looks good, just not sure what is plan with conftest.py and not sure we want to depend on `sklearn.utils` solely for one trivial function as this is not a stable module.

cleanlab/internal/multilabel_utils.py
```diff
@@ -0,0 +1,287 @@
...
+        >>> pred_probs = np.array([[0.1, 0.9, 0.1], [0.4, 0.1, 0.9]])
+        >>> scores = scorer(labels, pred_probs)
+        >>> scores
+        array([0.9, 0.4])
+        """
+        self.base_scorer = base_scorer
+        if aggregator is None:
+            self.aggregator = np.mean
+        else:
+            self.aggregator = aggregator
```
jwmueller, line 139: I think type check is failing somewhere around here FYI

elisno: Weirdly, the trick is to annotate the first assignment in cases like this.
The inferred type of `np.mean` doesn't quite match `Callable[..., np.ndarray]`, but the annotation fixes the failed check.

jwmueller (approved):
Great code structure!

PR #500, opened 2022-10-07
Mypy typechecking fix for count.py

ulya-tkch:
- added assert workaround

### DIFF
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -119,7 +119,7 @@ def num_label_issues(
             calibrate=False,
         )
     else:
-        computed_confident_joint = confident_joint.copy()
+        computed_confident_joint = confident_joint
 
     assert isinstance(computed_confident_joint, np.ndarray)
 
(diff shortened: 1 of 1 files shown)


cleanlab/count.py
```diff
@@ -113,15 +113,21 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
...
         # Original non-calibrated counts of confidently correctly and incorrectly labeled examples.
-        confident_joint = compute_confident_joint(
-            labels=labels, pred_probs=pred_probs, calibrate=False
+        computed_confident_joint = compute_confident_joint(
+            labels=labels,
+            pred_probs=pred_probs,
+            calibrate=False,
         )
+    else:
+        computed_confident_joint = confident_joint.copy()
```
jwmueller, line 122: is the `.copy()` truly necessary here?

ulya-tkch: Nope! I will remove but I try an avoid all direct assignments without it

ulya-tkch (commented):
Removed array copy

cleanlab/count.py
jwmueller: cool just wanted to make sure my understanding of mypy is not wrong :D

jwmueller (approved):
LGTM

Issue #501, opened 2022-10-13
Add mypy type annotations to the code


jwmueller:
We want all existing functions to have mypy type annotations which will make our codebase significantly more robust.
Eventually we want to switch our CI to use mypy strict mode (--strict) once we have full coverage with type annotations.

If you work on this, you can just open PRs with one file at a time. Make sure to run mypy with strict mode to verify your type annotations are properly added.  To figure out which file to work on, you can again just run mypy with `--strict` flag and see which files it complains about.

Example PR to add type annotations for one file: https://github.com/cleanlab/cleanlab/pull/317

Other files with high-quality type annotations for your reference are those in `cleanlab.token_classification`.

ChinoCodeDemon:
I will try my best.

jwmueller:
@ChinoCodeDemon Thanks for your great work on this PR:  https://github.com/cleanlab/cleanlab/pull/502

Any chance you want to help us make the package now pass: `mypy --strict` ?

This will involve declaring the types inside objects like numpy arrays, but would make our codebase much more robust and precise.

PR #502, opened 2022-10-15
Trying to fix typings

ChinoCodeDemon:
I am currently just trying to fix the mypy errors that occur with this project or at least those that i can fix.

This pr is focused to the issue #501

### DIFF
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -203,7 +203,7 @@ def find_label_issues(
     if not rank_by_kwargs:
         rank_by_kwargs = {}
 
-    if not pred_probs:
+    if pred_probs is None:
         pred_probs = np.ndarray([])
 
     assert filter_by in [
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -18,6 +18,7 @@
 Methods to identify which examples have label issues.
 """
 
+from audioop import mul
 import numpy as np
 from sklearn.metrics import confusion_matrix
 from sklearn.preprocessing import MultiLabelBinarizer
@@ -25,7 +26,7 @@ import multiprocessing
 from multiprocessing.sharedctypes import RawArray
 import sys
 import warnings
-from typing import Any, Dict
+from typing import Any, Dict, List, Optional, SupportsIndex, Tuple
 
 from cleanlab.count import calibrate_confident_joint
 from cleanlab.rank import order_label_issues
@@ -591,7 +592,9 @@ def find_label_issues_using_argmax_confusion_matrix(
 mp_params: Dict[str, Any] = {}  # Globals to be shared across threads in multiprocessing
 
 
-def _to_np_array(mp_arr, dtype="int32", shape=None) -> np.ndarray:  # pragma: no cover
+def _to_np_array(
+    mp_arr: bytearray, dtype="int32", shape: Tuple[int, int] = None
+) -> np.ndarray:  # pragma: no cover
     """multipropecessing Helper function to convert a multiprocessing
     RawArray to a numpy array."""
     arr = np.frombuffer(mp_arr, dtype=dtype)
@@ -658,7 +661,8 @@ def _get_shared_data() -> Any:  # pragma: no cover
     )
 
 
-def _prune_by_class(k, args=None) -> np.ndarray:
+# TODO figure out what the types inside args are.
+def _prune_by_class(k: int, args=None) -> np.ndarray:
     """multiprocessing Helper function for find_label_issues()
     that assumes globals and produces a mask for class k for each example by
     removing the examples with *smallest probability* of
@@ -702,7 +706,8 @@ def _prune_by_class(k, args=None) -> np.ndarray:
         return np.zeros(len(labels), dtype=bool)
 
 
-def _prune_by_count(k, args=None) -> np.ndarray:
+# TODO figure out what the types inside args are.
+def _prune_by_count(k: int, args=None) -> np.ndarray:
     """multiprocessing Helper function for find_label_issues() that assumes
     globals and produces a mask for class k for each example by
     removing the example with noisy label k having *largest margin*,
@@ -754,7 +759,7 @@ def _prune_by_count(k, args=None) -> np.ndarray:
     return label_issues_mask
 
 
-def _multiclass_crossval_predict(labels, pyx) -> np.ndarray:
+def _multiclass_crossval_predict(labels: List[List[str]], pyx: np.ndarray) -> np.ndarray:
     """Returns a numpy 2D array of one-hot encoded
     multiclass predictions. Each row in the array
     provides the predictions for a particular example.
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -18,7 +18,6 @@
 Methods to identify which examples have label issues.
 """
 
-from audioop import mul
 import numpy as np
 from sklearn.metrics import confusion_matrix
 from sklearn.preprocessing import MultiLabelBinarizer
@@ -54,19 +53,19 @@ except ImportError as e:  # pragma: no cover
 
 
 def find_label_issues(
-    labels: np.ndarray,
-    pred_probs: Optional[np.ndarray],
+    labels,
+    pred_probs: Optional[np.ndarray] = None,
     *,
-    return_indices_ranked_by=None,
+    return_indices_ranked_by: Optional[str] = None,
     rank_by_kwargs: Optional[Dict[str, Any]] = None,
-    filter_by="prune_by_noise_rate",
-    multi_label=False,
-    frac_noise=1.0,
+    filter_by: str = "prune_by_noise_rate",
+    multi_label: bool = False,
+    frac_noise: float = 1.0,
     num_to_remove_per_class: Optional[int] = None,
-    min_examples_per_class=1,
+    min_examples_per_class: int = 1,
     confident_joint: Optional[np.ndarray] = None,
     n_jobs: Optional[int] = None,
-    verbose=False,
+    verbose: bool = False,
 ) -> np.ndarray:
     """
     Identifies potentially bad labels in a dataset (with `N` examples) using confident learning.
@@ -288,7 +287,7 @@ def find_label_issues(
                 _labels = RawArray("I", labels)  # type: ignore
             _label_counts = RawArray("I", label_counts)
             _prune_count_matrix = RawArray("I", prune_count_matrix.flatten())  # type: ignore
-            _pred_probs = RawArray("f", pred_probs.flatten())
+            _pred_probs = RawArray("f", pred_probs.flatten())  # type: ignore
         else:  # Multiprocessing is turned off. Create tuple with all parameters
             args = (
                 labels,
@@ -405,7 +404,7 @@ def find_label_issues(
 
 
 def _keep_at_least_n_per_class(
-    prune_count_matrix: np.ndarray, n: int, *, frac_noise=1.0
+    prune_count_matrix: np.ndarray, n: int, *, frac_noise: float = 1.0
 ) -> np.ndarray:
     """Make sure every class has at least n examples after removing noise.
     Functionally, increase each column, increases the diagonal term #(true_label=k,label=k)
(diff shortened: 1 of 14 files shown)


elisno (commented):
Thanks for the PR @ChinoCodeDemon!

We're putting most of our typing-effort into modules that are in `cleanlab/`, not `docs/` or `tests/`. Maybe put your focus on `cleanlab/filter.py` for this PR and see how far you get?

Could you share how you come across these errors?

ChinoCodeDemon:
Since I didn't find many mypy problems, I used a simple `mypy $(git ls-files "*.py")`
The `mypy cleanlabs` has in the commit no errors anymore, but i should at least pass it through black.

cleanlab/filter.py
```diff
@@ -588,7 +588,7 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
 
 # Multiprocessing helper functions:
 
-mp_params = {}  # Globals to be shared across threads in multiprocessing
+mp_params: Dict[str, Any] = {}  # Globals to be shared across threads in multiprocessing
```
ChinoCodeDemon, line 591: I don't know if it's a `Dict[str, str]`.

ChinoCodeDemon:
After fixing some errors, i will then try to find out what types the functions in `filter.py` take.

cleanlab/filter.py
elisno: None of the values in `mp_params` should be considered strings.

Some of the values can be lists, arrays, booleans, integers, tuples...
Instead of nailing each case down, the way you did it with `Dict[str, Any]` should work for now.

ChinoCodeDemon:
I would like a review over the parameter typings for now.
I do not really know what `args` really contains and i can only guess the types of the stuff in there.
```py
(
            labels,
            label_counts,
            prune_count_matrix,
            pred_probs,
            multi_label,
            min_examples_per_class,
        ) = args
```

`labels` looks for me like a `List[str]` or `List[List[str]]`
`label_counts` is from my perspective an `int`

Guess I should make a named tuple for this.

docs/source/conf.py
```diff
@@ -81,7 +82,7 @@
 napoleon_use_param = True
 napoleon_use_rtype = False
 napoleon_preprocess_types = True
-napoleon_type_aliases = None
+napoleon_type_aliases: Optional[Dict[str, str]] = None
```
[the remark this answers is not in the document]
ChinoCodeDemon: That one just annoyed me with the mypy deamon, probably gonna remove that typing.

ChinoCodeDemon:
I also see some inconsistency with 3.10 and 3.8 annotations in different files.
Mixed use of `dict`, `Dict` and similar.

ChinoCodeDemon:
While I added some typings to a method/function I also fixed some dangerous default values.

cleanlab/filter.py
```diff
@@ -53,18 +54,18 @@
...
 
 def find_label_issues(
-    labels,
-    pred_probs,
+    labels: np.ndarray,
+    pred_probs: Optional[np.ndarray],
     *,
     return_indices_ranked_by=None,
-    rank_by_kwargs={},
+    rank_by_kwargs: Optional[Dict[str, Any]] = None,
```
ChinoCodeDemon, line 61: This is a bit more secure since, dict and list objects as default values are kinda "static".
I also don't really know what type the `return_indices_ranked_by` should be.

elisno:
> `labels` looks for me like a `List[str]` or `List[List[str]]` `label_counts` is from my perspective an `int`

Currently, each class assignment in a "`label`" is generally an `int` and not a `str`.
We are however looking into support for string labels throughout the package.

cleanlab/filter.py
elisno: Right now, those are strings.

cleanlab/filter.py
```diff
@@ -18,14 +18,15 @@
 Methods to identify which examples have label issues.
 """
 
+from audioop import mul
```
elisno, line 21: Was this some accidental auto-import?
```suggestion
```

cleanlab/filter.py
```diff
@@ -53,18 +54,18 @@
 
 
 def find_label_issues(
-    labels,
-    pred_probs,
+    labels: np.ndarray,
+    pred_probs: Optional[np.ndarray],
```
elisno, line 58: We use `Optional` as a shorthand for `Union[None, ...]` types we provide `None` as a default argument.

In this case, I wouldn't expect users to run:

```python
from cleanlab.filter import find_label_issues
import numpy as np

labels = np.random.randint(0,2, (10,))
find_label_issues(labels, None)
```
as this doesn't pass the data-validation check at the start of the function.
Here, it's better to keep `pred_probs` as a positional argument with no default value.



```suggestion
    pred_probs: np.ndarray,
```

cleanlab/filter.py
```diff
@@ -200,6 +201,11 @@ class 0, 1, ..., K-1.
       Obtain the *indices* of label issues in your dataset by setting
       `return_indices_ranked_by`.
     """
+    if not rank_by_kwargs:
+        rank_by_kwargs = {}
+
+    if not pred_probs:
+        pred_probs = np.ndarray([])
```
elisno, line 208: `pred_probs` should remain as required positional arguments.
It should be validated by `assert_valid_inputs` early in the function.

```suggestion
```

cleanlab/filter.py
```diff
@@ -398,7 +404,9 @@ class 0, 1, ..., K-1.
     return label_issues_mask
 
 
-def _keep_at_least_n_per_class(prune_count_matrix, n, *, frac_noise=1.0) -> np.ndarray:
+def _keep_at_least_n_per_class(
+    prune_count_matrix: np.ndarray, n: int, *, frac_noise=1.0
```
elisno, line 408:
```suggestion
    prune_count_matrix: np.ndarray, n: int, *, frac_noise: float = 1.0
```

cleanlab/filter.py
```diff
@@ -464,7 +472,7 @@ def _keep_at_least_n_per_class(prune_count_matrix, n, *, frac_noise=1.0) -> np.n
     return round_preserving_row_totals(new_mat).astype(int)
 
 
-def _reduce_prune_counts(prune_count_matrix, frac_noise=1.0) -> np.ndarray:
+def _reduce_prune_counts(prune_count_matrix: np.ndarray, frac_noise=1.0) -> np.ndarray:
```
elisno, line 475:
```suggestion
def _reduce_prune_counts(prune_count_matrix: np.ndarray, frac_noise: float = 1.0) -> np.ndarray:
```

cleanlab/filter.py
```diff
@@ -495,7 +503,9 @@ def _reduce_prune_counts(prune_count_matrix, frac_noise=1.0) -> np.ndarray:
     return new_mat.astype(int)
 
 
-def find_predicted_neq_given(labels, pred_probs, *, multi_label=False) -> np.ndarray:
+def find_predicted_neq_given(
+    labels: np.ndarray, pred_probs: np.ndarray, *, multi_label=False
```
elisno, line 507:
```suggestion
    labels: np.ndarray, pred_probs: np.ndarray, *, multi_label: bool = False
```

cleanlab/filter.py
```diff
@@ -754,7 +768,7 @@ def _prune_by_count(k, args=None) -> np.ndarray:
     return label_issues_mask
 
 
-def _multiclass_crossval_predict(labels, pyx) -> np.ndarray:
+def _multiclass_crossval_predict(labels: List[List[str]], pyx: np.ndarray) -> np.ndarray:
```
elisno, line 771: For now, let's just focus on the annotations:

```suggestion
def _multiclass_crossval_predict(labels: List[List[int]], pyx: np.ndarray) -> np.ndarray:
```

cleanlab/filter.py
elisno: Note that either annotation to `pred_probs` leads to a mypy-error:

```
cleanlab/filter.py:292: error: No overload variant of "RawArray" matches argument types "str", "ndarray[Any, Any]"  [call-overload]
```
at 
https://github.com/ChinoCodeDemon/cleanlab/blob/63400f926a5aac4308ffc22ad5f7cae1ed2e68c1/cleanlab/filter.py#L291

For now, try adding to the relevant line:

```diff
-            _pred_probs = RawArray("f", pred_probs.flatten())
+            _pred_probs = RawArray("f", pred_probs.flatten())  # type: ignore
```

cleanlab/filter.py
```diff
@@ -53,18 +54,18 @@
 
 
 def find_label_issues(
-    labels,
-    pred_probs,
+    labels: np.ndarray,
```
elisno, line 57: Annotating `labels` in _this particular function_ opens a whole can of worms with other mypy-errors.
Using `cleanlab.typing.LabelLike` does not seem to resolve those errors at this time.

I recommend that we leave this argument unannotated for now so we can keep the type-checker happy and work on merging your other changes.

```suggestion
    labels,
```

cleanlab/filter.py
ChinoCodeDemon: This makes it a lot easier, at the start I didn't even know what label actually was.

ChinoCodeDemon:
I learn a lot about the project like that, at the start I just searched for some issues to solve out of boredom.

cleanlab/filter.py
ChinoCodeDemon: So the doc string is wrong at that spot?
> pred_probs : np.ndarray, optional

cleanlab/filter.py
ChinoCodeDemon: yes this is an accidental import.

ChinoCodeDemon:
Had to handle some conflicts.

cleanlab/filter.py
```diff
@@ -406,9 +411,9 @@ class 0, 1, ..., K-1.
...
 
-def _find_label_issues_multilabel(
-    labels: list,
-    pred_probs: np.ndarray,
+def _keep_at_least_n_per_class(
+    prune_count_matrix: np.ndarray, n: int, *, frac_noise: float = 1.0
+) -> np.ndarray:
     return_indices_ranked_by: Optional[str] = None,
     rank_by_kwargs={},
     filter_by: str = "prune_by_noise_rate",
```
elisno, line 419: Looks like the merge conflicts weren't resolved properly. There's a syntax error on line 426.

```
cleanlab/filter.py:426: error: unmatched ')'  [syntax]
```

ChinoCodeDemon:
Ok, I will fix it when I am back home.

ChinoCodeDemon:
I hope that fixes it.

elisno:
> I hope that fixes it.

Unfortunately, no.
The target branch was recently updated, but your changes seem to delete those.

ChinoCodeDemon:
can that be resolved over github?
i reverted it again.

cleanlab/filter.py
elisno: I'll have to think about that.

elisno:
I've handled the merge conflict.
I've also changed the annotation for `pred_probs` in `find_label_issues`, as mentioned in https://github.com/cleanlab/cleanlab/pull/502#discussion_r996505388.

ChinoCodeDemon:
I think we got everything, at least I don't notice any more typing issues.

elisno (approved):
Thanks for these annotations @ChinoCodeDemon!

We'll leave the docstrings for now.

jwmueller:
@ChinoCodeDemon Thanks for your great work on this!  Any chance you want to help us make the package now pass: `mypy --strict` ?

This will involve declaring the types inside objects like numpy arrays, but would make our codebase much more robust and precise.

ChinoCodeDemon:
> @ChinoCodeDemon Thanks for your great work on this!  Any chance you want to help us make the package now pass: `mypy --strict` ?
> 
> This will involve declaring the types inside objects like numpy arrays, but would make our codebase much more robust and precise.

I could look over it, when I have the time. I am currently pretty much unavailable.

PR #503, opened 2022-10-17
update paper links

jwmueller:


huiwengoh (approved):
Made a small change to properly (un)nest the sidebar contents but otherwise LGTM!

PR #504, opened 2022-10-20
Add internal function to properly format labels

huiwengoh:
- Added internal function `format_labels` to map users' labels into proper formats for other cleanlab functions
- Added checks and more decresiptive error messages for validating if multiannotator labels are formatted properly
- Extracted multiannotator related internal functions to `multiannotator_utils.py` 

potential TODO: part of `format_multiannotator_labels` has a bunch of weird type conversions and checks because `np.unique`/`np.sort` does not support the mix of `np.nan` and `str` types. The function works now but feels a bit hacky, definitely room for revisiting in the future to make it cleaner. If there are obvious ways to improve that I'm missing please let me know!

PR #506, opened 2022-10-24
Mention internal format label function in multiannotator docs

huiwengoh:
- mention the newly added function to format labels (#504)  in multiannotator docstring + tutorial
- fix tiny bug in converting pd.NA to np.NaN


jwmueller (approved):
Great pointers!

PR #507, opened 2022-10-25
Add EMA multi-label score aggregator

elisno:
This PR adds more helper functions and classes to the internal `multilabel_utils.py` module

## New functionality

- `exponential_moving_average`: A function that aggregates scores, placing more significance on low label quality scores. After sorting a score vector, $s$, (in descending order), an exponential moving average (EMA) filter is applied to the vector and the last (worst) score is returned. ([relevant Wikipedia article](https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average))
  - The plan is to make this the default aggregator in the multi-label classification.
- `Aggregator`: A helper class that handles the calls to the aggregation functions and validates their inputs.
- `from_str` constructor for `ClassLabelScorer`: This should make it easy to integrate with existing label quality score functionality that picks methods based on their names.

## Docs updates
Docs for the internal `multilabel_utils.py` module have also been added.
- Missing docstrings have been added.
- Usage examples added to most of the classes/functions.

PR #508, opened 2022-10-25
Use get_label_quality_scores from multi-label utils

elisno:
Continued work from #507.

Some necessary changes to support existing functionality.

- Support separate kwargs for scorer and aggregator.
- `_find_multilabel_issues_per_class` need to return `labels` and `pred_probs` as arrays instead of lists of arrays.


elisno:
Close in favor of #509

PR #511, opened 2022-10-28
Add support for missing classes

cgnorthcutt:
## Extends cleanlab to support missing classes

This PR adds support when `pred_probs.shape[1] > len(set(labels))` for most of the main methods in the `count` and `filter` modules. See examles below.

## Completed

* Missing classes should now work for both single-labeled data and multi-labeled data (list of lists) for methods in `count` and `filter` modules. I tested `compute_confident_joint`, `find_label_issues`, and `get_confident_thresholds`.
* Added two tests to verify basic functionality.
* All other tests still pass.

If you're curious, the most complicated parts of this PR was getting everything to work smoothly with multi_label which we did not plan for in advance (including parallelization). It should work now, but more testing is needed.

## TODO

- [] Did not check if `CleanLearning` works with new functionality. Only supported `filter` and `count` modules.
- [] Much more extensive testing should be added. (recommended tests below)

## Tests TODO (strongly recommend adding/checking these tests)
* Check that all functions in `count` and `filter` work with missing classes, not just the three main ones I checked.
* Check that `CleanLearning` works with missing classes.
* Check that this works for all combinations of hyper-parameters, particularly al filter_by methods and for multi_label=True/False
* Check that this works if setting n_jobs=1 versus using parallelization (not setting n_jobs) because I had to edit how the parallelization works with multi_label to get this to work.
* Replicate variations of my sanity checks below (see screenshots)
* Double check that the classes are mapping correctly for all methods.
* Check that cleanlab.count.estimate_py_noise_matrices_and_cv_pred_proba works if you pass in labels [0, 1, 3] 
* (general check for multi_label) Check that for a dataset of 1000 datapoints, replacing the label of a single item from [0] --> [0,1] to make it multi_label does not have a significant impact on the predicted_probabilities produced. (this is how we'll know multi_label seems to work correctly.

### 1. removing a class by changing the fifth label from 1 --> 2
 
image

LOOKS GOOD! Confident joint updates correctly. Label errors update correctly.

## 2. Does multi-label work with missing classes?

```python
labels = [[0], [0], [0], [0, 2], [2], [2], [2], [2]]
# Third and fourth examples are the label errors
pred_probs = np.array([[0.9, 0.1, 0.0], [0.8, 0.1, 0.1], [0.1, 0.0, 0.9], [0.9, 0.0, 0.1],
                       [0.1, 0.3, 0.6], [0.1, 0.0, 0.9], [0.1, 0.0, 0.9], [0.1, 0.0, 0.9],
                      ])
cj = cleanlab.count.compute_confident_joint(labels, pred_probs, multi_label=True)
issues = cleanlab.filter.find_label_issues(labels, pred_probs, multi_label=True, filter_by='prune_by_class')
print('label issues found:', issues)
print(cj)
```
OUTPUT (looks good!)
```
label issues found: [False False  True  True False False False False]
[[3 0 1]
 [0 0 0]
 [1 0 4]]
 ```

## Two logical issues to be aware of (this PR does NOT renormalize predicted probabilities)

These are not errors, but consider alternatives... currently if the predicted probabilities are high on a missing class, a label error will not be detected since there are no labels in that class and thus no data for cleanlab to work with for that class. That probability mass is effectively lost. I left as is because I recommend merging this PR first. This PR handles the software mechanics of supporting missing classes. Its complicated enough as is. Easy to decide if you want to renormalize once this PR is merged.

## 1. Example of label error not counted because prob mass on missing class

image

### 2. Example of label error no longer being detected as probability mass shifts to a missing class
image

See the fourth True/False value

### DIFF
--- a/cleanlab/internal/util.py
+++ b/cleanlab/internal/util.py
@@ -264,6 +264,7 @@ def round_preserving_row_totals(confident_joint) -> np.ndarray:
     ).astype(int)
 
 
+# TODO: move to multilabel_utils.py
 def int2onehot(labels, K) -> np.ndarray:
     """Convert list of lists to a onehot matrix for multi-labels
 
@@ -282,6 +283,7 @@ def int2onehot(labels, K) -> np.ndarray:
     return mlb.fit_transform(labels)
 
 
+# TODO: move to multilabel_utils.py
 def onehot2int(onehot_matrix) -> list:
     """Convert a onehot matrix for multi-labels to a list of lists of ints
 
--- a/cleanlab/internal/util.py
+++ b/cleanlab/internal/util.py
@@ -21,7 +21,7 @@ Ancillary helper methods used internally throughout this package; mostly related
 import warnings
 import numpy as np
 import pandas as pd
-from typing import Any, Union, Tuple
+from typing import Union, Tuple
 
 from cleanlab.typing import DatasetLike, LabelLike
 from cleanlab.internal.validation import labels_to_array
--- a/cleanlab/internal/util.py
+++ b/cleanlab/internal/util.py
@@ -177,7 +177,8 @@ def value_counts(x, *, num_classes=None, multi_label=False) -> np.ndarray:
     if num_classes <= max(unique_classes):
         raise ValueError(f"Required: num_classes > max(x), but {num_classes} <= {max(x)}.")
     # Add zero counts for all missing classes in [0, 1,..., num_classes-1]
-    missing_classes = get_missing_classes(x, num_classes=num_classes, multi_label=multi_label)
+    # multi_label=False regardless because x was flattened.
+    missing_classes = get_missing_classes(x, num_classes=num_classes, multi_label=False)
     missing_counts = [(z, 0) for z in missing_classes]
     # Return counts with zeros for all missing classes.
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by cgnorthcutt** [comment]
Please edit directly. I don't plan to make further edits.

File: cleanlab/internal/util.py [hunk 1]
```diff
@@ -681,12 +726,22 @@ def num_unique_classes(labels, multi_label=None) -> int:
...
     like this: [1, [1,2], [0], [0, 1], 2, 1]"""
+    return len(get_unique_classes(labels, multi_label))
+
+
+def get_unique_classes(labels, multi_label=None) -> set:
+    """Returns the set of unique classes for both single-labeled
+    and multi-labeled labels. If multi_label is set to None (default)
+    this method will infer if multi_label is True or False based on
+    the format of labels.
+    This allows for a more general form of multiclass labels that looks
```
aditya1503, line 737: should we extend this to all multilabel formats?
just a question, I can do this once it's merged

aditya1503 (approved):

cleanlab/internal/util.py
jwmueller: only if you need it for multilabel find-label-issues to work, when some of the classes never appear in data.

Our main priority is just to get multilabel find-label-issues to work when some of the classes never appear in data.

PR #514, opened 2022-10-31
Separate word_coloring from token_replacement in color_sentence

### LINKED ISSUE #513: Ensure unit tests work with termcolor v2.1.0
tests/test_token_classification.py currently fails after termcolor just upgraded to v2.1.0, specifically `test_color_sentence` 

- [ ] update unit test code to make them pass with termcolor v2.1.0
- [ ] ensure new unit test code also works with older versions of termcolor pre v2.1.0 (suboptimal but ok if the unit test only works with versions post v2.0.0, as long as the package works with all termcolor versions currently supported).
- [ ] remove version upper bound on termcolor if it has been added to package in the meantime.

https://pypi.org/project/termcolor/
https://github.com/termcolor/termcolor/pull/25/files
https://github.com/cleanlab/cleanlab/actions/runs/3357515340/jobs/5563372689

elisno:
Tests in CI fail due to isatty-check in the latest update to the termcolor-library.

The regex/replace functionality has been moved to a separate function.

Tests can be reused in this case, without relying on termcolor.

It's best to patch this attribute for testing the termcolor integration.

Fixes #513

### DIFF
--- a/cleanlab/internal/token_classification_utils.py
+++ b/cleanlab/internal/token_classification_utils.py
@@ -249,10 +249,35 @@ def color_sentence(sentence: str, word: str) -> str:
     'This is a \x1b[31msentence\x1b[0m. This is another \x1b[31msentence\x1b[0m.'
     """
     colored_word = colored(word, "red")
-    colored_sentence, number_of_substitions = re.subn(
-        r"\b{}\b".format(re.escape(word)), colored_word, sentence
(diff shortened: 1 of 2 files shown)


aditya1503 (commented):
Just an additional test case, lgtm!

aditya1503 (approved):

PR #518, opened 2022-11-02
Better missing class support for label quality scoring

jwmueller:
Thanks @huiwengoh for suggesting these improvements!

### DIFF
--- a/cleanlab/internal/validation.py
+++ b/cleanlab/internal/validation.py
@@ -31,7 +31,7 @@ def assert_valid_inputs(
     pred_probs: Optional[np.ndarray] = None,
     multi_label: bool = False,
     allow_missing_classes: bool = True,
-    one_class_allowed: bool = False,
+    allow_one_class: bool = False,
 ) -> None:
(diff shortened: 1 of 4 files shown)


cleanlab/internal/validation.py
```diff
@@ -31,13 +31,16 @@ def assert_valid_inputs(
     pred_probs: Optional[np.ndarray] = None,
     multi_label: bool = False,
     allow_missing_classes: bool = True,
+    one_class_allowed: bool = False,
```
huiwengoh, line 34: [nit] should we change this arg name to `allow_one_class` to keep it consistent with `allow_missing_classes`? 
Alternatively, can change `allow_missing_classes` to `missing_classes_allowed`.

huiwengoh (approved):
Just had a small suggestion, otherwise LGTM!

cleanlab/internal/validation.py
jwmueller: renamed

PR #519, opened 2022-11-03
support missing classes in multiannotator functions

huiwengoh:


cleanlab/internal/multiannotator_utils.py
```diff
@@ -31,22 +31,30 @@ def assert_valid_inputs_multiannotator(
...
     """Validate format of multi-annotator labels"""
+    # Raise error if labels are not formatted properly
+    if any([isinstance(label, str) for label in labels_multiannotator.values.ravel()]):
+        raise ValueError(
+            "Labels cannot be strings, they must be zero-indexed integers corresponding to class indices."
+        )
+
+    all_labels_flatten = labels_multiannotator.replace({pd.NA: np.NaN}).astype(float).values.ravel()
+    all_labels_flatten = all_labels_flatten[~np.isnan(all_labels_flatten)]
+    assert_valid_class_labels(all_labels_flatten, allow_one_class=True)
```
jwmueller, line 42: is it really ok if every single label is the same class?

huiwengoh: Code will not break, but we will probably get bad results

jwmueller (approved):
LGTM!

PR #524, opened 2022-11-04
fix mypy typing

huiwengoh:


jwmueller (approved):
LGTM. 

Strange that this code change to naive for loop was needed to pass mypy type-check only  on CI ...

PR #527, opened 2022-11-07
Validate forgetting factor in EMA

elisno:
# Changes

- Added runtime check that `0 <= alpha <= 1` to ensure Exponential Moving Average uses proportional weighting.
- Moved all EMA tests to separate class.


aditya1503 (approved):
LGTM

Issue #487, opened 2022-09-29
CI: Replace pylint -> flake8


jwmueller:
Try removing `pylint` and replacing it with `flake8` linter.
Here's a flake8 config we'd like to try out:
```
[flake8]
ignore = E501, E203, W503
count = True
max-line-length = 100
statistics = True
```

Open PR after you make the changes and let's see what the new flake8 linter complains about.  You may have to disable some flake-8 warnings such as: `F401`, `F403`. 

Just FYI here is a reference CI file from a different project using flake8:
```
name: Unit Tests

on:
  push:
    branches: [main, staging]
  pull_request:
    branches: [main, staging]

jobs:
  unit-tests:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Set up Python
        uses: actions/setup-python@v4
        with:
          python-version: '3.10'
          cache: 'pip'
      - name: Install dependencies
        run: |
          python -m pip install --upgrade pip
          pip install flake8 pytest
          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
          if [ -f requirements-dev.txt ]; then pip install -r requirements-dev.txt; fi
      - name: Lint with flake8
        run: |
          flake8 .
      - name: Test with pytest
        working-directory: ./backend
        run: |
          pytest --benchmark-skip
```
where relevant part is:
```
- name: Lint with flake8
        run: |
          flake8 .
```

The current pylint checks for the following issues in code files (which we'd still like to check for using flake8, not sure if the above config will do that already or it needs to be modified):
- import *
- imports that are unused in the code

ilnarkz:
@elisno @jwmueller  I think I can do it. Do I have to fix the errors(E302, F541, F401 etc.)? I can also create a Makefile. If this is needed.

elisno:
Thanks @ilnarkz !
Maybe start with a PR that focuses on changing the configuration files and the CI workflow file.

Not sure we'd need a Makefile at this time.

jwmueller:
Closed by: https://github.com/cleanlab/cleanlab/pull/531

Thanks for your help @ilnarkz ! 
Let us know if you'd like to help with any other open issues :)

PR #531, opened 2022-11-08
replace pylint --> flake8

ilnarkz:
[#487] replace pylint --> flake8


ilnarkz:
I didn't fix the errors. I didn't ignore anything.

setup.cfg
```diff
@@ -0,0 +1,5 @@
+[flake8]
+ignore =
```
elisno, line 2: I know this goes against the suggestion in the issue mentioned in the description, but for this PR I think it's easiest to just focus on the errors `F401,F403` (which the CI workflow is supposed to handle).

Try adding this @ilnarkz:
```suggestion
ignore =
select = F401,F403
per-file-ignores =
    cleanlab/__init__.py: F401
    cleanlab/token_classification/__init__.py: F401
    cleanlab/benchmarking/__init__.py: F401
```

elisno: The .pylintrc config had the setting:
https://github.com/cleanlab/cleanlab/blob/886e6aee6e81df463afbf9ff6e55a7d489343a39/.pylintrc#L408-L409

which doesn't exactly have a flake8-counterpart.

ilnarkz: @elisno But then flake8 will only track these errors. Is this correct?

elisno: You're right. That's the only thing at the moment that pylint is doing in our CI.

To keep the scope of this PR manageable, you should aim to replace that functionality without adding anything extra.

We can always open a new PR later for tracking more (or all) errors with flake8.

elisno:
@ilnarkz Can you fill out the CLA linked above?

ilnarkz:
@elisno  Ok, I did it.

PR #538, opened 2022-11-17
Multiannotator Active Learning Support

huiwengoh:
Adds support for active learning and ensemble functions in the multiannotator module.


cleanlab/internal/multiannotator_utils.py
```diff
@@ -134,3 +150,122 @@ def format_multiannotator_labels(labels: LabelLike) -> Tuple[pd.DataFrame, dict]
...
+    num_classes = get_num_classes(pred_probs=pred_probs)
+
+    empirical_label_distribution = np.full((len(labels_multiannotator), num_classes), np.NaN)
+    for i in range(len(labels_multiannotator)):
+        s = labels_multiannotator.iloc[i]
+        empirical_label_distribution[i, :] = value_counts(
+            s.dropna(), num_classes=num_classes
+        ) / len(s.dropna())
+
+    clipped_pred_probs = np.clip(pred_probs, a_min=1e-6, a_max=None)
```
jwmueller, line 189: note here you still have 1e-6 instead of the smaller 1e-30.

I think it would be safest to define a constant variable = 1e-30 at the top of this file and then refer to this constant variable everywhere

cleanlab/internal/multiannotator_utils.py
```diff
@@ -134,3 +150,122 @@ def format_multiannotator_labels(labels: LabelLike) -> Tuple[pd.DataFrame, dict]
...
+    soft_cross_entropy = -np.sum(
+        empirical_label_distribution * np.log(clipped_pred_probs), axis=1
+    ) / np.log(num_classes)
+
+    return soft_cross_entropy
+
+
```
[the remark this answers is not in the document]
jwmueller: Another thing to consider to allow user to exert even greater control of runtime would be whether the `find_best_temp_scaler()` search can be conducted on a random subsample of the data and pred_probs.  One probably doesn't need to use all the data to get close to the best temperature value, eg. a maximum sized subsample of num_classes*1e4 or 1e5 probably suffices.

cleanlab/multiannotator.py
```diff
@@ -92,10 +99,16 @@ def get_label_quality_multiannotator(
 
         * ``crowdlab``: an emsemble method that weighs both the annotators' labels as well as the model's prediction.
         * ``agreement``: the fraction of annotators that agree with the consensus label.
+    temp_scale : bool, default = False
+        Boolean value that specifies if the pred_probs will be temperature scaled to better match the annotators' empirical label distribution.
```
jwmueller, line 103:
```suggestion
        Boolean value that specifies whether the provided `pred_probs` should be re-calibrated to better match the annotators' empirical label distribution.
```

cleanlab/multiannotator.py
```diff
@@ -92,10 +99,16 @@ def get_label_quality_multiannotator(
 
         * ``crowdlab``: an emsemble method that weighs both the annotators' labels as well as the model's prediction.
         * ``agreement``: the fraction of annotators that agree with the consensus label.
+    temp_scale : bool, default = False
```
jwmueller, line 102: at user-facing level, can we call this arg: `calibrate_probs` or `calibrate_pred_probs` ?  That provides more flexibility in the future to change the algorithm used to perform the calibration

cleanlab/multiannotator.py
```diff
@@ -92,10 +99,16 @@ def get_label_quality_multiannotator(
...
+        Boolean value that specifies if the pred_probs will be temperature scaled to better match the annotators' empirical label distribution.
+        This is especially relevant in active learning to prevent overconfident models.
```
jwmueller, line 104:
```suggestion
        We recommend setting this to True in active learning applications, in order to prevent overconfident models from suggesting the wrong examples to collect labels for.
```

cleanlab/multiannotator.py
```diff
@@ -92,10 +99,16 @@ def get_label_quality_multiannotator(
...
     return_detailed_quality: bool, default = True
         Boolean to specify if `detailed_label_quality` is returned.
     return_annotator_stats : bool, default = True
         Boolean to specify if `annotator_stats` is returned.
+    return_weights : bool, default = False
+        Boolean to specify if `model_weight` and `annotator_weight` is returned.
+        Model and annotator weights are applicable for ``quality_method == crowdlab``, will return ``None`` for any other quality methods.
```
jwmueller, line 111: what about just throwing ValueError if return_weights is True but quality_method is not crowdlab?

cleanlab/multiannotator.py
```diff
@@ -126,6 +139,13 @@ def get_label_quality_multiannotator(
             * ``num_examples_labeled``: number of examples annotated by a given annotator.
             * ``agreement_with_consensus``: fraction of examples where a given annotator agrees with the consensus label.
             * ``worst_class``: the class that is most frequently mislabeled by a given annotator.
+
+        ``model_weight`` : float (returned if `return_weights=True`)
+            float specifying the model weight used in weighted averages (only applicable for ``quality_method == crowdlab``), ``None`` otherwise
+
+        ``annotator_weight`` : np.ndarray (returned if `return_weights=True`)
+            An array of shape ``(M,)`` where M is the number of annotators, specifying the annotator weights used in weighted averages
```
jwmueller, line 147:
```suggestion
        ``model_weight`` : float
            Only returned if `return_weights=True`, this number specifies the weight of classifier model in weighted averages used to estimate label quality (only applicable for ``quality_method == crowdlab``), ``None`` otherwise. This number is an estimate of how trustworthy the model is relative the annotators.

        ``annotator_weight`` : np.ndarray (returned if `return_weights=True`)
            An array of shape ``(M,)`` where M is the number of annotators, specifying the weight of each annotator in weighted averages used to estimate label quality.  These weights are estimates of how trustworthy each annotator is relative to the other annotators.
```

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+    """Returns label quality scores for each example and for each annotator.
+
+    This function is for multiclass classification datasets where examples have been labeled by
+    multiple annotators (not necessarily the same number of annotators per example).
+
+    It computes one consensus label for each example that best accounts for the labels chosen by each
+    annotator (and their quality), as well as a score for how confident we are that this consensus label is actually correct.
+    It also computes the the quality scores for each annotator's individual labels, and the quality of each annotator.
+    This function allows for an ensemble of models to be trained, the models' predictions and confidences will be used to compute the
+    consensus label and annotator quality.
```
jwmueller, line 338:
```suggestion
    This function is similar to :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>` but for settings where you have trained an ensemble of multiple classifier models rather than a single model.
    For multiclass classification datasets where examples have been labeled by
    multiple annotators, this function computes: a consensus label for each example, a score for how confident we are that this consensus label is actually correct, and quality scores for each annotator's individual labels as well as the quality of each annotator.
    Assuming an ensemble of multiple classifier models has been trained, all of the models' predictions and confidences are used to estimate consensus label and annotator quality in an aggregated fashion that can produce more accurate estimates than only training a single classifier model.
```

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+            * ``annotator_quality``: overall quality of a given annotator's labels, calculated using the crowdlab method`.
+            * ``num_examples_labeled``: number of examples annotated by a given annotator.
+            * ``agreement_with_consensus``: fraction of examples where a given annotator agrees with the consensus label.
+            * ``worst_class``: the class that is most frequently mislabeled by a given annotator.
+
+        ``model_weight`` : np.ndarray (returned if `return_weights=True`)
+            An array of shape ``(P,)`` where is the number of models in the ensemble, specifying the model weight used in weighted averages
+
+        ``annotator_weight`` : np.ndarray (returned if `return_weights=True`)
+            An array of shape ``(M,)`` where M is the number of annotators, specifying the annotator weights used in weighted averages
```
jwmueller, line 402: may be better to just point user to refer to the documentation of `get_label_quality_multiannotator`, except for the args which now have different meanings in this function.

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+    """
```
jwmueller, line 403:
```suggestion
    See Also
    --------
    get_label_quality_multiannotator
    """
```

cleanlab/multiannotator.py
```diff
@@ -427,6 +866,47 @@ def get_majority_vote_label(
...
+        and lastly by random selection.
+    """
+
+    if pred_probs.ndim != 3:
+        raise ValueError(
+            "pred_probs must be a 3d array containing the predicted probabilites of the ensemble models"
+        )
+
+    avg_pred_probs = np.mean(pred_probs, axis=0)
+    return get_majority_vote_label(labels_multiannotator, avg_pred_probs, verbose)
```
jwmueller, line 907: I don't think we need to provide this function? 

User can just do: 
```
avg_pred_probs = np.mean(pred_probs, axis=0)
get_majority_vote_label(labels_multiannotator, avg_pred_probs)
```

so IMO this function is as likely to confuse as it is to help (including extra functions always increases potential confusion for users of a package)

cleanlab/multiannotator.py
```diff
@@ -845,6 +1340,117 @@ def _get_post_pred_probs_and_weights(
...
+def _get_post_pred_probs_and_weights_ensemble(
+    labels_multiannotator: pd.DataFrame,
+    consensus_label: np.ndarray,
+    prior_pred_probs: np.ndarray,
+    num_annotations: np.ndarray,
+    annotator_agreement: np.ndarray,
+    quality_method: str = "crowdlab",
+    verbose: bool = True,
+) -> Tuple[np.ndarray, Any, Any]:
+    """Return the posterior predicted probabilites of each example given a specified quality method.
```
jwmueller, line 1352:
```suggestion
    """Return the posterior predicted class probabilites of each example given a specified quality method and prior predicted class probabilities from an ensemble of multiple classifier models.
```

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+            Returns overall statistics about each annotator, sorted by lowest annotator_quality first.
+            pandas DataFrame in which each row corresponds to one annotator (the row IDs correspond to annotator IDs), with columns:
+
+            * ``annotator_quality``: overall quality of a given annotator's labels, calculated using the crowdlab method`.
+            * ``num_examples_labeled``: number of examples annotated by a given annotator.
+            * ``agreement_with_consensus``: fraction of examples where a given annotator agrees with the consensus label.
+            * ``worst_class``: the class that is most frequently mislabeled by a given annotator.
+
+        ``model_weight`` : np.ndarray (returned if `return_weights=True`)
+            An array of shape ``(P,)`` where is the number of models in the ensemble, specifying the model weight used in weighted averages
```
jwmueller, line 399: restructure docstring like I advised above, but now for ensemble setting

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+        detailed_label_quality = detailed_label_quality.add_prefix("quality_annotator_")
+
+    if return_annotator_stats:
+        annotator_stats = _get_annotator_stats(
+            labels_multiannotator=labels_multiannotator,
+            pred_probs=post_pred_probs,
+            consensus_label=consensus_label,
+            num_annotations=num_annotations,
+            annotator_agreement=annotator_agreement,
+            model_weight=np.mean(model_weight),  # take mean model weight
```
jwmueller, line 512:
```suggestion
            model_weight=np.mean(model_weight),  # use average model weight when scoring annotators
```

cleanlab/multiannotator.py
```diff
@@ -845,6 +1340,117 @@ def _get_post_pred_probs_and_weights(
...
+    ----------
+    labels_multiannotator : pd.DataFrame
+        2D pandas DataFrame of multiple given labels for each example with shape ``(N, M)``,
+        where N is the number of examples and M is the number of annotators.
+        For more details, labels in the same format expected by the :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`.
+    consensus_label : np.ndarray
+        An array of shape ``(N,)`` with the consensus labels aggregated from all annotators.
+    prior_pred_probs : np.ndarray
+        An array of shape ``(N, K)`` of prior predicted probabilities, ``P(label=k|x)``, usually the out-of-sample predicted probability computed by a model.
+        For details, predicted probabilities in the same format expected by the :py:func:`get_label_quality_multiannotator <cleanlab.multiannotator.get_label_quality_multiannotator>`.
```
jwmueller, line 1364: this doesn't need to be update to reflect ensemble setting?

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+    pred_probs: np.ndarray,
+    pred_probs_unlabeled: Optional[np.ndarray] = None,
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Returns an active learning score for each example in the dataset. The score is in between 0 and 1; lower scores indicate less confidence on the examples
+    and that additional labels should be collected for those examples. For example:
+
+    * 1 - confident example, no additional labeling is required.
+    * 0 - unconfident example, it is suggested to collect more labels for this example.
+
+    This function allows for an ensemble of models to be trained and will use the aggregation of predictions from the multiple models to compute the active learning score.
```
jwmueller, line 641:
```suggestion
    """Returns an active learning score for each example in the dataset, which can be used to prioritize what data to collect additional labels for. We consider settings where one example can be labeled by multiple annotators and some examples have no labels at all so far. The score is in between 0 and 1; lower scores indicate examples whose true label we are less confident about based on the current data (the examples with lowest scores are those for which additional labels should be collected). For example:

    * 1 - confidently-labeled example, no additional labeling is required.
    * 0 - unconfident example, want to collect more labels for this example (either because it is currently unlabeled and very different than other examples, it has only received few annotations from less reliable annotators, or the annotations it has received don't agree with each other).

    This function allows for an ensemble of multiple classifier models to be trained and will aggregate predictions from the models to compute the active learning score.
```

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+
+    Parameters
+    ----------
+    labels_multiannotator : pd.DataFrame of np.ndarray
```
jwmueller, line 645:
```suggestion
    labels_multiannotator : pd.DataFrame or np.ndarray
```

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+        2D pandas DataFrame or array of multiple given labels for each example with shape ``(N, M)``,
+        where N is the number of examples and M is the number of annotators.
+        ``labels_multiannotator[n][m]`` = label for n-th example given by m-th annotator.
```
jwmueller, line 648: docstring should clarify this input has same format as expected by the `get_label_quality_multiannotator()` function.

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+
+        For a dataset with K classes, each given label must be an integer in 0, 1, ..., K-1 or ``NaN`` if this annotator did not label a particular example.
+        If you have string or other differently formatted labels, you can convert them to the proper format using :py:func:`format_multiannotator_labels <cleanlab.internal.multiannotator_utils.format_multiannotator_labels>`.
+        If pd.DataFrame, column names should correspond to each annotator's ID.
```
jwmueller, line 652: clarify examples with no labels collected thus far should not appear in this object?

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+    pred_probs : np.ndarray
+        An array of shape ``(P, N, K)`` where P is the number of models, consisting of predicted class probabilities from the ensemble models.
+        Each set of predicted probabilities with shape ``(N, K)`` is in the same format expected by the :py:func:`get_label_quality_scores <cleanlab.rank.get_label_quality_scores>`.
+    pred_probs_unlabeled : np.ndarray, optional
+        An array of shape ``(N, K)`` of predicted class probabilities from a trained classifier model for examples that have no annotator labels.
```
jwmueller, line 657:
```suggestion
        An array of shape ``(N, K)`` of predicted class probabilities from a trained classifier model for examples that have no annotated labels so far (but which we may want to label in the future).
```

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+
+    Returns
+    -------
+    active_learning_scores : np.ndarray
+        Array of shape ``(N,)`` indicating the active learning confidence scores for each example.
+    active_learning_scores_unlabeled : np.ndarray
+        Array of shape ``(N,)`` indicating the active learning confidence scores for each unlabeled example.
+        Returns an empty array if no unlabeled data is provided.
+    """
```
jwmueller, line 666: I think we are going with the name "active learning quality scores" right?
If so, make sure to stick with this name throughout everything

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+    pred_probs : np.ndarray
+        An array of shape ``(P, N, K)`` where P is the number of models, consisting of predicted class probabilities from the ensemble models.
+        Each set of predicted probabilities with shape ``(N, K)`` is in the same format expected by the :py:func:`get_label_quality_scores <cleanlab.rank.get_label_quality_scores>`.
+    pred_probs_unlabeled : np.ndarray, optional
+        An array of shape ``(N, K)`` of predicted class probabilities from a trained classifier model for examples that have no annotator labels.
+
+    Returns
+    -------
+    active_learning_scores : np.ndarray
+        Array of shape ``(N,)`` indicating the active learning confidence scores for each example.
```
jwmueller, line 662:
```suggestion
        Array of shape ``(N,)`` indicating the active learning confidence scores for each example.
        Examples with the lowest scores are those we should label next in order to maximally improve our classifier model. 
```

cleanlab/multiannotator.py
```diff
@@ -275,26 +300,439 @@ def get_label_quality_multiannotator(
...
+    active_learning_scores_unlabeled : np.ndarray
+        Array of shape ``(N,)`` indicating the active learning confidence scores for each unlabeled example.
+        Returns an empty array if no unlabeled data is provided.
```
jwmueller, line 665:
```suggestion
        Returns an empty array if no unlabeled data is provided.
        Examples with the lowest scores are those we should label next in order to maximally improve our classifier model (scores for unlabeled data are directly comparable with the `active_learning_scores` for labeled data).
```

tests/test_multiannotator.py
```diff
@@ -6,43 +6,48 @@
 from cleanlab import count
```
jwmueller, line 6: Do we have tests for the following edge cases:

1. There is only a single labeled example to start with (ok if code crashes just provide informative error msg)
2. No labeled example to start with (ok if code crashes just provide informative error msg)
3. Only a single unlabeled example
4. In the starting labeled dataset, every example has only been labeled by a single annotator

cleanlab/multiannotator.py
huiwengoh: Removed the extra function!

cleanlab/multiannotator.py
huiwengoh: Imo there is no need to throw an error, since all the other outputs will still be computed correctly, and I think it will be nice to return those to users regardless. Also, `return_weights` is by default `False`, and `quality_method` is by default crowdlab so I don't think that scenario will happen often? 
Instead of an error perhaps we can throw a warning?

jwmueller: > return_weights is by default False, and quality_method is by default crowdlab so I don't think that scenario will happen often

That was the reasoning behind my point. Why is somebody overriding return_weights when they are using a different quality method?  If they have done this, then IMO they are misunderstanding something and should get an error message.

Is there a downside to giving them the error message? Won't they just get the error message immediately and can just call the function again with return_weights = False this time?

huiwengoh: Currently the `if return_weights == True` check is done at the very end (after everything is computed and right before returning the dictionary with results), so in the current setup everything will already be computed.
It is possible to throw an error immediately, I just need a add a check at the beginning of the function before doing the computes.

jwmueller: Ya I would just add that at the beginning. Want to be suspicious of a user overriding defaults in a nonsensical manner

tests/test_multiannotator.py
huiwengoh: Some checks existed + added a few new ones to cover the ones that were missed. 
Expected behavior:
1. Code works (probably really bad results, but code will run)
2. Error message will be printed if any example contains no labels, having no labeled examples at all falls within this error catching
3. Code works
4. `get_label_quality_multiannotator` will throw an error saying 

     > Each example only has one label, collapse the labels into a 1-D array and use cleanlab.rank.get_label_quality_scores instead

cleanlab/multiannotator.py
huiwengoh: Not sure what is the cleanest way to reference the docstring here. Currently I am listing every arg individually and pasting the same reference on each line. It is a lot of repeated references but my idea was to still list every argname (thought it'll be nice to list them one by one). 
But if it is too much repetition I could remove the individual args and just reference the entire function as a whole.

cleanlab/internal/multiannotator_utils.py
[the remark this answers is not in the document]
huiwengoh: There was an inefficiency in the previous iteration of my code - we wanted to scale the labeled and unlabeled pred probs by the same temp. However because the temp scaling for the labeled pred_probs was done within `get_label_quality_multiannotator`, I was not able to get the optimal temperature and had to do the grid search a second time (on the labeled pred_probs) to scale the unlabeled pred_probs.
I've changed that now and so `temp_scale_pred_probs` is not always called right after `find_best_temp_scaler` since the optimal temp is found earlier in the code. 
In this situation making it one `calibrate_pred_probs()` function is a little confusing because I am using the same optimal temp for two sets of pred_probs, so did not implement that yet. But I agree it can be helpful and we can add that function in the future.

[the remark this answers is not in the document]
huiwengoh: Other notes:
- added optional args that allows specifying `course_search_range` and `fine_search_size`
- TODO: add a maximum sized subsample to run the grid search on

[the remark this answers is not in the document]
jwmueller: Ah ok sounds good to leave as you said. Just remember this for a future PR to add `calibrate_pred_probs()`, both for multi-annotator and regular classification!  

That future PR could even also consider extending `calibrate_pred_probs()` for multi-label classification, for which the optimal approach is to treat the pred_probs for each class separately as a binary task (one-vs-rest, since they don't need to sum to one across classes) and then apply [sklearn.isotonic](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.isotonic)) as it is utilized inside of sklearn's [CalibratedClassifierCV](https://scikit-learn.org/stable/modules/generated/sklearn.calibration.CalibratedClassifierCV.html#sklearn.calibration.CalibratedClassifierCV)

tests/test_multiannotator.py
jwmueller: For 4, can we handle this special case inside `get_active_learning_scores()` so it does not error?  That actually seems like a common scenario, ie. every example has exactly one label to begin with and we want to consider collecting more labels for these vs collecting labels for additional unlabeled examples.

jwmueller: ^ To be clear I mean just have a special `if` check to see if we are in this scenario and if so, just internally call `cleanlab.rank.get_label_quality_scores()` to get the active learning scores for the labeled examples (and penalize the resulting scores appropriately relative to the scores for the unlabeled data).

jwmueller: Let's also have a unit test that does a couple iterations of active-learning (with tiny dataset) in the standard fashion  where every example only gets one label at most. So in each iteration, just select the unlabeled example with the lowest active learning score currently and give it a label. No need to even retrain classifier for the unit test.  I just want to verify this code still generalizes to the traditional active learning setting.

cleanlab/multiannotator.py
jwmueller: Yes let's list every arg individually, specify its type, and give one phrase  description about what it is. This phrase should immediately be followed by:

> ..., in the same format as expected by the [get_label_quality_multiannotator](link) function.
 
A good example of this is the docstrings for individual label quality scores, cf:

https://docs.cleanlab.ai/stable/cleanlab/rank.html#cleanlab.rank.get_confidence_weighted_entropy_for_each_label

https://docs.cleanlab.ai/stable/cleanlab/rank.html#cleanlab.rank.get_normalized_margin_for_each_label

cleanlab/multiannotator.py
```diff
@@ -275,26 +313,425 @@ def get_label_quality_multiannotator(
...
+    pred_probs: np.ndarray,
+    *,
+    calibrate_probs: bool = False,
+    return_detailed_quality: bool = True,
+    return_annotator_stats: bool = True,
+    return_weights: bool = False,
+    verbose: bool = True,
+    label_quality_score_kwargs: dict = {},
+) -> Dict[str, Any]:
+    """Returns label quality scores for each example and for each annotator.
```
jwmueller, line 342:
```suggestion
    """Returns label quality scores for each example and for each annotator, based on predictions from an ensemble of models.
```

jwmueller: I don't think we ever want to have two different functions whose one-liner description is exactly the same.

cleanlab/multiannotator.py
```diff
@@ -275,26 +313,425 @@ def get_label_quality_multiannotator(
...
+
+    return active_learning_scores, active_learning_scores_unlabeled
+
+
+def get_active_learning_scores_ensemble(
+    labels_multiannotator: Union[pd.DataFrame, np.ndarray],
+    pred_probs: np.ndarray,
+    pred_probs_unlabeled: Optional[np.ndarray] = None,
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Returns an active learning qualityscore for each example in the dataset, which can be used to prioritize what data to collect additional labels for. We consider settings where one example can be labeled by multiple annotators and some examples have no labels at all so far. The score is in between 0 and 1; lower scores indicate examples whose true label we are less confident about based on the current data (the examples with lowest scores are those for which additional labels should be collected). For example:
```
jwmueller, line 629: Let's reason by analogy here. The docstring for this function should be structured relative to the docstring for the `get_active_learning_scores()` function as the `get_multiannotator_label_quality_ensemble()` docstring currently is written relative to the  `get_multiannotator_label_quality()` function.

In both cases, the one-liner for the `_ensemble` version should already explain what is different from the non-ensemble version.  Then the second/third sentence of docstring should always say this method is similar to the non-ensemble version (with link to non-ensemble version) except it is for settings  where multiple models have been trained.

jwmueller: For each of the args' individual description, you also want to link the user back to the non-ensemble version's docstring and keep the docstring in ensemble version more minimal

jwmueller (approved):
Stellar work!!

PR #543, opened 2022-11-24
Improve and standardize documentation in label error detection methods for classification datasets

jwmueller:
Documentation needs to start specifying the ML task more explicitly, now that package supports multiple tasks like: multi-class classification, multi-label classification, token classification, and multi-annotator datasets

### DIFF
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -34,7 +34,7 @@ import warnings
 from typing import Tuple, Union, Optional
 
 from cleanlab.typing import LabelLike
-
+from cleanlab.internal.multilabel_utils import stack_complement, get_onehot_num_classes
 from cleanlab.internal.util import (
     value_counts_fill_missing_classes,
     clip_values,
@@ -48,8 +48,6 @@ from cleanlab.internal.util import (
     is_tensorflow_dataset,
     TINY_VALUE,
 )
-from cleanlab.internal.multilabel_utils import stack_complement, get_onehot_num_classes
-
 from cleanlab.internal.latent_algebra import (
     compute_inv_noise_matrix,
     compute_py,
@@ -74,7 +72,7 @@ def num_label_issues(
 
     Parameters
     ----------
-    labels :
+    labels : np.ndarray or list
       Given class labels for each example in the dataset, some of which may be erroneous,
       in same format expected by :py:func:`filter.find_label_issues <cleanlab.filter.find_label_issues>` function.
 
@@ -177,7 +175,9 @@ def _num_label_issues_multilabel(
     return sum(issues_idx)
 
 
-def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) -> np.ndarray:
+def calibrate_confident_joint(
+    confident_joint: np.ndarray, labels: LabelLike, *, multi_label: bool = False
+) -> np.ndarray:
     """Calibrates any confident joint estimate ``P(label=i, true_label=j)`` such that
     ``np.sum(cj) == len(labels)`` and ``np.sum(cj, axis = 1) == np.bincount(labels)``.
 
@@ -199,7 +199,7 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) -> 
       If not provided, it is computed from the given (noisy) `labels` and `pred_probs`.
       If `multi_label` is True, then the `confident_joint` should be a one-vs-rest array of shape ``(K, 2, 2)``, and an array of the same shape will be returned.
 
-    labels : np.ndarray
+    labels : np.ndarray or list
       Given class labels for each example in the dataset, some of which may be erroneous,
       in same format expected by :py:func:`filter.find_label_issues <cleanlab.filter.find_label_issues>` function.
 
@@ -224,7 +224,10 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) -> 
     """
 
     if multi_label:
-        return _calibrate_confident_joint_multilabel(confident_joint, labels)
+        if not isinstance(labels, list):
+            raise TypeError("`labels` must be list when `multi_label=True`.")
+        else:
+            return _calibrate_confident_joint_multilabel(confident_joint, labels)
     else:
         num_classes = len(confident_joint)
         label_counts = value_counts_fill_missing_classes(labels, num_classes, multi_label=False)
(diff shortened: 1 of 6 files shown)


cleanlab/count.py
```diff
@@ -67,34 +69,24 @@ def num_label_issues(
...
+    estimate of number of label issues when you don't need the indices of the examples with label issues.
 
     Parameters
     ----------
     labels :
-      An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
-      Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
-      For multi-label classification where each example can belong to multiple classes,
-      refer to documentation for this argument in :py:func:`find_label_issues <cleanlab.filter.find_label_issues>` for details.
+      Given class labels for each example in the dataset, some of which may be erroneous,
```
ulya-tkch, line 78: Is it clear enough to link param documentation if that documentation comes from a different file (like here linking cleanlab.filter from cleanlab.count) without further explanation?

cleanlab/count.py
```diff
@@ -111,8 +103,8 @@ class 0, 1, ..., K-1. `pred_probs` should have been computed using 3 (or
        TL;DR: use this method to get the most accurate estimate of number of label issues when you don't need the indices of the label issues.
 
     multi_label : bool, optional
-       Refer to documentation for this argument in :py:func:`compute_confident_joint <cleanlab.count.compute_confident_joint>` with `multi_label=True` for details.
-
+      Whether dataset is for multi-label classification or multi-class classfication,
```
ulya-tkch, line 106: Does both multi-label and multi-class classification use the `multi_label=True` boolean? Seems a little confusing to mention multi-class classification under the `multi_label` explanation.

cleanlab/count.py
```diff
@@ -202,26 +194,22 @@ def calibrate_confident_joint(confident_joint, labels, *, multi_label=False) ->
       An array of shape ``(K, K)`` representing the confident joint, the matrix used for identifying label issues, which
       estimates a confident subset of the joint distribution of the noisy and true labels, ``P_{noisy label, true label}``.
       Entry ``(j, k)`` in the matrix is the number of examples confidently counted into the pair of ``(noisy label=j, true label=k)`` classes.
+      This should be in same format expected by :py:func:`filter.find_label_issues <cleanlab.filter.find_label_issues>` function.
```
ulya-tkch, line 197: Is the format listed in cleanlab.find_label_issues identical to the description here? Suggesting to either remove some of this description and link more information on the format or remove the link as it does not provide additional information.

cleanlab/count.py
ulya-tkch: One idea is to fully redefine all parameters once in a file for a big function like this one and then link them for the rest of the documentation in this file.

cleanlab/count.py
```diff
@@ -425,20 +400,12 @@ def compute_confident_joint(
     Parameters
     ----------
     labels : np.ndarray
-      An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
-      Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
-      ``len(set(labels)) == pred_probs.shape[1]`` for standard multi-class classification with single-labeled data (e.g. ``labels =  [1,0,2,1,1,0...]``).
-      For multi-label classification where each example can belong to multiple classes(e.g. ``labels = [[1,2],[1],[0],..]``),
-      your labels should instead satisfy: ``len(set(k for l in labels for k in l)) == pred_probs.shape[1])``.
+      Given class labels for each example in the dataset, some of which may be erroneous,
+      in same format expected by :py:func:`filter.find_label_issues <cleanlab.filter.find_label_issues>` function.
```
ulya-tkch, line 404: If we redefine the labels parameter in num_label_issues then links like this could be more localized.

cleanlab/count.py
jwmueller: `multi_label=True` specifies that the ML task is multi-label not multi-class. Where did you see discussion of multi-class classification under `multi_label=True`?

cleanlab/dataset.py
```diff
@@ -152,11 +152,11 @@ def find_overlapping_classes(
     Parameters
     ----------
     labels : np.ndarray, optional
-      An array of shape ``(N,)`` of noisy labels, i.e. some labels may be erroneous.
-      Elements must be in the set 0, 1, ..., K-1, where K is the number of classes.
-      All the classes (0, 1, ..., and K-1) MUST be present in ``labels``, such that
+      An array of shape ``(N,)`` of noisy labels for the dataset, i.e. some labels may be erroneous.
```
ulya-tkch, line 155: Clarify "Provides dataset-level and class-level overviews of issues in your classification dataset." for continuity?

cleanlab/count.py
ulya-tkch: The logic of "Whether dataset is for multi-label classification or multi-class classification see link" and then link starting with "If True, this is multi-label classification rather than multi-class classification" was be a bit confusing. Might be clearer to say "If false, this is regular multi-class classification. If true, see link for details."

cleanlab/count.py
jwmueller: I think it's ok to link function from different module when that is one of the main functions in the package. I imagine everybody using cleanlab knows the `find_label_issues` function. So when they are using another function, they'll be happy to hear they can just have their inputs in the same format for the famous function they already know.  I find it less ideal to ask them to refer to another function they may have never heard of, even if this function is in the same module. 

Anyway clicking on the link to the function should directly take them to the appropriate docstring in docs.cleanlab.ai, so having the function be in different file does not matter much when browsing this website.

cleanlab/count.py
jwmueller: deleted link to filter.find_label_issues and mention of it. I just included that to reassure user they don't need to change anything from that format (they may already be familiar with), but I don't think this reassurance is critical

cleanlab/dataset.py
jwmueller: good call, done

cleanlab/count.py
jwmueller: updated everywhere to follow this structure

cleanlab/filter.py
```diff
@@ -694,9 +697,9 @@ def _reduce_prune_counts(prune_count_matrix: np.ndarray, frac_noise: float = 1.0
 
 
 def find_predicted_neq_given(
-    labels: np.ndarray, pred_probs: np.ndarray, *, multi_label: bool = False
+    labels: LabelLike, pred_probs: np.ndarray, *, multi_label: bool = False
```
ulya-tkch, line 700: Change docstring to "np.ndarray or list"

ulya-tkch (approved):
Change makes it a lot clearer to understand multi-class vs multi-label and which models deal with what. LGTM!

cleanlab/filter.py
jwmueller: good catch, done

PR #546, opened 2022-11-26
Fix types to work with latest pandas-stubs

anishathalye:
The latest pandas-stubs (1.5.2.221124) changed the type of `.apply(f, ...)` such that the function `f` must return either a `Scalar` or a `Series`. This patch updates the functions to return a `Series`. Note that this has minimal impact on runtime performance, because by default, the pd.Series constructor does *not* copy the underlying data.


elisno (approved):
:shipit:

PR #547, opened 2022-11-26
Fix passing of kwargs to get_label_quality_scores

anishathalye:
The functions `_get_consensus_stats` and `_get_annotator_label_quality_score` take an argument `label_quality_score_kwargs`, a dictionary of keyword arguments to pass to `get_label_quality_scores`. When passing a `label_quality_score_kwargs` dictionary to these functions, using the unpacking operator is incorrect: that would be an extra level of unpacking. The _implementations_ of these functions will unpack the `label_quality_score_kwargs` when calling `get_label_quality_scores`. This patch fixes the issue and adds a basic regression test.

### DIFF
--- a/cleanlab/multiannotator.py
+++ b/cleanlab/multiannotator.py
@@ -158,7 +158,7 @@ def get_label_quality_multiannotator(
             consensus_label=majority_vote_label,
             quality_method=quality_method,
             verbose=verbose,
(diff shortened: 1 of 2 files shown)


elisno (approved):
Good catch!

huiwengoh:
Nice catch, thanks for fixing!

PR #548, opened 2022-11-26
Switch CI cron schedule

anishathalye:
Previously, the CI cron schedule, copy-pasted from a personal project, ran jobs at 8am UTC (3am EST or 4am EDT) on Saturday. This isn't a great time to get emails about CI breakages.

This patch switches the schedule so it runs on Monday morning (7am EST / 8am EDT).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on:
   push:
   pull_request:
   schedule:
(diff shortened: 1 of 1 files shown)


elisno (approved):
Much better day of the week!
We'll see if the time of day works out.

anishathalye:
Motivation for early morning was to have the CI results available when people start work (for US time zones).

PR #561, opened 2022-12-11
Pin Ubuntu version in CI test

anishathalye:
`ubuntu-latest` now resolves to `ubuntu-22.04` on GitHub actions, and `ubuntu-22.04` + `actions/setup-python` doesn't support Python 3.6. While we still want to support Python 3.6, we should test against it, so we have to stick with an older version of Ubuntu (that we were previously using, as `ubuntu-latest` resolved to `ubuntu-20.04`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-latest
+          - ubuntu-20.04
           - macos-latest
           - windows-latest
         python:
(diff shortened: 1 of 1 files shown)


jwmueller:
Do you think more people are running Ubuntu 0.22 or 0.20?  

I prefer to test on the version that bulk of users is running, and am willing to remove testing for python 3.6 if needed for that (it has been deprecated for a while now). What do you think?

anishathalye:
Oh I actually didn't realize that we had already removed the testing for Python 3.6. I think we should either keep testing against it, or drop support for it. Closed in favor of #562.

jwmueller:
Ya I thought more people will probably use Ubuntu 0.22 (if not now then soon) so am no longer committed to supporting Python 3.6. Will just remember to run extra tests for python 3.6 in upcoming releases and eventually will say we no longer support at the point this test fails.

PR #562, opened 2022-12-11
Add back testing Python 3.6

anishathalye:
If we want to continue supporting Python 3.6 (as we claim in the README), we should make sure we keep testing with it.


jwmueller (approved):
lgtm

PR #563, opened 2022-12-12
Remove double slash in math-mode in EMA docstring

### LINKED ISSUE #556: Docs: Fix math in EMA docstring
I may have broken the math display format in EMA label-score docstring by adding extra slashes, but the docs build was also broken the way it was before with a single slash (there was some complaint about escaping this character that I don't precisely recall).

Image

After this issue is fixed, we will want to replace the relevant section of the HTML file of the v2.2 docs in cleanlab-docs repo to update the stable documentation as well. I would be very cautious about replacing the entire HTML file and rather manually replace just the relevant section of this file.

elisno:
Fixes #556

### DIFF
--- a/cleanlab/internal/multilabel_scorer.py
+++ b/cleanlab/internal/multilabel_scorer.py
@@ -165,11 +165,11 @@ def exponential_moving_average(
 
     .. math::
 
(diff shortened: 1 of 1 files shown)


jwmueller (approved):
Interesting that you didn't get any complaints about escape characters but I did, maybe I used the wrong sphinx version...

PR #571, opened 2022-12-22
Specify Sphinx to order functions by source code order

huiwengoh:
Sphinx currently orders the functions in [docs.cleanlab.ai](docs.cleanlab.ai) by alphabetical order, setting it to sort by source code order will give us more control over the order in which the functions appear in our docs.

We might want to decide if we want to reorder some functions in our source code before merging this.


jwmueller (approved):
LGTM! I will merge and then we can just iteratively keep improving order of functions as we view the master branch docs

Issue #576, opened 2022-12-28
Drop support for python 3.6


jwmueller:
- [x] remove 3.6 from CI tests
- [x] remove all mentions of 3.6 from readme and documentation
- [ ] upgrade minimum versions of dependencies to reflect only the 3.7+ compatible versions, eg. sklearn, pandas, numpy, etc

jwmueller:
first task done by: https://github.com/cleanlab/cleanlab/pull/577

jwmueller:
second task done by: https://github.com/cleanlab/cleanlab/pull/578

PR #579, opened 2022-12-30
Drop python 3.6 support from dependencies in setup.py

sanjanag:
Closes https://github.com/cleanlab/cleanlab/issues/576

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -91,7 +91,7 @@ setup(
         "numpy>=1.20.0",
         "scikit-learn>=1.0",
         "tqdm>=4.53.0",
-        "pandas>=1.2.0",
+        "pandas>=1.1.5",
         "termcolor>=2.0.0",
     ],
 )
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,6 @@ with open(path.join(here, "README.md"), encoding="utf-8") as f:
 # Get version number and store it in __version__
 exec(open("cleanlab/version.py").read())
 
-
 setup(
     name="cleanlab",
     version=__version__,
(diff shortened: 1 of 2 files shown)


sanjanag:
Looked up release history of packages to check the python versions supported.

- Supports >=3.6.1 until [1.1.5](https://pypi.org/project/pandas/1.1.5/), and >=3.7.1 from  [1.2.0](https://pypi.org/project/pandas/1.2.0/)
- Latest [version](https://pypi.org/project/tqdm/4.64.1/) also supports >=2.7, hence did not change the version.
- Numpy 3.6 support dropped in [this](https://numpy.org/doc/stable/release/1.20.0-notes.html) release
- Supported 3.x until [1.1.0](https://pypi.org/project/termcolor/1.1.0/), supporting >=3.7 from [2.0.0](https://pypi.org/project/termcolor/2.0.0/)

jwmueller:
Any idea why pandas 1.2.0 does not support python 3.7.0 but only > 3.7.1? Should we stick with pandas >= 1.1.5 instead?

setup.py
```diff
@@ -89,10 +88,10 @@ def run(self):
...
     # https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/
     install_requires=[
-        "numpy>=1.11.3",
-        "scikit-learn>=0.18",
+        "numpy>=1.20.0",
+        "scikit-learn>=1.0",
         "tqdm>=4.53.0",
-        "pandas>=1.0.0",
-        "termcolor>=1.1.0",
+        "pandas>=1.2.0",
```
jwmueller, line 94:
```suggestion
        "pandas>=1.1.5",
```

jwmueller: Should we do this ^ instead?  Since pandas 1.2.0 does not support python 3.7.0 but only > 3.7.1; I'm not sure why that is?

jwmueller (approved):
Lgtm

PR #583, opened 2023-01-04
add maximum line length

cgnorthcutt:


cgnorthcutt:
need to include this because the current contributing guidelines suggest users 'generally follow [PEP-8 ](https://peps.python.org/pep-0008/#maximum-line-length)' which is max line length of 79 chars.

jwmueller (approved):
let's hold off on merging PRs as something funky has gone wrong with codecov that team is trying to resolve

cgnorthcutt:
@jwmueller merge when ready

PR #590, opened 2023-01-06
Remove temp scaling from single annotator case

huiwengoh:
There is no per-example empirical annotator label distribution in the single-label case, hence there is no need (+ does not make much sense) to temperature scale model pred probs based on that

### DIFF
--- a/cleanlab/multiannotator.py
+++ b/cleanlab/multiannotator.py
@@ -587,7 +587,7 @@ def get_active_learning_scores(
 
     # if all examples are only labeled by a single annotator
     if labels_multiannotator.apply(lambda s: len(s.dropna()) == 1, axis=1).all():
-        optimal_temp = 1.0  # do not temp scale for single annotator case, defined to be used later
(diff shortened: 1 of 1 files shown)


cleanlab/multiannotator.py
```diff
@@ -585,11 +585,10 @@ def get_active_learning_scores(
 
     num_classes = get_num_classes(pred_probs=pred_probs)
 
-    optimal_temp = find_best_temp_scaler(labels_multiannotator, pred_probs)
-    pred_probs = temp_scale_pred_probs(pred_probs, optimal_temp)
-
     # if all examples are only labeled by a single annotator
     if labels_multiannotator.apply(lambda s: len(s.dropna()) == 1, axis=1).all():
+        optimal_temp = 1.0  # do not temp scale for single annotator case, defined to be used later
```
jwmueller, line 590:
```suggestion
        optimal_temp = 1.0  # do not temp scale for single annotator case, temperature is defined here for later use
```

jwmueller (approved):
LGTM, made minor clarification to the comment's language

PR #584, opened 2023-01-04
Revamp text tutorial

huiwengoh:
- Updated text tutorial to remove scikeras dependency and use cleanlab's `KerasWrapperSequential` instead
- Text tutorial now demonstrates finding label issues using `CleanLearning` + how to train a model given the issues found
- Add numpy support to cleanlab's Keras wrappers

PR #593, opened 2023-01-06
Update docs dependencies

huiwengoh:
Misc cleanup of doc dependencies after text tutorial revamp in https://github.com/cleanlab/cleanlab/pull/584

note: since we are no longer pinning sklearn version, it does not need to be specified here anymore because it is installed when installing cleanlab


jwmueller (approved):
Thanks for remembering!

PR #595, opened 2023-01-08
Add two methods to filter.find_label_issues

cgnorthcutt:
Base code to add two new (the methods used in the pervasive label errors paper and used in labelerrors.com) to `filter.find_label_issues`.

TODOs are list. Other TODOs:
* update docs to include these new names anywhere that all the find_label_issues `filter_by` methods are listed (e.g. in `CleanLearning.find_label_issues`)
* Add tests
* test this thoroughly

I have not tested this code or even tried it (I haven't even checked if its runnable). I wrote entirely from memory just to get a draft up. Please take from here.

### DIFF
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -181,7 +181,7 @@ def _num_label_issues_multilabel(
         pred_probs=pred_probs,
         confident_joint=confident_joint,
         multi_label=True,
-        filter_by="confident_learning",
+        filter_by="confident_learning",  # specified to match num_label_issues
     )
     return sum(issues_idx)
 
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -169,12 +169,19 @@ def _num_label_issues_multilabel(
     -------
     num_issues : int
        The estimated number of examples with label issues in the multi-label dataset.
+
+    Note: We set the filter_by method as 'confident_learning' to match the non-multilabel case
+    (analog to the off_diagonal estimation method)
     """
 
     from cleanlab.filter import find_label_issues
 
     issues_idx = find_label_issues(
-        labels=labels, pred_probs=pred_probs, confident_joint=confident_joint, multi_label=True
+        labels=labels,
+        pred_probs=pred_probs,
+        confident_joint=confident_joint,
+        multi_label=True,
+        filter_by="confident_learning",
     )
     return sum(issues_idx)
 
--- a/cleanlab/filter.py
+++ b/cleanlab/filter.py
@@ -32,10 +32,8 @@ import warnings
 from typing import Any, Dict, List, Optional, Tuple, Union
 from functools import reduce
 
-from cleanlab.count import calibrate_confident_joint
-from cleanlab.rank import (
-    order_label_issues,
-)
+from cleanlab.count import calibrate_confident_joint, num_label_issues
+from cleanlab.rank import order_label_issues, get_label_quality_scores
 import cleanlab.internal.multilabel_scorer as ml_scorer
 from cleanlab.internal.validation import assert_valid_inputs
 from cleanlab.internal.util import (
@@ -132,13 +130,13 @@ def find_label_issues(
     filter_by : {'prune_by_class', 'prune_by_noise_rate', 'both', 'confident_learning', 'predicted_neq_given'}, default='prune_by_noise_rate'
       Method to determine which examples are flagged as having label issue, so you can filter/prune them from the dataset. Options:
 
-      - ``'low_normalized_margin'``: filters the examples with *smallest* normalized margin label quality score. The number of issues returned matches :py:func:`count.num_label_issues <cleanlab.count.num_label_issues>`.
-      - ``'low_self_confidence'``: filters the examples with *smallest* self confidence label quality score. The number of issues returned matches :py:func:`count.num_label_issues <cleanlab.count.num_label_issues>`.
       - ``'prune_by_noise_rate'``: filters examples with *high probability* of being mislabeled for every non-diagonal in the confident joint (see `prune_counts_matrix` in `filter.py`). These are the examples where (with high confidence) the given label is unlikely to match the predicted label for the example.
       - ``'prune_by_class'``: filters the examples with *smallest probability* of belonging to their given class label for every class.
       - ``'both'``: filters only those examples that would be filtered by both ``'prune_by_noise_rate'`` and ``'prune_by_class'``.
       - ``'confident_learning'``: filters the examples counted as part of the off-diagonals of the confident joint. These are the examples that are confidently predicted to be a different label than their given label.
       - ``'predicted_neq_given'``: filters examples for which the predicted class (i.e. argmax of the predicted probabilities) does not match the given label.
+      - ``'low_normalized_margin'``: filters the examples with *smallest* normalized margin label quality score. The number of issues returned matches :py:func:`count.num_label_issues <cleanlab.count.num_label_issues>`.
+      - ``'low_self_confidence'``: filters the examples with *smallest* self confidence label quality score. The number of issues returned matches :py:func:`count.num_label_issues <cleanlab.count.num_label_issues>`.
 
     multi_label : bool, optional
       If ``True``, labels should be an iterable (e.g. list) of iterables, containing a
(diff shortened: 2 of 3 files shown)


cleanlab/filter.py
```diff
@@ -132,6 +132,8 @@ class 0, 1, ..., K-1.
     filter_by : {'prune_by_class', 'prune_by_noise_rate', 'both', 'confident_learning', 'predicted_neq_given'}, default='prune_by_noise_rate'
       Method to determine which examples are flagged as having label issue, so you can filter/prune them from the dataset. Options:
 
+      - ``'low_normalized_margin'``: filters the examples with *smallest* normalized margin label quality score. The number of issues returned matches :py:func:`count.num_label_issues <cleanlab.count.num_label_issues>`.
+      - ``'low_self_confidence'``: filters the examples with *smallest* self confidence label quality score. The number of issues returned matches :py:func:`count.num_label_issues <cleanlab.count.num_label_issues>`.
```
jwmueller, line 136: Y/N: Do you want to allow somebody to do: 

```
find_label_issues(filter_by="low_self_confidence", return_indices_ranked_by="normalized_margin")
```
?

cleanlab/filter.py
```diff
@@ -290,6 +294,28 @@ class 0, 1, ..., K-1.
...
+    if filter_by in ["low_normalized_margin", "low_self_confidence"]:
+        from cleanlab.rank import get_label_quality_scores
+        from cleanlab.count import num_label_issues
+
+        # TODO: consider setting adjust_pred_probs to true based on benchmarks (or adding it kwargs, or ignoring and leaving as false by default)
+        scores = get_label_quality_scores(
+            labels,
+            pred_probs,
+            method=filter_by[7:],
+            adjust_pred_probs=False,
```
jwmueller, line 307: should `adjust_pred_probs` value come from `rank_by_kwargs["adjust_pred_probs"]` if that was specified by user? Or always restrict to False?

cleanlab/filter.py
jwmueller: Currently if somebody does: `find_label_issues(filter_by="low_self_confidence", return_indices_ranked_by="self_confidence")` it seems like there'll be some redundant computation, but I guess we don't care too much because the extra return_indices_ranked_by computation is only over the flagged label issues.

cleanlab/filter.py
cgnorthcutt: I'm not sure what the benchmarks suggested in terms of it being on by default for these methods. If no information, would be nice for it to be added to rank_by_kwargs["adjust_pred_probs"] with a default value of `False`

cleanlab/filter.py
cgnorthcutt: > Y/N: Do you want to allow somebody to do:
> 
> ```
> find_label_issues(filter_by="low_self_confidence", return_indices_ranked_by="normalized_margin")
> ```
> 
> ?


Yes. It returns a difference answer and both answers are reasonable that a user might compute them (they might trust filtering done by `self_confidence` more and ranking done by `norm margin` more, so we should keep it this way and let them support both.

The other benefit of keeping the API this way is now the way it works for these two methods is the same as how it works for all 7 methods -- consistency is better.

Should def keep as this API.

cleanlab/filter.py
jwmueller: I think having the default adjust_pred_probs be False is the way to go. I am more asking whether we want to support allow the user to filter_by low_self_confidence but with self_confidence being computed with adjusted pred_probs. There is no way to do this in your current PR AFAICT, one can only find the label issue mask with unadjusted self_confidence/normalized_margin and then choose to later use adjusted pred_probs when ranking the datapoints with label issues

cgnorthcutt: passing this off to you but let me know if you still want me inolved @jwmueller

huiwengoh:
I've ran tests with missing classes, CleanLearning and those work. 

One remaining inconsistency if the number of issues returned when using these two new methods in the multi-label setting, ie. `sum(filter.find_label_issues(...)) != count.num_label_issues(...)`. 

This happens because the `count.num_label_issues` method for multi-label sums the number of label issues identified by `filter.find_label_issues` using default parameters (ref [here](https://github.com/cleanlab/cleanlab/blob/5c1ba64bea3892709a7f558c04725c468e09974f/cleanlab/count.py#L212)), which is `filter_by="prune_by_noise_rate"`, and in `count.num_label_issues` for multi-label the label issues are computed slightly differently [here](https://github.com/cleanlab/cleanlab/blob/5c1ba64bea3892709a7f558c04725c468e09974f/cleanlab/filter.py#L300). 

I'm not sure if we want to make this number consistent? 
If that is a goal, the "logical" way to do this (following the logic of the method description, ie. count number issues first then return lowest scoring labels) is to add an `if` clause to `find_label_issues` so that if `filter_by == "low_self_confidence" or "low_ normalized_margin"` then it does not call the multi-label helper functions but instead uses the count from `num_label_issues` like the non-multiclass case. However, `num_label_issues` for the multi-label case is also cyclical (and also in current implementation uses the default `filter_by="prune_by_noise_rate"` to determine number of issues) so that is quite strange in my opinion.

I'm not the most familiar with the multi-label methods so I am not sure what the best way to approach this. 
@aditya1503 @elisno @jwmueller would y'all have any insights on this?

cleanlab/filter.py
```diff
@@ -290,6 +294,28 @@ class 0, 1, ..., K-1.
             multi_label=multi_label,
             return_indices_of_off_diagonals=True,
         )
+    # TODO: (delete once read and/or make a comment) -- We don't re-use confident_joint inside this method in num_label_issues(confident_joint=confident_joint), bcconfident_joint is calibrated and by default the method uses an uncalibrated confident_joint. Delete this todo once confirmed this is correct.
```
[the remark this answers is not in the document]
jwmueller: Can delete the comment

cleanlab/filter.py
```diff
@@ -137,6 +135,8 @@ class 0, 1, ..., K-1.
       - ``'both'``: filters only those examples that would be filtered by both ``'prune_by_noise_rate'`` and ``'prune_by_class'``.
       - ``'confident_learning'``: filters the examples counted as part of the off-diagonals of the confident joint. These are the examples that are confidently predicted to be a different label than their given label.
       - ``'predicted_neq_given'``: filters examples for which the predicted class (i.e. argmax of the predicted probabilities) does not match the given label.
+      - ``'low_normalized_margin'``: filters the examples with *smallest* normalized margin label quality score. The number of issues returned matches :py:func:`count.num_label_issues <cleanlab.count.num_label_issues>`.
+      - ``'low_self_confidence'``: filters the examples with *smallest* self confidence label quality score. The number of issues returned matches :py:func:`count.num_label_issues <cleanlab.count.num_label_issues>`.
```
jwmueller, line 139: Should add these two options to the list of list of filter_by options above.

cleanlab/count.py
```diff
@@ -169,12 +169,19 @@ def _num_label_issues_multilabel(
...
 
     from cleanlab.filter import find_label_issues
 
     issues_idx = find_label_issues(
-        labels=labels, pred_probs=pred_probs, confident_joint=confident_joint, multi_label=True
+        labels=labels,
+        pred_probs=pred_probs,
+        confident_joint=confident_joint,
+        multi_label=True,
+        filter_by="confident_learning",
```
jwmueller, line 184:
```suggestion
        filter_by="confident_learning",  # specified to match num_label_issues
```

cleanlab/filter.py
```diff
@@ -290,6 +302,26 @@ class 0, 1, ..., K-1.
             multi_label=multi_label,
             return_indices_of_off_diagonals=True,
         )
+
+    if filter_by in ["low_normalized_margin", "low_self_confidence"]:
+        # TODO: consider setting adjust_pred_probs to true based on benchmarks (or adding it kwargs, or ignoring and leaving as false by default)
+        scores = get_label_quality_scores(
```
jwmueller, line 308: Can you add unit test to ensure this gets covered?

cleanlab/filter.py
```diff
@@ -290,6 +302,26 @@ class 0, 1, ..., K-1.
...
+            pred_probs,
+            method=filter_by[4:],
+            adjust_pred_probs=False,
+        )
+        num_errors = num_label_issues(
+            labels, pred_probs, multi_label=multi_label  # TODO: Check usage of multilabel
+        )
+        # Find label issues O(nlogn) solution (mapped to boolean mask later in the method)
+        cl_error_indices = np.argsort(scores)[:num_errors]
+        # TODO: delete the comment below (or keep anything you want in docs/comments)
```
jwmueller, line 319:
```suggestion
```

cleanlab/filter.py
```diff
@@ -290,6 +302,26 @@ class 0, 1, ..., K-1.
...
+    if filter_by in ["low_normalized_margin", "low_self_confidence"]:
+        # TODO: consider setting adjust_pred_probs to true based on benchmarks (or adding it kwargs, or ignoring and leaving as false by default)
+        scores = get_label_quality_scores(
+            labels,
+            pred_probs,
+            method=filter_by[4:],
+            adjust_pred_probs=False,
+        )
+        num_errors = num_label_issues(
+            labels, pred_probs, multi_label=multi_label  # TODO: Check usage of multilabel
```
jwmueller, line 315: can we delete this comment now?

jwmueller (changes requested):
looks good overall, just requesting minor changes on things like comments

cleanlab/filter.py
huiwengoh: I don't actually know why codecov thinks this line is not covered, I have added multiple unittests that should call this line (ref [here](https://github.com/cleanlab/cleanlab/pull/595/files/e83d7ea4b1bd19ae626b0134db598cce5d2c41e2#diff-817ca800504a08d514a20253405df1f597ee171107cc9e47e564fdacd68f0473R1014)), also the lines below are covered (and they must access the variable `scores`, ie. cannot be covered unless this line is called), so I'm a little confused.

jwmueller: no worries lets ignore this then

jwmueller (approved):
LGTM

PR #596, opened 2023-01-09
10x speedup in find_label_issues on linux via better multiprocessing

clu0:
see details here: https://good-jumbo-503.notion.site/find_label_issue-multiprocessing-0af61f20fdc04780908b2f0e8b2a38b8


cleanlab/filter.py
```diff
@@ -26,11 +26,11 @@
 import numpy as np
```
jwmueller, line 26: Could you add one more unit test that runs find_label_issues with n_jobs = 1 vs 2 vs None and verifies the returned label issues match?  This is sort of being minimally tested here:
https://github.com/cleanlab/cleanlab/blob/master/tests/test_filter_count.py#L866

but I think another test of this would be good. You could easily add it here in this method:
https://github.com/cleanlab/cleanlab/blob/master/tests/test_filter_count.py#L698

cleanlab/filter.py
```diff
@@ -310,83 +314,63 @@ class 0, 1, ..., K-1.
...
-            args = (
-                labels,
-                label_counts,
-                prune_count_matrix,
-                pred_probs,
-                multi_label,
-                min_examples_per_class,
-            )
+        chunksize = max(1, K // n_jobs)
+        os_name = platform.system()
```
jwmueller, line 318: could you add brief comment here explaining what will differ based on operating system type?

cleanlab/filter.py
```diff
@@ -310,83 +314,63 @@ class 0, 1, ..., K-1.
...
-                    pred_probs.shape,
-                    multi_label,
-                    min_examples_per_class,
-                ),
-            ) as p:
+        if n_jobs > 1:
+            with multiprocessing.Pool(n_jobs) as p:
                 if verbose:  # pragma: no cover
-                    print("Parallel processing label issues by class.")
+                    print("Parallel processing label issues by noise rate.")
```
jwmueller, line 336: this should still read "by class" no?

cleanlab/filter.py
```diff
@@ -902,41 +895,31 @@ def _prune_by_class(k: int, args=None) -> np.ndarray:
...
-        rank = np.partition(class_probs[label_filter], num_issues)[num_issues]
-        return label_filter & (class_probs < rank)
+        rank = np.partition(class_probs, num_issues)[num_issues]
+        return class_probs < rank
     else:
         warnings.warn(
             f"May not flag all label issues in class: {k}, it has too few examples (see argument: `min_examples_per_class`)"
         )
-        return np.zeros(len(labels), dtype=bool)
+        return np.zeros(label_counts, dtype=bool)
```
jwmueller, line 918: try to extend existing unit test to cover this line of code as well, thanks!

tests/test_filter_count.py
```diff
@@ -875,6 +875,42 @@ def test_missing_classes():
         assert all(filter.find_label_issues(labels, pred_probs, filter_by=fb) == issues)
 
 
+@pytest.mark.filterwarnings("ignore:WARNING!")
+def test_find_label_issues_match_multiprocessing():
+    # minimal version of this test was run in test_missing_classes
+    # here testing with larger input matrices
+
+    # test with ground truth labels:
+    n = 200000
```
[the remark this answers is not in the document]
clu0: Happy to change to n=5000, but when I was testing the v2.2.0 version, I only consistently got different outputs from single vs multi with this input size.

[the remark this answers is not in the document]
jwmueller: oh hm how long does this unit test take with n=200000?
I think the runtime for this test function alone needs to be capped at 30-45sec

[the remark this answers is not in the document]
clu0: yeah with n=200000 and all the options the test would run (at least on my mac) for over 10 mins, so maybe n=5000 for now

cleanlab/filter.py
clu0: prune_by_class seems to have some bugs, will fix

tests/test_filter_count.py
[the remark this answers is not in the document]
jwmueller: you can just include comment there that previously some bugs were observed only with larger sample-sizes like  n=200000

cleanlab/filter.py
clu0: i realized that the previous code has another bug, but this one should rarely matter for real datasets: the prune_by_noise_rate and prune_by_count use np.partition to get a threshold, and then prunes with probs < threshold. This incurs bugs when there are more than one example that have probs exactly equal to the threshold value.

I changed the new code to using np.sort, which avoids this bug and is only very marginally slower, and the time difference is trivial compared to other time usages. Still doing other tests to make sure the outputs are correct

jwmueller: @cgnorthcutt just FYI in case you want to have second look

tests/test_filter_count.py
```diff
@@ -875,6 +875,52 @@ def test_missing_classes():
         assert all(filter.find_label_issues(labels, pred_probs, filter_by=fb) == issues)
 
 
+@pytest.mark.filterwarnings("ignore:WARNING!")
+def test_find_label_issues_match_multiprocessing():
+    # minimal version of this test was run in test_missing_classes
+    # here testing with larger input matrices
+
+    # test with ground truth labels:
+    n = 5000
```
jwmueller, line 884:
```suggestion
    n = 5000  # consider replacing this with larger value
    # some past bugs observed only with larger sample-sizes like n=200000
```

tests/test_filter_count.py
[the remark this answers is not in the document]
jwmueller: added comment for you in a suggestion

tests/test_filter_count.py
clu0: added this

cleanlab/filter.py
clu0: I put some logs about this in the notion doc. Some differences between the old and new versions do seem to be caused by this change, otherwise the new version seems correct from the test cases.

jwmueller: I suspect so too, but @cgnorthcutt  would know forsure

cgnorthcutt: @clu0 is right. this is the correct fix.

I actually just discovered this same issue in my PR last week @clu0 @jwmueller  here:  https://github.com/cleanlab/cleanlab/blob/febcaa9375e0205c6f52c48c1295b02a348b9280/cleanlab/filter.py#L315

basically anything that depends on pred probs downstream can have this issue so I've also been switching my code to sort instead of using the O(n) algorithm.

looks good @clu0 !

jwmueller (approved):
LGTM, awesome work on this!!

Do you still have any pending changes or should I merge this?

clu0:
> 
switched to counting physical cores for n_jobs, otherwise unchanged

cleanlab/filter.py
```diff
@@ -59,6 +59,21 @@
     w = """To see estimated completion times for methods in cleanlab.filter, "pip install tqdm"."""
     warnings.warn(w)
 
+# psutil used to count physical cores for multiprocessing
```
jwmueller, line 62: Thought you would add the change to count physical cores in separate PR, but happy to just take it from here in this one!

A few changes I'd request:

- [ ] Delete the warning here if `not psutil_exists`. I know there is an analogous warning for tqdm, but that's actually one of our dependencies, so that code is actually  unnecessary now.
- [ ] instead move this warning to be a print statement that occurs after the `if psutil_exists:` nested block inside of `find_label_issues()`.
- [ ] The message can read: """To default to the number of physical cores for multiprocessing in `find_label_issues()`, please: "pip install psutil"."""`
- [ ] Add psutil to list of dependencies in: https://github.com/cleanlab/cleanlab/blob/master/requirements-dev.txt

cleanlab/filter.py
```diff
@@ -59,6 +59,21 @@
...
+# Module not neccessary, because can always fall back to logical cores
+try:
+    import psutil
+
+    psutil_exists = True
+except ImportError as e:
+    psutil_exists = False
```
jwmueller, line 69:
```suggestion
except ImportError as e:  # pragma: no cover
    psutil_exists = False
```

tests/test_filter_count.py
```diff
@@ -875,6 +875,56 @@ def test_missing_classes():
...
+    labels[(n // 2) :] = 0
+    pred_probs = np.zeros((n, 4))
+    pred_probs[:, 0] = 0.95
+    pred_probs[:, 1] = 0.05
+    pred_probs[0, 0] = 0.94
+    pred_probs[0, 1] = 0.06
+    ground_truth = np.ones(n, dtype=bool)
+    ground_truth[(n // 2) :] = False
+    ground_truth[0] = False  # leave one example for min_example_per_class
+    issues = filter.find_label_issues(labels, pred_probs)
```
jwmueller, line 897:
```suggestion
    # TODO: consider also testing this line without psutil installed
    issues = filter.find_label_issues(labels, pred_probs)
```

cleanlab/filter.py
```diff
@@ -248,8 +261,24 @@ class 0, 1, ..., K-1.
...
+    # even for faily large input arrays, so we default to n_jobs=1 in this case
+    os_name = platform.system()
     if n_jobs is None:
-        n_jobs = multiprocessing.cpu_count()
+        if multi_label and os_name != "Linux":
+            n_jobs = 1
+        else:
+            if psutil_exists:
+                n_jobs = psutil.cpu_count(logical=False)  # physical cores
+            else:
```
jwmueller, line 273:
```suggestion
            else:  # pragma: no cover
```

cleanlab/filter.py
jwmueller: thanks for fixing these!

cleanlab/filter.py
```diff
@@ -248,8 +261,24 @@ class 0, 1, ..., K-1.
...
+                print(
+                    """To default to the number of physical cores for multiprocessing in find_label_issues(), please: "pip install psutil"."""
+                )
```
jwmueller, line 276:
```suggestion
                )  # pragma: no cover
```

PR #597, opened 2023-01-09
Remove temp scaling for single-label case in ensemble method

huiwengoh:
The counterpart of https://github.com/cleanlab/cleanlab/pull/590 for the ensemble active learning method (was missed out in the last PR)


jwmueller (approved):
Good catch!!

PR #598, opened 2023-01-10
Adding type hints for cleanlab/filter

unna97:


cleanlab/filter.py
```diff
@@ -60,21 +62,23 @@
...
     frac_noise: float = 1.0,
     num_to_remove_per_class: Optional[int] = None,
-    min_examples_per_class=1,
-    confident_joint: Optional[np.ndarray] = None,
+    min_examples_per_class: int=1,
+    confident_joint: Optional[npt.NDArray["np.floating[T]"]] = None,
     n_jobs: Optional[int] = None,
     verbose: bool = False,
-) -> np.ndarray:
+) -> npt.NDArray["np.floating[T]"]:
```
elisno, line 81: I'd take a closer look at the docstring for the return value of `find_label_issues`.
It can be an array of boolean masks or integer indices.

unna97: Yes, I am really sorry, I wasn't supposed to mark you for review just yet.

elisno: No worries 👍, this was just something I noticed when you opened the draft PR.

cleanlab/filter.py
```diff
@@ -298,16 +298,16 @@ class 0, 1, ..., K-1.
         # Create `prune_count_matrix` with the number of examples to remove in each class and
         # leave at least min_examples_per_class examples per class.
         # `prune_count_matrix` is transposed relative to the confident_joint.
-        prune_count_matrix = _keep_at_least_n_per_class(
+        prune_count_matrix:npt.NDArray[np.int_] = _keep_at_least_n_per_class(
```
unna97, line 301: The prune_count_matrix changes the data type (from an array of int to float). Should I declare it as a float? Although the function here returns int explicitly.

cleanlab/filter.py
```diff
@@ -427,18 +427,19 @@ class 0, 1, ..., K-1.
...
     frac_noise: float = 1.0,
     num_to_remove_per_class: Optional[int] = None,
-    min_examples_per_class=1,
+    min_examples_per_class: int=1,
     confident_joint: Optional[np.ndarray] = None,
     n_jobs: Optional[int] = None,
     verbose: bool = False,
-) -> np.ndarray:
+) -> npt.NDArray[Union[np.bool_, np.int_]]:
+    #TODO: add docstring in the same format as other functions with args and returns
```
unna97, line 442: Can I add to-dos wherever the doc doesn't have the same format as the other functions?

cleanlab/filter.py
elisno: I'll have to take a better look later today.

cleanlab/filter.py
elisno: Are you referring to the "shape"-format?

unna97: The format of doc itself.

elisno: Ohh right, now I see what you mean! Thank you for pointing this out!

That's not related to this PR, so you shouldn't add this particular to-do in this PR.
Feel free to [open an issue](https://github.com/cleanlab/cleanlab/issues/new/choose)!
If you find other instances, you can add them to the same issue.

We want to address these kinds of docstrings in a separate PR.

cleanlab/filter.py
unna97: @elisno  did you have a chance to look at this?

elisno: I don't see where the datatype should change in this function. I think this annotation is fine for now.

elisno:
I've updated some of the type-hints to pass more of the strict checks (not 100% just yet).

CI is now failing on "[Test without optional dependencies and with minimum compatible versions of dependencies](https://github.com/cleanlab/cleanlab/actions/runs/5414776388/jobs/9842318709#logs)", where numpy.typing is not recognized.

PR #600, opened 2023-01-19
pylint for python 3.5+ tweaks

marksmayo:
sorted imports with isort
removed unnecessary f-string
reordered imports
converted some strings to f-strings
added trailing commas

PR #601, opened 2023-01-19
Create new cleanlab.models module

huiwengoh:
Migrated `fasttext.py` and `keras.py` to the newly created `cleanlab.models` modules from the existing `cleanlab.experimental` module


tests/test_frameworks.py
```diff
@@ -40,14 +41,22 @@
...
 def python_version_ok():  # tensorflow and torch do not play nice with older Python
     version = sys.version_info
     return (version.major >= 3) and (version.minor >= 7)
 
 
+def run_fasttext_test():
+    # run test only if os enviroment is set of true and os is not Windows
+    if os.environ.get("TEST_FASTTEXT") == "true" and os.name != "nt":
+        return True
+    return False
```
jwmueller, line 57:
```suggestion
    return (os.environ.get("TEST_FASTTEXT") == "true" and os.name != "nt")
```

tests/test_frameworks.py
```diff
@@ -286,3 +295,48 @@ def forward(self, X):
     cl = CleanLearning(net)
     cl.fit(dataset, data["y"], clf_kwargs={"epochs": 2})
     pred_probs = cl.predict(dataset)
+
+
+# test fasttext
```
jwmueller, line 300:
```suggestion
# test fasttext only if not on windows and environment variable TEST_FASTTEXT has been set to "true"
```

tests/test_frameworks.py
```diff
@@ -286,3 +295,48 @@ def forward(self, X):
...
+    "not run_fasttext_test()", reason="fasttext is not easily pip install-able on windows"
+)
+def test_fasttext():
+    from cleanlab.models.fasttext import FastTextClassifier, data_loader
+
+    dir = "tests/fasttext_data"
+    if not os.path.isdir(dir):
+        os.makedirs(dir)
+
+    if not os.path.isfile("tests/fasttext_data/tweets_train.txt"):
```
jwmueller, line 311: how long does it take to run this dataset? Any way to use a smaller version of the dataset so that the test will run quicker?

cleanlab/models/keras.py
```diff
@@ -190,7 +196,9 @@ def fit(self, X, y=None, **kwargs):
         self.net.fit(X, **kwargs)
 
     def predict_proba(self, X, *, apply_softmax=True, **kwargs):
-        """Set extra argument `apply_softmax` to True to indicate your network only outputs logits not probabilities."""
+        """Predict class probabilities ``P(true label=k)`` using the wrapped Keras classifier.
```
jwmueller, line 199: why are we saying P(true label ) here?

More generally why isn't first sentence from all these docstrings just identical to say the first sentence of the analogous docstring from sklearn?

cleanlab/models/keras.py
```diff
@@ -121,10 +124,12 @@ def predict_proba(self, X, *, apply_softmax=True, **kwargs):
         return pred_probs
 
     def predict(self, X, **kwargs):
+        """Predict class labels using the wrapped Keras classifier."""
```
jwmueller, line 127: For all `predict()` and `predict_proba()` methods, should clarify that the new data `X` must be in same format as the original `X` provided to `fit()`?  

I assume user cannot train with TF dataset and then switch to numpy array for prediction

tests/test_frameworks.py
elisno: You can omit the parentheses. That would be in line with changes in https://github.com/cleanlab/cleanlab/pull/600.

jwmueller: thanks omitted

cleanlab/models/keras.py
```diff
@@ -82,6 +82,7 @@ def __init__(
         self.net = None
 
     def get_params(self, deep=True):
+        """Returns the parameters of the Keras classifier."""
```
jwmueller, line 85: [nit] should we rewrite "Keras classifier" -> "Keras model" everywhere here to make it clearer these are the exact same parameters/methods as for a standard Keras model?  (I'm not sure if Keras docs ever call their model a "classifier" but if so then this seems ok as is).

tests/test_frameworks.py
huiwengoh: Including the download, this unittest takes about 5 or 6 seconds. I can make the dataset smaller but I'm not sure if it will cut down the time by much?

cleanlab/models/keras.py
huiwengoh: Technically this is doable but I don't think we want users doing it, will add a statement to clarify

cleanlab/models/keras.py
huiwengoh: Keras model is more commonly used, will change accordingly

cleanlab/models/keras.py
huiwengoh: The `P(true label = k)` part was taken from the [CleanLearning docstrings](https://github.com/cleanlab/cleanlab/blob/master/cleanlab/classification.py#L609), I can change it to make it more similar to sklearn. 

Also if the `P(true label = k)` part is confusing should we also edit it for the CleanLearning documentation?

jwmueller: Ah there is a difference in that CleanLearning is explicitly intended to estimate `P(true label = k)` rather than `P(given label = k)` but most models will just estimate the latter.

So here we should just imitate a combination of sklearn & keras docstrings, since this module is not about training on cleaned data like CleanLearning

cleanlab/models/keras.py
jwmueller: Yep I think it's ok not to add too much. But isn't there currently no documentation for `X` in the predict() methods?
I was just suggesting something minimal like:

> X : data in the same format as the original \`X\` provided to \`\`fit()\`\`

tests/test_frameworks.py
jwmueller: 5-6 sec is fine! I wasn't sure.
Could you add a print statement that clarifies how developer can skip this unit test (via env variable) in the case where the download fails (eg. because no internet connection)

cleanlab/models/keras.py
```diff
@@ -87,8 +89,15 @@ def get_params(self, deep=True):
             "model": self.model,
             "model_kwargs": self.model_kwargs,
             "compile_kwargs": self.compile_kwargs,
+            "params": self.params,
         }
 
+    def set_params(self, **params):
+        """Set the parameters of the Keras model."""
+        for key, value in params.items():
+            self.params[key] = value
```
elisno, line 98: What's wrong with `update`?

```suggestion
        self.params.update(params)
```

cleanlab/models/keras.py
```diff
@@ -75,10 +75,12 @@ def __init__(
         compile_kwargs: dict = {
             "loss": tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True)
         },
+        params: dict = {},
```
elisno, line 78: Safer to use `None` as a default.
```suggestion
        params: Optional[dict] = None,
```

Then set the attribute to an empty dictionary if the argument is `None`

cleanlab/models/keras.py
```diff
@@ -75,10 +75,12 @@ def __init__(
...
     ):
         self.model = model
         self.model_kwargs = model_kwargs
         self.compile_kwargs = compile_kwargs
+        self.params = params
```
elisno, line 83:
```suggestion
        self.params: dict = params or {}
```

You can ignore the type-hint if you want.

cleanlab/models/keras.py
```diff
@@ -110,7 +119,7 @@ def fit(self, X, y=None, **kwargs):
             assert_valid_inputs(X, y)
             kwargs["y"] = y
 
-        self.net.fit(X, **kwargs)
+        self.net.fit(X, **self.params, **kwargs)
```
elisno, line 122: Here you might raise a `SyntaxError: keyword argument repeated:`.

Should any repeated keyword argument override the parameters during the call to `fit`?

cleanlab/models/keras.py
```diff
@@ -159,10 +168,12 @@ def __init__(
         compile_kwargs: dict = {
             "loss": tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True)
         },
+        params: dict = {},
```
elisno, line 183: Switch to `None`, etc.

cleanlab/models/keras.py
```diff
@@ -193,7 +211,7 @@ def fit(self, X, y=None, **kwargs):
             assert_valid_inputs(X, y)
             kwargs["y"] = y
 
-        self.net.fit(X, **kwargs)
+        self.net.fit(X, **self.params, **kwargs)
```
elisno, line 226: Risk of `SyntaxError: keyword argument repeated: ...`

cleanlab/models/keras.py
huiwengoh: Nothing wrong! Just completely slipped my mind that this method exists 😶‍🌫️

cleanlab/models/keras.py
huiwengoh: What is the reason that `None` is safer than an empty dictionary? (are there risks to using the empty dict as default?)

cleanlab/models/keras.py
huiwengoh: In my head the design choice is that keyword arguments (if passed) could be allowed to override params previously stored, although I do understand the potential confusion. Perhaps I can throw a warning to let users know some params have been changed, or would just a straight up error be better to prevent users from rewriting their own params?

@jwmueller wondering if you have any insight/opinions

cleanlab/models/keras.py
elisno: It's just risky to use an empty dictionary as a default argument if we end up referring to the argument elsewhere and accidentally modify it.

```python
def add_to_empty(empty_dict: dict = {}):
    assert not empty_dict, f"empty_dict must be empty, but it is not: \n{empty_dict}"
    empty_dict["key"] = "value"

if __name__ == "__main__":
    print("First call:")
    add_to_empty() 
    print("Second call:")
    add_to_empty()
```

Running the above prints out something like:

```
First call:
Second call:
...
AssertionError: empty_dict must be empty, but it is not: 
{'key': 'value'}

File: cleanlab/models/keras.py (hunk 10 above)
Comment (elisno) [reply cleanlab/models/keras.py:122]: You have to either update `self.params` with the relevant keys from `kwargs`, before the method call.
```suggestion
        self.net.fit(X, **self.params)
```
or merge these dictionaries during unpacking:
```suggestion
        self.net.fit(X, **{**self.params, **kwargs})
```

In both cases, keys in `self.params` would be overridden by those in `kwargs`.

tests/test_frameworks.py
```diff
@@ -233,6 +236,64 @@ def test_tensorflow_rarelabel(batch_size, data=DATA_RARE_LABEL, hidden_units=8):
     preds = cl.predict(dataset_tf)
 
 
+def test_keras_sklearn_compatability(data=DATA, hidden_units=128):
+    # test pipeline on Sequential API
+    model = KerasWrapperSequential(
+        [
+            tf.keras.layers.Dense(128, input_shape=[data["num_features"]], activation="relu"),
```
jwmueller, line 243: If we are not testing accuracy of this model, then make it much smaller so the test runs faster? Eg. 2 hidden units instead of 128?

tests/test_frameworks.py
```diff
@@ -233,6 +236,64 @@ def test_tensorflow_rarelabel(batch_size, data=DATA_RARE_LABEL, hidden_units=8):
     preds = cl.predict(dataset_tf)
 
 
+def test_keras_sklearn_compatability(data=DATA, hidden_units=128):
```
jwmueller, line 239: argument `hidden_units` here seems to be ignored below

cleanlab/models/keras.py
jwmueller: I think allowing them to override the params previously stored seems fine to me.
I'd go with Elias' second suggestion, but I don't have much preference, so whatever you think is best @huiwengoh !

cleanlab/models/keras.py
```diff
@@ -36,7 +36,7 @@
 """
 
 import tensorflow as tf
-import keras
+import keras  # type: ignore
```
huiwengoh, line 39: Ignoring type here because keras has no stubs available and will raise error (ref [here](https://github.com/huiwengoh/cleanlab/actions/runs/4084016149/jobs/7040218721))

jwmueller (approved):
LGTM!

PR #603, opened 2023-01-19
fix typo in outliers.ipynb

eltociear:
occuring -> occurring


jwmueller (approved):
Thanks for the fix @eltociear !

Issue #292, opened 2022-06-24
Reminder: upgrade dependency versions in audio tutorial once their bugs are fixed.


jwmueller:
Current versions are there to: 

- Avoid this bug in the tensorflow-io library:
https://github.com/tensorflow/io/issues/1687

- Avoid this bug in speechbrain library:
https://github.com/speechbrain/speechbrain/issues/1459

Remaining TODOs (need to wait until next version of tensorflow-io > 0.26.0 is released to test these:
- [ ] Try upgrading speechbrain, huggingface_hub, and tensorflow-io, tensorflow to latest versions and check if audio tutorial runs
- [ ] Try to get tensorflow_io to work on Mac M1/M2

jwmueller:
Partially addressed by: https://github.com/cleanlab/cleanlab/pull/325

But I think we still need to wait until we can use huggingface_hub versions >= 0.8 to consider this fully addressed, right @ulya-tkch ?

jwmueller:
Remaining item to revisit in future :  Try upgrading speechbrain, huggingface_hub, and tensorflow-io to latest versions and check if audio tutorial runs

jwmueller:
Another issue: tensorflow_io does not work on Mac M1/M2:
https://github.com/tensorflow/io/issues/1643

ulya-tkch:
- tensorflow-io 0.27.0: https://pypi.org/project/tensorflow-io/
- huggingface-hub 0.9.1: https://pypi.org/project/huggingface-hub/#description

jwmueller:
@ulya-tkch doesnt seem like 0.27 tensorflow-io will resolve M1 issue:

https://github.com/tensorflow/io/issues/1625

ulya-tkch:
Yeah I was reading through that, darn.
For all other laptops, newest versions work ✅

PR #607, opened 2023-01-20
upgrade torch in docs

jwmueller:


jwmueller:
Re: https://github.com/cleanlab/cleanlab/issues/292

@ulya-tkch I still don't think we can fully resolve that on M1 macs, right? 
I'm not sure but seems like we would still need to upgrade tensorflow-io and that isn't done right now on M1?

ulya-tkch:
Seems like https://github.com/tensorflow/io/issues/1625 reflects tensorflow.io installation on M1 has not been resolved yet. Same with https://github.com/tensorflow/io/issues/1757.

jwmueller:
> Seems like [tensorflow/io#1625](https://github.com/tensorflow/io/issues/1625) reflects tensorflow.io installation on M1 has not been resolved yet. Same with [tensorflow/io#1757](https://github.com/tensorflow/io/issues/1757).

Cool then upgrading tensorflow-io can be saved for a future PR

PR #608, opened 2023-01-24
feature/mypy-types-cleanlab-internal-utils

unna97:


elisno:
Thanks for the PR!

The internal utils module is quite large, so how about we work on those in smaller segments (say 5-10 functions/methods)?

I think we can just focus on the functions you've annotated up to this point:


- [ ] remove_noise_from_class
- [ ] clip_noise_rates
- [ ] clip_values
- [ ] value_counts
- [ ] round_preserving_row_totals
- [ ] smart_display_dataframe

elisno:
In `remove_noise_from_class`([link](https://github.com/cleanlab/cleanlab/blob/8604b9b7ad78131ae1f07819285b4f11a5838bcb/cleanlab/internal/util.py#L60)):

Note that the `np.copy()` function is not typed.

It's outside of our control, so we should ignore the specific warning associated with that function call.

```diff
- x = np.copy(noise_matrix)
+ x: npt.NDArray["np.floating[T]"] = np.copy(noise_matrix)  # type: ignore[no-untyped-call]
```

We simply assume that the copied variable should have the same type as the input (`noise_matrix`), so we'll help mypy there.

unna97:
Sure @elisno, do you mean with a separate PR for the above functions?

elisno:
You should use this PR for those functions.

unna97:
Yes, for bigger files I am using one PR. while for other segments as you have mentioned before.

unna97:
@elisno  sorry for doing so much in a single commit. I did add type hints for functions other than those you mentioned. There are specific errors in outputs that I am unsure about.

unna97:
@elisno  Did you have a chance to review this?

unna97:
@elisno  I was waiting for suggestions before making any further changes. Will you still be reviewing this branch?

PR #615, opened 2023-02-01
Method to estimate label issues with limited memory via mini-batches

jwmueller:
Adds mini-batch implementations to find label issues in big datasets with limited memory. These are also much faster, but do not currently use multiprocessing and can be sped up even further for datasets with thousands of classes.

These new methods closely approximate the same strategy as: `filter.find_label_issues(..., filter_by=METHOD1, return_indices_ranked_by=METHOD2)`

where by default: `METHOD1="low_self_confidence"`, `METHOD2="self_confidence"`

or if the `LabelInspector` object was instantiated with `quality_score_kwargs = {"method": "normalized_margin"}` then we instead have:
        `METHOD1="low_normalized_margin"`, `METHOD2="normalized_margin"`.
        
But these methods do not construct a full confident joint matrix and only read in labels & pred_probs in mini-batches.
Not all arguments of `find_label_issues()` are currently supported, more arguments can be added in future PRs. 

The new methods are not guaranteed to match `filter.find_label_issues` exactly but always give very close results in our benchmarks on many datasets. The approximations are done to achieve big gains in efficiency.


jwmueller:
Not sure what is going on with the `black` action but they just recently released new black version 2h ago and something is off:  https://pypi.org/project/black/23.1.0/

ulya-tkch:
> Not sure what is going on with the `black` action but they just recently released new black version 2h ago and something is off: https://pypi.org/project/black/23.1.0/



> Not sure what is going on with the `black` action but they just recently released new black version 2h ago and something is off: https://pypi.org/project/black/23.1.0/

We can revert back to stable 22.0 for now with:
```
- uses: psf/black@stable
  with:
    version: "~= 22.0"
```

cleanlab/experimental/README.md
```diff
@@ -2,7 +2,7 @@
 
 Methods in this `experimental` module are bleeding edge and may have sharp edges. They are not guaranteed to be stable between different cleanlab versions.
 
-Some of these files include various models that can be used with cleanlab to find issues in specific types of data. These require dependencies on deep learning and other machine learning packages that are not official cleanlab dependencies. You must install these dependencies on your own if you wish to use them.
+Some of these files include various models that can be used with cleanlab to find issues in specific types of data. These require dependencies on deep learning and other machine learning packages that are not official cleanlab dependencies. You must install these dependencies on your own if you wish to use them.
```
ulya-tkch, line 5: I would still list label_issues_batched in the list below with a couple word summary about what it does (just with no dependencies).
Maybe change title of list to "The files and their dependencies are as follows:"

docs/source/tutorials/faq.ipynb
```diff
@@ -376,6 +376,67 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "41714b51",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from cleanlab.experimental.label_issues_batched import LabelInspector\n",
```
ulya-tkch, line 406: Why is the suggested response to use the `LabelInspector` object instead of `cleanlab.experimental.label_issues_batched import find_label_issues_batched`

cleanlab/experimental/label_issues_batched.py
```diff
@@ -0,0 +1,461 @@
...
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
+
+"""
+Implementation of :py:func:`filter.find_label_issues <cleanlab.filter.find_label_issues>`
+that does not need much memory by operating in mini-batches.
+The recommended usage is two passes over your data: 
+one pass to compute `confident_thresholds`,another to evaluate each label.
```
ulya-tkch, line 21: [nit] add space after comma

cleanlab/experimental/label_issues_batched.py
```diff
@@ -0,0 +1,461 @@
...
+      The only supported kwarg here for now is: `estimation_method`.
+    """
+
+    def __init__(
+        self,
+        *,
+        num_class: int,
+        store_results: bool = True,
+        verbose: bool = True,
+        quality_score_kwargs: Optional[dict] = None,
```
ulya-tkch, line 120: Why not do below as default args? This way you don't need the first two if statements
```
def __init__(
   self,
   *,
   num_class: int,
   ...
   quality_score_kwargs: dict = {},
   num_issues_kwargs: dict = {},
):
```

cleanlab/experimental/label_issues_batched.py
```diff
@@ -0,0 +1,461 @@
...
+    Do NOT modify any of the attributes of this class yourself!
+    Multi-label classification is not supported by this class, it is only for multi-class classification.
+
+    Parameters
+    ----------
+    num_class : int
+      The number of classes in your multi-class classification task.
+
+    store_results : bool, optional
+      Whether this object will store all label quality scores.
```
ulya-tkch, line 96: Specify max length of results stored. `(N,)`

docs/source/tutorials/faq.ipynb
jwmueller: because people with really big datasets should not be loading the entire `labels` and `pred_probs` arrays into memory, which is what is required for `find_label_issues_batched()`

cleanlab/experimental/label_issues_batched.py
jwmueller: you are never supposed to empty dicts as default args, core Python principle

https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

jwmueller:
> > Not sure what is going on with the `black` action but they just recently released new black version 2h ago and something is off: https://pypi.org/project/black/23.1.0/
> 
> We can revert back to stable 22.0 for now with:
> 
> ```
> - uses: psf/black@stable
>   with:
>     version: "~= 22.0"
> ```

I'm sure the black team will resolve it themselves, does not seem worth investing in changing our CI unless this issue persists for a while

cleanlab/count.py
```diff
@@ -163,6 +163,7 @@ def num_label_issues(
         joint = estimate_joint(labels, pred_probs, confident_joint=calculated_confident_joint)
         frac_issues = 1.0 - joint.trace()
         num_issues = np.rint(frac_issues * len(labels)).astype(int)
+        print("testing")
```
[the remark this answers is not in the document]
jwmueller: good catch, removed

cleanlab/experimental/label_issues_batched.py
jwmueller: good catch, fixed

cleanlab/experimental/label_issues_batched.py
```diff
@@ -0,0 +1,465 @@
...
+            + batch_class_counts * batch_thresholds
+        ) / np.clip(
+            self.examples_per_class + batch_class_counts, a_min=1, a_max=None
+        )  # avoid division by 0
+        self.examples_per_class += batch_class_counts
+        self.examples_processed_thresh += batch_size
+
+    def score_label_quality(self, labels: LabelLike, pred_probs: np.ndarray) -> np.ndarray:
+        """
+        Updates the estimate of confident_thresholds stored in this class using a new batch of data.
```
ulya-tkch, line 299: Docstring meant for `update_confident_thresholds()`

ulya-tkch (approved):
LGTM!

cleanlab/experimental/label_issues_batched.py
jwmueller: good catch, fixed it

cleanlab/rank.py
```diff
@@ -494,9 +501,9 @@ def get_self_confidence_for_each_label(
       Lower scores indicate more likely mislabeled examples.
     """
 
-    # np.mean is used so that this works for multi-labels (list of lists)
-    label_quality_scores = np.array([np.mean(pred_probs[i, l]) for i, l in enumerate(labels)])
-    return label_quality_scores
+    # To make this work for multi-label (but it will slow down runtime), replace:
+    # pred_probs[i, l] -> np.mean(pred_probs[i, l])
+    return np.array([pred_probs[i, l] for i, l in enumerate(labels)])
```
ulya-tkch, line 506: Can `get_self_confidence_for_each_label` take in a flag `multi_label` to have this work for both?

ulya-tkch (approved):
LGTM!

cleanlab/rank.py
jwmueller: That is not for this PR, but yes good point. 

Can you open a PR pointing users to how they should get label-quality scores for multi-label datasets in these functions (ie. linking to the proper multilabel functions). Note users are not supposed to use these functions here for multilabel datasets.

cleanlab/experimental/label_issues_batched.py
```diff
@@ -20,15 +20,18 @@
 You can also use this to estimate label quality scores or the number of label issues
 for big datasets with limited memory.
 
-The recommended usage is two passes over your data: 
+With default settings, the results returned from this method closely approximate those returned from:
```
ulya-tkch, line 23: Unsure which method this is referring to since it is in the introduction, maybe clarify "with default settings, the results returned from `find_label_issues_batched()` or example script..."

cleanlab/experimental/label_issues_batched.py
```diff
@@ -401,30 +404,49 @@ def _batch_check(labels: LabelLike, pred_probs: np.ndarray, num_class: int) -> n
...
     Parameters
     ----------
-    labels: np.ndarray or list
-      Given class labels for each example in the dataset, values in ``0,1,2,...,K-1``.
+    labels_file: str, optional
+      Path to .npy file where the entire 1D `labels` numpy array is stored on disk (list format is not supported).
+      This is loaded using: ``np.load(labels_file, mmap_mode="r")``
+      so make sure this file was created via: ``np.save()`` or other compatible methods.
+
+    pred_probs_file: str
```
ulya-tkch, line 438: pred_probs_file: str, optional

cleanlab/experimental/label_issues_batched.py
```diff
@@ -446,6 +468,37 @@ def find_label_issues_batched(
     issue_indices : np.ndarray
       Indices of examples with label issues, sorted by label quality score.
     """
+    if labels_file is not None:
+        if labels is not None:
```
ulya-tkch, line 472: Should we also raise value error if `labels` and `labels_file` are both `None`? Or is that self explanatory?

ulya-tkch: Same with `pred_probs`

tests/test_filter_count.py
```diff
@@ -930,9 +932,21 @@ def test_batched_label_issues():
...
+    labels_file = path.join(mkdtemp(), "labels.npy")
+    pred_probs_file = path.join(mkdtemp(), "pred_probs.npy")
+    np.save(labels_file, data["labels"])
+    np.save(pred_probs_file, data["pred_probs"])
+    f8 = find_label_issues_batched(
+        labels_file=labels_file,
+        pred_probs_file=pred_probs_file,
+        batch_size=int(len(data["labels"]) / 4.0),
+    )
+    assert np.all(f8 == f3)
```
ulya-tkch, line 949: Doesn't f8 batch size directly compare to f2 not f3?

ulya-tkch (approved):
Overall code LGTM! Double check unit test for loading is implemented correctly.

cleanlab/experimental/label_issues_batched.py
jwmueller: this method = Implementation of find_label_issues that does not need much memory by operating in mini-batches.

They are all the same method. "This method" here was referring to mathematical procedure, not a code method. To avoid confusion I've switched it to be called "This approach"

ulya-tkch: Sounds good! It was the same line as in the actual `find_label_issues_batched()` function docstring so felt a bit strange.

cleanlab/experimental/label_issues_batched.py
jwmueller: good catch added

cleanlab/experimental/label_issues_batched.py
jwmueller: it will raise an error already

jwmueller: but I've made the error message more clear

ulya-tkch: You mean in `assert labels is not None`? I was wondering if we should raise a more verbose error explaining one of the two needs to be passed in (since technically both the parameters are optional)

tests/test_filter_count.py
jwmueller: I purposefully want to compare it against the run with a different batch_size. The batch_size should never affect the results

cleanlab/experimental/label_issues_batched.py
jwmueller: ya have added one later in the code:

```
if labels is None:
        raise ValueError("must provide one of: `labels` or `labels_file`")
    if pred_probs is None:
        raise ValueError("must provide one of: `pred_probs` or `pred_probs_file`")
```

PR #616, opened 2023-02-01
Fix dictionary type annotation for OutOfDistribution object

ulya-tkch:


elisno (approved):

cleanlab/outlier.py
```diff
@@ -93,7 +93,7 @@ class OutOfDistribution:
 
     OUTLIER_PARAMS = {"k", "t", "knn"}
     OOD_PARAMS = {"confident_thresholds", "adjust_pred_probs", "method"}
-    DEFAULT_PARAM_DICT: Dict[Union[str, int, None], Union[str, int, None, np.ndarray]] = {
+    DEFAULT_PARAM_DICT: Dict[str, Union[str, int, None]] = {
```
jwmueller, line 96: This change seems to cause failing type check?

cleanlab/outlier.py:366: error: Incompatible types in assignment (expression has type "ndarray[Any, Any]", target has type "Union[str, int, None]")  [assignment]

jwmueller (approved):
LGTM

PR #625, opened 2023-02-10
Removed $ from shell commands to avoid it being copied

sanjanag:


jwmueller (approved):
Thanks for the improvement!

PR #630, opened 2023-02-11
label_issues_batched multiprocessing

clu0:
Current version only multiprocessing for base version. Multiprocessing for calibrated option is trickier, might need to wait for another PR.

Some benchmarking results for `_update_num_label_issues` (all on default estimation_method) with 48 cores:
Single process: 570s
Single process, but with faster hack (use pred_class instead of max_ind): 77s
n_jobs = 48: 21.5s
n_jobs = 48, with faster hack: 18s


clu0:
updated code to include multiprocessing for calibrated option as well.

I found a much better way to do multiprocessing, resulting in following run times:

Not calibrated:
Single process: 491s (running on cluster so a little different from the previous machine used)
Single process, but with faster hack: 76s
n_jobs = 48: 14s
n_jobs = 48, with faster hack: 3s

calibrated:
Single process: 518s
Single process, but with faster hack: 110s
n_jobs = 48: 14s
n_jobs = 48, with faster hack: 3.3s

clu0 (commented):
looks good!

PR #631, opened 2023-02-12
Fix KerasWrapper summary method

huiwengoh:
The `.summary()` method would fail if we did not first call `.fit()`, as `self.net` was not compiled until we called `.fit()`


jwmueller (approved):
Good catch. Can you also demonstrate usage of summary() in the example notebook & blogpost?  We want to demonstrate to users that these wrappers mostly behave like regular Keras models.

PR #634, opened 2023-02-15
Fix batched multiprocessing slow on tall matrices

clu0:
Benchmarking: https://gist.github.com/clu0/38a64505c5151408e86c7564a3e245ec


jwmueller:
For future reference, in the linked plots here: https://gist.github.com/clu0/38a64505c5151408e86c7564a3e245ec 

“multi proc” is the multiprocessing code before this PR, and “multi proc stacked” is the new multiprocessing code from this PR.

jwmueller (approved):
Great further speedups @clu0 !

PR #640, opened 2023-02-24
List more tasks in the docs Quickstart

jwmueller:


docs/source/index.rst
```diff
@@ -86,9 +86,9 @@ When the :py:meth:`.fit() <cleanlab.classification.CleanLearning.fit>` method is
 4. Dataset curation: fix dataset-level issues
 ---------------------------------------------
 
-cleanlab's :py:mod:`dataset <cleanlab.dataset>` module helps you deal with dataset-level issues by  :ref:`finding overlapping classes <cleanlab.dataset.find_overlapping_classes>` (classes to merge), :ref:`rank class-level label quality <cleanlab.dataset.rank_classes_by_label_quality>` (classes to keep/delete), and :ref:`measure overall dataset health <cleanlab.dataset.overall_label_health_score>` (to track dataset quality as you make adjustments).
+cleanlab's :ref:`dataset <dataset_health>` module helps you deal with dataset-level issues -- :py:meth:`find overlapping classes <cleanlab.dataset.find_overlapping_classes>` (classes to merge), :py:meth:`rank class-level label quality <cleanlab.dataset.rank_classes_by_label_quality>` (classes to keep/delete), and :py:meth:`measure overall dataset health <cleanlab.dataset.overall_label_health_score>` (to track dataset quality as you make adjustments).
```
huiwengoh, line 89: This chunk of text here is weirdly formatted because we are using :py:meth: to link to the functions

image

I don't think there is an alternative way to link to functions so we probably want to edit this sentence slightly differently so that the text _find overlapping classes_, _rank class-level label quality_ etc are not formatted differntly

jwmueller: why can't these texts just be formatted differently? It makes it clear you can click on them to see how to actually do what is being stated.

huiwengoh: Sorry I wasn't too clear, these texts are formatted in the form of `code` (ie the font and highlight we usually associate with code), while usually the hyperlinks in our docs are just highlighted blue with an underline (an example from our current docs page below). 

image

I just thought this highlight/formatting was a little odd but I do not have strong feelings about it, so can keep it as is if you think it is alright

jwmueller: Will keep as is in this PR. feel free to update it in the future (eg. via github suggestions) if you have precise formatting improvements in mind. It's probably easiest to just suggest the precise formatting improvement and show a before/after view

PR #642, opened 2023-02-28
Fix tutorial hyperlinks in docs

huiwengoh:


DEVELOPMENT.md
```diff
@@ -213,7 +213,8 @@ Use relative linking to connect information between docs and jupyter notebooks,
 
 - Link another function from within a source code docstring: ``:py:func:`function_name <cleanlab.file.function_name>` ``
 - Link another class from within a source code docstring: ``:py:class:`class_name <cleanlab.file.class_name>` ``
-- Link a tutorial notebook from within a source code docstring: ``:ref:`notebook_name <notebook_name>` ``
+- Link a tutorial (rst file) from within a source code docstring: ``:ref:`tutorial_name <tutorial_name>` ``
+- Link a tutorial notebook (ipynb file) from within a source code docstring: `` `notebook_name <tutorials/notebook_name.ipynb>`_ ``
```
jwmueller, line 217: what about linking to a tutorial notebook from a rst file? That seems to be the situation you are in when editing index.rst rather than being within a source code docstring?

If it's the same either way, can we say:
"from within a source code docstring or rst file"

jwmueller (approved):
Thank you!

Issue #639, opened 2023-02-19
Fix label issues when model does not predict some class ever


jwmueller:
Some users have reported bugs in two situations. Investigate these to verify bugs exist and then address any that come up.

1) When there exists some class k, such that the model never predicts class k for any of the examples in the dataset.
2) When there exists some class k such that pred_probs[:,k] == 0 for all examples whose given label is k. Then the confident_thresholds[k] == 0 which can lead to abnormalities.

One simple fix for case 2, is if `pred_probs[:,k] == 0` for all examples, then can just delete this column from the pred_probs. Subsequently can simply delete all examples with label k from the dataset before finding label issues, keeping track of the how indices in the reduced dataset correspond to indices in the original dataset.

cgnorthcutt:
For clarity, the issue occurs when the model always predicts zero probability for some class. This happens sometimes when a class is underrepresented. When this happens, the class threshold for that class is zero. That means every example s probability is greater than or equal to that class. That means every example always is confidently guessed as that class as the true class. That means anytime that example is actually confidently guessed as some other class, the algorithm will think it's confidently two classes. When that happens this line of code (screenshot) will just choose the argmax prediction, this reducing all label error detection methods to just choosing the argmax (the most naive approach)

![Screenshot_20230302-193902.png](https://user-images.githubusercontent.com/27030270/222625631-4d536a71-aaa0-4b5c-b061-fc51aab7d1cf.png)



Super simplest naive solution is don't allow class thresholds that are zero. (Clip at 1e-6)

jwmueller:
Fixed in https://github.com/cleanlab/cleanlab/pull/647

Issue #621, opened 2023-02-03
ValueError: Values in pred_probs must be between 0 and 1


edmoman:
CleanLab crashes with RandomForestClassifier from scikit-learn. I am able to run LogisticRegression with several datasets but RandomForest fails with all.

# Stack trace

`The least populated class in y has only 1 members, which is less than n_splits=5.                                       Duplicated some data across multiple folds to ensure training does not fail because these classes do not have enough data for proper cross-validation: {23}.                                                                                    Duplicated some data across multiple folds to ensure training does not fail because these classes do not have enough data for proper cross-validation: {20}.                                                                                    Traceback (most recent call last):                                                                                        File "/mnt/c/Users/emoman/Desktop/RISK/scripts/Score/re_rate_cleanlab.py", line 41, in <module>                           cl.fit(X, y)                                                                                                          File "/home/emoman/.local/lib/python3.10/site-packages/cleanlab/classification.py", line 477, in fit                      label_issues = self.find_label_issues(                                                                                File "/home/emoman/.local/lib/python3.10/site-packages/cleanlab/classification.py", line 859, in find_label_issues        label_issues_mask = filter.find_label_issues(                                                                         File "/home/emoman/.local/lib/python3.10/site-packages/cleanlab/filter.py", line 224, in find_label_issues                assert_valid_inputs(                                                                                                  File "/home/emoman/.local/lib/python3.10/site-packages/cleanlab/internal/validation.py", line 99, in assert_valid_inputs                                                                                                                          raise ValueError("Values in pred_probs must be between 0 and 1.")                                                   ValueError: Values in pred_probs must be between 0 and 1.`

# Steps to reproduce

Multiclass classification. I would rather treat is as ordinal classification or regression, when supported. I have 26 classes, some with less than 5 members, but that is another issue. Even if I reduce the number of classes to 10 (in which case I have over 100 members per class, the issues with pred_probs still occurs). Could this be because pred_probs for multiclass is not in the format that CleanLab is expecting? Or because some probabilities are exactly 0?

```
cl = CleanLearning(clf=RandomForestClassifier(random_state=31416, n_jobs=-1), cv_n_folds=5)
cl.fit(X, y)
y_pred_cl = cl.predict(X)
```

# Additional information

- **Cleanlab version**:  CleanLab 2.2.0 
- **Operating system**: Ubuntu 22.04 on WSL2
- **Python version**: Python 3.10.6
- **scikit-learn version**: scikit-learn1.2.1

[...]

PR #647, opened 2023-03-03
Added clipping to address issue #639

ulya-tkch:


jwmueller:
closes #639

cleanlab/count.py
```diff
@@ -1449,7 +1449,8 @@ def get_confident_thresholds(
             np.mean(pred_probs[:, k][labels == k]) if k in unique_classes else BIG_VALUE
             for k in all_classes
         ]
-        return np.array(confident_thresholds)
+        confident_thresholds = np.clip(confident_thresholds, 2e-6, 1.0)
```
jwmueller, line 1452: We do not want a bunch of constants like this floating around throughout the codebase. I think we should address this via proper engineering. 

The proper way to do things is to define a `constants.py` file somewhere in `internal/`

And then define this as a variable in that file like this: ```CONFIDENT_THRESHOLDS_LOWER_BOUND = 2e-6  # lower bound imposed to clip confident thresholds from below, not 1e-6 because of the floating point comparison
```
and then we should import this variable in all files where it's needed:

```
from cleanlab.internal.constants import CONFIDENT_THRESHOLDS_LOWER_BOUND 
```

jwmueller (changes requested):
suggested some engineering improvements for this

jwmueller (approved):
LGTM once this passes black formatting.

Do not merge this until after this PR has been merged:
https://github.com/cleanlab/cleanlab/pull/646

jwmueller:
Closes https://github.com/cleanlab/cleanlab/issues/621

PR #661, opened 2023-04-05
Pass confident joint computed in CleanLearning to filter.find_label_issues

huiwengoh:
Prior to this `filter.find_label_issues` will recompute the confident joint that has already been computed once in `CleanLearning`

### DIFF
--- a/cleanlab/classification.py
+++ b/cleanlab/classification.py
@@ -839,11 +839,6 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
                 pred_probs=pred_probs,
                 thresholds=thresholds,
             )
-        # Add confident joint to find label issue args if it is not previously specified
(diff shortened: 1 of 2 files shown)


cleanlab/classification.py
```diff
@@ -839,6 +839,12 @@ def find_label_issues(
                 pred_probs=pred_probs,
                 thresholds=thresholds,
             )
+        # Add confident joint to find label issue args if it is not previously specified
```
jwmueller, line 842: Isn't this being done too early?  Doesn't the block of code below on "if pulearning" modify self.confident_joint?

To be safe I would only do this addition right before `find_label_issues_kwargs` is actually passed into `filter.find_label_issues`

jwmueller (changes requested):
I think your new block of code needs to be moved a little later in the file?

jwmueller (approved):
LGTM

PR #662, opened 2023-04-06
Add Example codeblock to the docstrings of important functions in the dataset module

Steven-Yiran:
Added a code example to the docstring for the `find_overlapping_classes` method in the cleanlab.dataset module.

The code block is not currently runnable because of the arbitrary `get_data_labels_from_dataset(yourFavoriteDataset)`. I can define the `get_data_labels_from_dataset` using a simple dataset, but I am not sure where to put this function since other docstring examples will also benefit from having this function.

Before addition:
before
After addition:
after

### DIFF
--- a/cleanlab/dataset.py
+++ b/cleanlab/dataset.py
@@ -142,16 +142,21 @@ def find_overlapping_classes(
     Examples
     --------
     >>> from cleanlab.dataset import find_overlapping_classes
-    >>> from cleanlab.classification import CleanLearning
     >>> from sklearn.linear_model import LogisticRegression
+    >>> from sklearn.model_selection import cross_val_predict
     >>> data, labels = get_data_labels_from_dataset(yourFavoriteDataset)
     >>> yourFavoriteModel = LogisticRegression()
-    >>> cl = CleanLearning(yourFavoriteModel)
-    >>> _ = cl.fit(data, labels) # fit model to cleaned data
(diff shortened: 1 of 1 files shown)


cleanlab/dataset.py
```diff
@@ -139,6 +139,27 @@ def find_overlapping_classes(
...
+            data,
+            labels,
+            cv=num_crossval_folds,
+            method="predict_proba",
+        )
+    >>> df = find_overlapping_classes(
+            labels=labels,
+            pred_probs=pred_probs,
+        )
+    >>> df # output pairs of classes that are often mislabeled
```
jwmueller, line 161:
```suggestion
    >>> df = find_overlapping_classes(
            labels=labels,
            pred_probs=pred_probs,
        )  # output pairs of classes that are often mislabeled
```

jwmueller: don't think we need to include the display commands in examples

Steven-Yiran: make sense!

jwmueller (approved):
Thanks for your contribution!  Feel free to add more nice examples like this!

jwmueller:
And remember: once we have a few example codeblocks in the docstrings, it would be nice to add testing capabilities in CI that automatically run these example codeblocks (as additional tests). Eg. using some tool like: https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

PR #664, opened 2023-04-07
Extract function for computating ood scores from distances

elisno:
# Description

In this PR, the `OutOfDistribution._get_ood_features_scores` method is refactored by extracting the logic for converting distances to Out-of-Distribution (OOD) feature scores into a separate function. 

This change makes the code more modular and readable, allowing for easier maintenance and testing.

The newly introduced function is called `transform_distances_to_scores`. Its arguments include:

- An array of (sorted) distances, 
- Number of neighbors (k),
- The scale parameter (t).
 
It returns an array of OOD feature scores.

## Example usage

```python
>>> import numpy as np
>>> from cleanlab.outlier import transform_distances_to_scores
>>> distances = np.array([[0.0, 0.1, 0.25],
...                       [0.15, 0.2, 0.3]])
>>> transform_distances_to_scores(distances, k=2, t=1)
array([0.95122942, 0.83945702])
```

## Other changes
As a result of this refactoring, the `_get_ood_features_scores` method now calls the `transform_distances_to_scores` function to compute the OOD feature scores before returning the final tuple.

### DIFF
--- a/cleanlab/outlier.py
+++ b/cleanlab/outlier.py
@@ -450,52 +450,6 @@ def _get_ood_features_scores(
     return (ood_features_scores, knn)
 
 
-def transform_distances_to_scores(distances: np.ndarray, k: int, t: int) -> np.ndarray:
-    """Returns an outlier score for each example based on its average distance to its k nearest neighbors.
-
-    The transformation of a distance, :math:`d` , to a score, :math:`o` , is based on the following formula:
-
-    .. math::
-        o = \\exp\\left(-dt\\right)
-
-    where :math:`t` scales the distance to a score in the range [0,1].
-
-    Parameters
-    ----------
-    distances : np.ndarray
-        An array of distances of shape ``(N, num_neighbors)``, where N is the number of examples.
-        Each row contains the distances to each example's `num_neighbors` nearest neighbors.
-        It is assumed that each row is sorted in ascending order.
-
-    k : int
-        Number of neighbors used to compute the average distance to each example.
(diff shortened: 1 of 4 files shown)


cleanlab/outlier.py
```diff
@@ -446,12 +446,54 @@ def _get_ood_features_scores(
     # neighbor of each point is the point itself, at a distance of zero.
     distances, _ = knn.kneighbors(features)
 
+    ood_features_scores = transform_distances_to_scores(distances, cast(int, k), t)
+    return (ood_features_scores, knn)
+
+
+def transform_distances_to_scores(distances: np.ndarray, k: int, t: int) -> np.ndarray:
```
jwmueller, line 453:
```suggestion
def _transform_distances_to_scores(distances: np.ndarray, k: int, t: int) -> np.ndarray:
```

jwmueller: safer as internal method?  Not sure this should be exposed to users since we want to avoid offering a library with tons of functions

elisno: > safer as internal method? Not sure this should be exposed to users since we want to avoid offering a library with tons of functions

I extracted this function for it to be shared across the library, outside of the `cleanlab.outlier` module.

The leading underscore indicates that the function only belongs within the `cleanlab.outlier` module, which I don't want.

## Solution

I'll make a `cleanlab.internal.outlier` module and move the function there (without the underscore) so it won't be exposed as a part of the public API. I think that's safe enough.

ulya-tkch (approved):
lgtm to keep it in internal!

PR #667, opened 2023-04-14
Update softmax to more stable variant

ulya-tkch:
Additional improvement could be to update softmin to avoid `1 - scores` overflow error like so:
```python3
def softmin(scores, temperature=0.99, axis=0):
    scores = np.array(scores)
    softmax_scores = softmax(-1 * scores, temperature, axis)
    return np.dot(softmax_scores, scores)
```
This however changes the results and its effect on performance would need to be tested further.


jwmueller (approved):
LGTM

PR #668, opened 2023-04-16
Add testable example codeblock to functions in the dataset module

Steven-Yiran:


Steven-Yiran:
Modified docstring example style and added doctest setup code in "docs/source/cleanlab/dataset.rst" so that the docstring examples in the dataset modules are runnable. To execute doctest, run `make doctest` in the "/docs" directory. 

The test currently results in two failures because it is hard to write out the expected output for the `health_summary` and `overall_label_health_score` functions. 
As shown by:
f1

f2

 The new docstring example display looks like this:
Screen Shot 2023-04-16 at 14 09 56
Screen Shot 2023-04-16 at 14 09 07
Screen Shot 2023-04-16 at 14 10 07
Screen Shot 2023-04-16 at 14 10 15

cleanlab/dataset.py
```diff
@@ -338,27 +317,18 @@ def overall_label_health_score(
...
-            yourFavoriteModel,
-            data,
-            labels,
-            cv=3,
-            method="predict_proba",
-        )  # generate cross-validation estimates for each input data point
-    >>> score = overall_label_health_score(
-            labels=labels,
-            pred_probs=pred_probs,
-        )  # a score measuring the overall quality of all labels in a dataset.
```
elisno, line 361: Note that docstest is failing in functions that print the output instead of returning it.

Sadly, the `# doctest: +ELLIPSIS` directive won't work in this case.

It would be nice if we could somehow suppress the printouts for doctest without changing the docstrings.

elisno (commented):
Hey @Steven-Yiran , thanks for the PR. This looks good. I have one request and some feedback that I would like your opinion on.

docs/source/tutorials/pred_probs.npz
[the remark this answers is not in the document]
Steven-Yiran: Thanks for pointing this out! I will remove them from the tracking list.

cleanlab/dataset.py
Steven-Yiran: Thanks for pointing out the reason for the failed tests. Currently, I can think of two ways to suppress the outputs. The first one is to supply the `verbose=False` parameter to the two printing functions. But this will mean two additional not-so-relevant parameters in the examples.

Another one is overriding the print function in the doctest setup code in "dataset.rst". It would look like this:
Screen Shot 2023-04-16 at 23 46 22
But I am not sure why this method did not work (i.e. it seems like the built-in print function is still being used during doctest). Do you have any thoughts on the reason?

elisno: I'm not sure if this redefines `builtin.print`. Maybe this is a namespace issue?

elisno:
Let's skip the doctests that are failing in this module for now.

In a separate PR (or multiple PRs), we need to

- [ ] Run doctest successfully for dataset.health_summary
- [ ] Run doctest successffully for dataset.overall_label_health_score

elisno:
Awesome stuff! We really appreciate the effort you put into adding doctests, @Steven-Yiran!

We'll worry about full doctest coverage at a later time. If you come up with a fix for the failing doctests, could you make a new PR?

elisno (approved):

Steven-Yiran:
Sounds great! Thanks for your feedback @elisno !

PR #683, opened 2023-04-27
Remove the "health_summary_kwargs" dictionary argument

elisno:
kwargs for health_summary() will now be handled by `health_summary_parameters` exclusively

Users only have to use `health_summary_parameters` to interact with `health_summary()` after this change.

Also, ensured that the `health_summary_parameters` are copied so they won't get overwritten by Datalab.


jwmueller (approved):
Feel free to merge whenever!

PR #689, opened 2023-05-01
MAINT: standardize documentation and simplify code for outlier

DerWeh:
This is a small maintenance pull request:

* Make docstrings adhere to PEP 257
* Use 'early returns' to reduce indention
* Simplify few lines of code by dedicated functions

----

I have two more questions:

### 1 Implementation of entropy

Is there a particular reason clip is used, this seems very weird to me, as it increases entropy but setting a minimal.
The effect should be negligible, as the value is tiny, but still it seems like a dirty fix to me.
Personally, I would have avoided the issue of zeros using `where`. Something along the line of:

```python
def get_normalized_entropy(
    pred_probs: np.ndarray, min_allowed_prob: float = CLIPPING_LOWER_BOUND
) -> np.ndarray:
    # Note that dividing by log(num_classes) changes the base of the log which rescales entropy to 0-1 range
    num_classes = pred_probs.shape[1]

    finite = pred_probs > min_allowed_prob
    entropy = -np.sum(pred_probs * np.log(clipped_pred_probs, where=finite), axis=1, where=finite) 
    return entropy / np.log(num_classes)
```
The logarithm is only calculated where `where` is `True`, and likewise the sum contains only terms `where` (so we calculate the product for all elements, but we ignore the undefined ones).

### 2 References using `:py:func:`
You use, e.g., `` :py:func:`find_top_issues <cleanlab.rank.find_top_issues>` ``.
Is there a particular reason for this? To me, using [tilde](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-syntax) is more readable in the raw docstrings, as the syntax is more concise:
 `` :py:func:`~cleanlab.rank.find_top_issues` ``
 Furthermore, personally I would prefer the usage of default roles, further shortening it to:
 `` `~cleanlab.rank.find_top_issues` ``

### DIFF
--- a/cleanlab/internal/label_quality_utils.py
+++ b/cleanlab/internal/label_quality_utils.py
@@ -76,9 +76,7 @@ def _subtract_confident_thresholds(
 
     # Re-normalize by shifting data to take care of negative values from the subtraction
     pred_probs_adj += confident_thresholds.max()
-    pred_probs_adj /= pred_probs_adj.sum(axis=1)[
-        :, None
-    ]  # The [:, None] adds a dimension to make the /= operator work for broadcasting.
+    pred_probs_adj /= pred_probs_adj.sum(axis=1, keepdims=True)
 
     return pred_probs_adj
 
--- a/cleanlab/internal/label_quality_utils.py
+++ b/cleanlab/internal/label_quality_utils.py
@@ -68,10 +68,7 @@ def _subtract_confident_thresholds(
                 "Cannot calculate confident_thresholds without labels. Pass in either labels or already calculated "
                 "confident_thresholds parameter. "
             )
-        else:
-            confident_thresholds = get_confident_thresholds(
-                labels, pred_probs, multi_label=multi_label
-            )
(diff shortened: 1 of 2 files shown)


huiwengoh:
Hi @DerWeh! 
Thanks for you contribution - your PR looks great!

To address your questions:

1. Clipping is primarily used to prevent doing computations on any undefined value, no particular reason otherwise. I agree that our current implementation does increase this entropy by a minuscule amount. If you have a cleaner solution, feel free to open another PR to address this! 
You could also check out other modules where clipping is being done and perhaps find a cleaner way to address those too 🙂 (eg. [this function](https://github.com/cleanlab/cleanlab/blob/2d00c91f4069d09dd9d5f1d74c809da1aa960199/cleanlab/rank.py#L282) uses a similar clipping trick)

2. Again, no particular reason - we were probably just unaware that there was an alternative way to display these docs. Once again feel free to open another PR to improve our docs!

I will approve and merge this PR as it already has some improvements, please open new PRs to address the remaining points that were discuss. Thank you!

PR #691, opened 2023-05-02
Rename get_summary -> get_issue_summary for Datalab.

elisno:
This only affects Datalab/DataIssues. 

## Sidenote

IssueManager has a different method with the get_summary name that is used in a different way.


sanjanag (approved):

jwmueller (approved):
LGTM! 

Unrelated to this PR, I noticed `get_summary` still exists in the code of  custom_issue_manager.rst, and no longer seems to be an instance method, so that probably needs to be fixed.

PR #693, opened 2023-05-03
Change sentence transformer model name

huiwengoh:
Prevent reusing the same name as the Logistic Regression model below


jwmueller (approved):
Lgtm

PR #694, opened 2023-05-04
Seed noniid permutation tests

elisno:
This PR adds a `seed` argument to the constructor of `NonIIDIssueManager` that can be used to reproduce the results of the permutation tests run when calling the `find_issues` method of the class.

### DIFF
--- a/cleanlab/datalab/issue_manager/noniid.py
+++ b/cleanlab/datalab/issue_manager/noniid.py
@@ -103,6 +103,7 @@ class NonIIDIssueManager(IssueManager):
         metric: Optional[str] = None,
         k: int = 10,
         num_permutations: int = 25,
+        seed: Optional[int] = None,
         **_,
     ):
         super().__init__(datalab)
@@ -113,6 +114,7 @@ class NonIIDIssueManager(IssueManager):
             "ks": simplified_kolmogorov_smirnov_test,
         }
         self.background_distribution = None
+        self.seed = seed
 
     def find_issues(self, features: Optional[npt.NDArray] = None, **kwargs) -> None:
         knn_graph = self._process_knn_graph_from_inputs(kwargs)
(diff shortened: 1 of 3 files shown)


cleanlab/datalab/issue_manager/noniid.py
```diff
@@ -103,6 +103,7 @@ def __init__(
         metric: Optional[str] = None,
         k: int = 10,
         num_permutations: int = 25,
+        seed: Optional[int] = None,
```
jwmueller, line 106: why choose default None? Don't we want this to be deterministic by default?
We should still allow user to pass None if they dont want to set seed, but my suggestion is to set default = 0 say

jwmueller, line 106:
```suggestion
        seed: Optional[int] = 0,
```

cleanlab/datalab/issue_manager/noniid.py
```diff
@@ -239,6 +241,8 @@ def _build_statistics_dictionary(self, knn_graph: csr_matrix) -> Dict[str, Dict[
     def _permutation_test(self, num_permutations) -> float:
         N = self.N
 
+        if self.seed is not None:
+            np.random.seed(self.seed)
```
jwmueller, line 245: @jecummin can you confirm all non-determinism in this method happens after here and is only in np methods?

elisno:
There's flaky unit test that's unrelated to this PR that's failing CI.

It's addressed in https://github.com/cleanlab/cleanlab/pull/695.

cleanlab/datalab/issue_manager/noniid.py
elisno: Changed behavior to be deterministic by default.

cleanlab/datalab/issue_manager/noniid.py
jecummin: Confirmed!

jwmueller (approved):
Thanks for this!

PR #699, opened 2023-05-04
Drop summary scores from Datalab.report()

elisno:
The issue summary has a "scores" column that's useful for comparing differents runs of the same issue checks.

It may be too confusing for users looking at the top of the report, so by default we just omit it.

### Before

```
lab.report()

    issue_type    score  num_issues
         label 0.909091          11
       outlier 0.522080           6
near_duplicate 0.246459           4

(Note: A lower score indicates a more severe issue across all examples in the dataset.)

Dataset Information: num_examples: 132, num_classes: 3
...
``` 

### After

```
lab.report()

    issue_type  num_issues
         label          11
       outlier           6
near_duplicate           4

Dataset Information: num_examples: 132, num_classes: 3
...
```

### DIFF
--- a/cleanlab/datalab/datalab.py
+++ b/cleanlab/datalab/datalab.py
@@ -273,6 +273,7 @@ class Datalab:
         num_examples: int = 5,
         verbosity: Optional[int] = None,
         include_description: bool = True,
+        show_summary_score: bool = False,
     ) -> None:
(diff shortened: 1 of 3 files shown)


huiwengoh (approved):
Swapped the if statement around to remove unnecessary negation - otherwise LGTM

PR #700, opened 2023-05-04
DOC: use default rules for shorter, more readable links

DerWeh:
Use the Spinx setting for [default_role](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-default_role) to shorten links in docstrings.

This way `` `cleanlab.filename.qualifier` `` automatically generates a link, a tilde can be used (`` `~cleanlab.filname.qualifier` ``) to shorten the link only to ``qualifier``.

I exemplarily changed the links in the docstrings of `cleanlab.outlier`. To my mind, this results in more readable doc-strings, if you were e.g. to explore it in an idea (not in the rendered HTML pages).

-----
P.S.: If desired, I could also set up [intersphinx](https://www.sphinx-doc.org/en/master/usage/quickstart.html#intersphinx), to also link to external function, e.g., the `np.ndarray` would link to NumPy's documentation. This is a personal preference of mine, but certainly nothing more than a matter of taste (like the whole pull request).

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -293,8 +293,8 @@ Try to adhere to this standardized terminology unless you have good reason not t
 
 Use relative linking to connect information between docs and jupyter notebooks, and make sure links will remain valid in the future as new cleanlab versions are released! Sphinx/html works with relative paths so try to specify relative paths if necessary. For specific situations:
 
-- Link another function of class from within a source code docstring: `` `~cleanlab.file.name` ``.
-  This uses the default role 'py:obj', the leading tilde shortens the link to only display `name`.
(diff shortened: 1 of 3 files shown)


huiwengoh (approved):
Thank you for the contribution! 
Our docstrings will look so much nicer in github/python files thanks to this change!

Thanks for suggesting intersphinx, but in my opinion we don't really need that now. I believe most of our external links are pretty straightforward (ie. common libraries), and having too many hyperlinks on our docs might get overwhelming. I'll keep in mind that it is an option tho!

I'm merging this PR, great work making our outliers docs look better 😁 
If you'd like, we'll be super grateful if you want to improve our other docs as well!

PR #701, opened 2023-05-04
CI update

sanjanag:
Added min dependency test and updated tests without optional dependencies

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
         run: |
           python -m  pip install --upgrade pip
           pip install .
-          pip install pytest
+          pip install pytest pytest-lazy-fixture
       - name: Run tests
         run: | # these tests provide 50% coverage
           pytest tests/test_multilabel_classification.py tests/test_multiannotator.py tests/test_filter_count.py
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,10 +28,6 @@ jobs:
         run: pip install -e .
       - name: Install development dependencies
         run: pip install -r requirements-dev.txt
-      - name: Check cleanlab runs without optional dependencies
-        # these tests provide 50% coverage
-        run: |
-          pytest tests/test_multilabel_classification.py tests/test_multiannotator.py tests/test_filter_count.py
       - name: Install fasttext for non-Windows machines
         if: matrix.os != 'windows-latest'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -44,6 +42,38 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.7'
+      - name: Install min version of dependencies
+        run: |
+          python -m  pip install --upgrade pip
+          pip install .
+          pip install pytest pytest-lazy-fixture numpy==1.20.0 scikit-learn==1.0 tqdm==4.53.0 pandas==1.1.5 termcolor==2.0.0
```
jwmueller, line 73: are you sure this second install overwrites the dependencies versions if they were already installed during `pip install .` ?

jwmueller: this is not super forwards compatible in the sense somebody will need to remember to update this every time the min versions are updated. I guess it's ok because presumably this part of CI would break (if the old min version no longer works) and then the developer will know they need to update these lines too.  But this check wouldn't be thorough anymore if somebody adds a new dependency with a minimum version and forgets to update CI, so at the very least need to ensure the pip install of cleanlab did not install any other dependencies. 

But stronger version of this would be to programmatically fetch what are the dependencies, what are their minimum versions, uninstall all dependencies, install min versions of them. Is that a lot of work though?

.github/workflows/ci.yml
```diff
@@ -44,6 +42,38 @@ jobs:
...
+      - name: Run tests
+        run: | # this test provides 33% coverage
+          pytest tests/test_classification.py
```
jwmueller, line 76: why not run the same 3 unit tests as above for this check?

.github/workflows/ci.yml
sanjanag: Let me check if there's a command to get answers to these 2 questions: `what are the dependencies, what are their minimum versions,` rest of the part is easy

sanjanag: Added code that automatically gets min versions of dependencies and installs them

.github/workflows/ci.yml
sanjanag: Combined the two checks

jwmueller (approved):
LGTM!

PR #705, opened 2023-05-09
Simplify reporting for outlier issues

elisno:
When constructing a report for outlier issues, we no longer add information about nearest neighbors (as it doesn't immediately provide more signal or a reason why a datapoint is an outlier).

The information is still computed and stored in info, so it should still be accessible via `Datalab.get_info("outlier")...`.

This should affect some notebooks that write a report, like the quickstart tutorial.
image

### Other changes

One of the tests in `test_datalab.py` used info about the nearest neighbors for `get_issues`, so the tests is now changed to handle near_duplicate issues instead.

### DIFF
--- a/cleanlab/datalab/data_issues.py
+++ b/cleanlab/datalab/data_issues.py
@@ -118,14 +118,6 @@ class DataIssues:
                 given_label=info["given_label"], predicted_label=info["predicted_label"]
             )
 
-        if issue_name == "outlier":
-            column_dict = {
-                k: info.get(k)
-                for k in ["nearest_neighbor", "distance_to_nearest_neighbor"]
-                if info.get(k) is not None
-            }
-            specific_issues = specific_issues.assign(**column_dict)
-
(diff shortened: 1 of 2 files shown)


cleanlab/datalab/data_issues.py
```diff
@@ -118,14 +118,6 @@ def get_issues(self, issue_name: Optional[str] = None) -> pd.DataFrame:
                 given_label=info["given_label"], predicted_label=info["predicted_label"]
             )
 
-        if issue_name == "outlier":
```
elisno, line 121: Note that the `info` can still contain the keys.

elisno: It's not within the scope of this PR to remove this from info entirely.

jwmueller (approved):
LGTM!  I assume the failing CI is due to a different issue entirely

Screen Shot 2023-05-09 at 12 41 01 PM

PR #707, opened 2023-05-09
Set non iid num issues with an extra rule

elisno:
With this PR, we perform an additional check for the number of issues when we have a low p-value (under 0.05), vs a high p-value.

We have to change the number of issues in the `is_non_iid_issue` column for this to take effect in the issue summary.

### DIFF
--- a/cleanlab/datalab/issue_manager/noniid.py
+++ b/cleanlab/datalab/issue_manager/noniid.py
@@ -163,20 +163,15 @@ class NonIIDIssueManager(IssueManager):
         score_median_threshold = np.median(scores) * 0.7
         issue_mask = scores < score_median_threshold
         if self.p_value >= self.significance_threshold:
-            self.issues = pd.DataFrame(
-                {
-                    f"is_{self.issue_name}_issue": np.zeros(self.N, dtype=bool),
-                    self.issue_score_key: scores,
-                },
-            )
+            issue_mask = np.zeros(self.N, dtype=bool)
         elif issue_mask.sum() == 0:
             issue_mask[scores.argmin()] = True
-            self.issues = pd.DataFrame(
-                {
-                    f"is_{self.issue_name}_issue": issue_mask,
-                    self.issue_score_key: scores,
-                },
-            )
+        self.issues = pd.DataFrame(
+            {
+                f"is_{self.issue_name}_issue": issue_mask,
+                self.issue_score_key: scores,
+            },
+        )
 
         self.summary = self.make_summary(score=self.p_value)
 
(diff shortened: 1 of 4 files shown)


tests/datalab/issue_manager/test_noniid.py
```diff
@@ -89,7 +89,7 @@ def test_find_issues(self, issue_manager, embeddings):
             issue_manager.summary,
```
jwmueller, line 89: can we add more end-to-end tests of the noniid check at the datalab level?

cleanlab/datalab/issue_manager/noniid.py
```diff
@@ -164,6 +164,16 @@ def find_issues(self, features: Optional[npt.NDArray] = None, **kwargs) -> None:
             },
         )
 
+        num_issues = self.issues[f"is_{self.issue_name}_issue"].sum()
+        if self.p_value >= 0.05:
+            # Set all issues to False if the p-value is greater than 0.05
```
jwmueller, line 169:
```suggestion
        SIGNIFICANCE_THRESHOLD = 0.05
        if self.p_value >= SIGNIFICANCE_THRESHOLD:
            # Set all issues to False if the p-value is insignificant
```

jwmueller: could alternatively make `SIGNIFICANCE_THRESHOLD` an optional hyperparameter of this issuemanager

cleanlab/datalab/issue_manager/noniid.py
```diff
@@ -164,6 +164,16 @@ def find_issues(self, features: Optional[npt.NDArray] = None, **kwargs) -> None:
...
+            self.issues[f"is_{self.issue_name}_issue"] = False
+        elif num_issues == 0:
+            # Make the lowest scoring example an issue if there are no issues
+            self.issues.loc[
+                self.issues[self.issue_score_key].idxmin(), f"is_{self.issue_name}_issue"
+            ] = True
+
```
jwmueller, line 176: why not get rid of these lines?
https://github.com/cleanlab/cleanlab/blob/e0a34113942fcba3986a0adebb85da0097709fe6/cleanlab/datalab/issue_manager/noniid.py#L160

And only define self.issues here in the first place?

cleanlab/datalab/issue_manager/noniid.py
elisno: Added a `significance_threshold` parameter to the constructor.

We should remember not to confuse this with a `threshold` typically used to flag individual data points as having issues.

cleanlab/datalab/issue_manager/noniid.py
elisno: Right. I've reduced the heuristic to only update an `issue_mask`, then `self.issues` is defined only once here.

Thanks!

PR #709, opened 2023-05-11
Fix unbound knn variable in NonIIDIssueManager

elisno:
This PR really just fixed one bug where the `knn` object isn't set within the NonIIDIssuemanager if a `knn_graph` is already accessible.

It's a simple as just initializing the parameter to None. That's already the expected type for `knn` when just the `knn_graph` is used.

Along with this change, I've added some end-to-end tests in Datalab that ensure the summary is updated appropriately with incremental search.
That's how the `knn` bug was discovered in the first place, because the IssueManagers would otherwise fail silently when calling `Datalab.find_issues`.


jwmueller (approved):
Good catch!

PR #710, opened 2023-05-11
clarify when datalab quality score is / isnt comparable

jwmueller:


elisno (approved):
LGTM!

One of the clarifications applied to `get_issue_summary()` instead of `get_issues()`, so I fixed that part.

PR #723, opened 2023-05-19
Include non-iid in default issue checks

elisno:
## Changes in this PR

- `Datalab.find_issues` will run a `NonIIDIssueManager by default (currently if feature embeddings are provided).
- The "is_non_iid_issue" column will only flag the "worst" example if the dataset is considered to have a non-iid issue.
  - This means that the issue summary will only count one non-iid issue.


huiwengoh (approved):
LGTM

PR #729, opened 2023-05-26
Move datalab tutorials

elisno:
The image & audio tutorials have been replaced by their Datalab counterparts, so that Datalab is more prominent in our code base, etc.

We still want to keep the text & tabular tutorials, but I've added a link to the corresponding Datalab tutorials.

image


docs/source/tutorials/tabular.ipynb
```diff
@@ -7,6 +7,19 @@
...
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"datalab-alert\">\n",
+    "<h4>Consider Using Datalab</h4>\n",
+    "<br/>\n",
+    "If you are just interested in detecting a wide variety of issues in these datasets, check out the Datalab version of this tutorial instead by \n",
+    "[clicking here](datalab/tabular.ipynb).\n",
+    "</div>"
```
jwmueller, line 20: Questions about this:  Users are accessing it from: docs.cleanlab.ai, ipynb notebook run locally, colab.

Is the experience good on all 3? I don't get that impression but maybe I'm wrong.

jwmueller: eg. the link seems to just be wrong when accessed from certain ways? In such cases should always explicitly write out what you are linking, instead of just "click here" (so even if link is broken somebody can search it themselseves)

jwmueller: also our team invested tons of time finding a "box-like" display that looks good across: docs.cleanlab.ai, in github viewer of ipynb file, in ipynb file locally in jupyter, in colab.

That is what we use for the quickstart block. Why not just use that same block format instead of re-inventing the wheel here.

docs/source/tutorials/text.ipynb
```diff
@@ -7,6 +7,19 @@
     "# Text Classification with TensorFlow, Keras, and Cleanlab\n"
    ]
   },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"datalab-alert\">\n",
```
jwmueller, line 15: see my other comments about this in tabular noteboook

docs/source/tutorials/image.ipynb
```diff
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Image Classification with PyTorch and Cleanlab\n"
+    "# Image Classification with PyTorch and Datalab\n"
```
jwmueller, line 7:
```suggestion
    "# Image Classification with PyTorch and Cleanlab\n"
```

jwmueller: We want to minimize mentions of Datalab outside of code-level discussions. This is just a variable inside of cleanlab going forward, but we should reference cleanlab whenever we are talking about methods and not specific code

docs/source/tutorials/audio.ipynb
```diff
@@ -6,7 +6,7 @@
     "id": "eVufWTY3jRPx"
    },
    "source": [
-    "# Audio Classification with SpeechBrain and Cleanlab\n",
+    "# Audio Classification with SpeechBrain and Datalab\n",
```
jwmueller, line 9:
```suggestion
    "# Audio Classification with SpeechBrain and Cleanlab\n",
```

jwmueller: see my other comment about this

docs/source/tutorials/text.ipynb
```diff
@@ -7,6 +7,19 @@
...
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert alert-info\">\n",
+    "Consider Using Datalab\n",
+    "<br/>\n",
+    "\n",
+    "If you are just interested in detecting a wide variety of issues in these datasets, check out the [Datalab version of this text tutorial](datalab/text.ipynb).\n",
```
jwmueller, line 19: TODO: revisit link

docs/source/tutorials/tabular.ipynb
```diff
@@ -7,6 +7,19 @@
...
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert alert-info\">\n",
+    "Consider Using Datalab\n",
+    "<br/>\n",
+    "\n",
+    "If you are just interested in detecting a wide variety of issues in these datasets, check out the [Datalab version of this tabular tutorial](datalab/tabular.ipynb).\n",
```
jwmueller, line 19: TODO: revisit link

elisno:
@jwmueller I think this is good to merge as all of the comments have been addressed. Just needs approval.

jwmueller (approved):
LG thanks!

PR #738, opened 2023-06-07
Add note about tensorflow>=2.11

huiwengoh:
The Optimizer class was updated in https://github.com/tensorflow/tensorflow/issues/58973, to use deepcopy in tensorflow>=2.11 we now need to use the legacy class


jwmueller (approved):
Can you also add the phrase:

This wrapper class is only fully compatible with ``tensorflow<2.11``

To the other places (i.e. notebooks) that use this class?

huiwengoh:
This is the only place it is mentioned in this repo, but made those edits and tagged you as reviewer in the examples and blog repo!

PR #739, opened 2023-06-07
Improve visualize function for object detection

ulya-tkch:
- fix linestyle when overlay=False
- add option to save image as different types of files

### DIFF
--- a/cleanlab/object_detection/summary.py
+++ b/cleanlab/object_detection/summary.py
@@ -77,7 +77,7 @@ def visualize(
         Optional dictionary mapping one-hot-encoded class labels back to their original class names in the format ``{"integer-label": "original-class-name"}``.
 
     save_path:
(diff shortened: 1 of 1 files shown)


jwmueller (approved):
LGTM!

cleanlab/object_detection/summary.py
```diff
@@ -135,27 +135,29 @@ def visualize(
...
-            )
-    elif save_path:
+        bbox_extra_artists = (legend,)
+
+    if save_path:
+        allowed_image_formats = set(["png", "pdf", "ps", "eps", "svg"])
+        image_format = (
+            save_path[-3:]
+            if len(save_path) > 2
+            and save_path[-4] == "."
```
huiwengoh, line 148: I believe this check will fail for `filename.ps` since `save_path[-4] != "."` 
Would it instead be better to use `save_path.split(".")`? Tho you might need some try-except blocks to catch in case the path provided does not contain a period

PR #742, opened 2023-06-10
Fix CI

anishathalye:
Among other changes, this patch also disables HTML validation. This is a bit tricky to get right, because when the Markdown is processed by GitHub, it does some transformations that aren't mimicked by Pandoc. The result of converting the existing Markdown to HTML using Pandoc is in fact invalid HTML5.

The best way to fix this would be to use the [GitHub Markdown API](https://docs.github.com/en/rest/markdown) to render the Markdown, but that doesn't seem worth the effort right now. Instead, this patch just disables HTML validation but preserves the link checking functionality (and makes it start passing again).


elisno (approved):
LGTM!

PR #744, opened 2023-06-13
include n_jobs in cleanlearning kwargs docstring

jwmueller:


ulya-tkch (approved):
LGTM, params exist and act as expected. 

[nit] `n_jobs` description in `cleanlab.filter.find_label_issues` says "Set this to 1 to *disable* parallel processing (if its causing issues)." if we want to use same exact wording.

PR #745, opened 2023-06-15
Update object detection tutorial to clarify out-of-sample predictions

ulya-tkch:
- Add out-of-sample clarification to notebook
- Fix filename check to avoid index error with bad filenames (good catch @huiwengoh!)

### DIFF
--- a/cleanlab/object_detection/summary.py
+++ b/cleanlab/object_detection/summary.py
@@ -142,13 +142,9 @@ def visualize(
 
     if save_path:
         allowed_image_formats = set(["png", "pdf", "ps", "eps", "svg"])
-        image_format = (
-            save_path[-3:]
-            if len(save_path) > 3
-            and save_path[-4] == "."
-            and save_path[-3:] in allowed_image_formats
-            else None
-        )
+        image_format = save_path.split('.')[-1]
+        if '.' not in save_path or image_format not in allowed_image_formats:
+            image_format = None
 
         plt.savefig(
             save_path,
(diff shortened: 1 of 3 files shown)


cleanlab/object_detection/summary.py
```diff
@@ -144,7 +144,7 @@ def visualize(
         allowed_image_formats = set(["png", "pdf", "ps", "eps", "svg"])
         image_format = (
             save_path[-3:]
-            if len(save_path) > 2
+            if len(save_path) > 3
```
huiwengoh, line 147: Wouldn't this still not work for `save_path = "image.ps"` ? In this case `save_path[-4] == "e"` and `save_path[-3:] == ".ps"`

cleanlab/object_detection/summary.py
```diff
@@ -142,13 +142,9 @@ def visualize(
...
-        image_format = (
-            save_path[-3:]
-            if len(save_path) > 2
-            and save_path[-4] == "."
-            and save_path[-3:] in allowed_image_formats
-            else None
-        )
+        image_format: Optional[str] = save_path.split(".")[-1]
+        if "." not in save_path or image_format not in allowed_image_formats:
+            image_format = None
```
huiwengoh, line 147: Would perhaps be nice to add a test to make sure that the following `plt.savefig()` call still works as expected when this `image_format=None`

huiwengoh (approved):
Small comment re adding a test case to ensure the plt code still works on that edge case. Otherwise LGTM!

PR #746, opened 2023-06-17
Fixed bug that shows warning message even after specifying estimation…

gordon-lim:
…_method = 'off_calibrated_custom'

The warning message that says 'confident_joint' will be recomputed shows up even after specifying estimation_method = "off_calibrated_custom". Also, "off_calibrated_custom" is missing from the list of valid methods.


jwmueller (approved):
Good catch!  Thanks for your contribution @gordon-lim

PR #748, opened 2023-06-20
Temporary Github Actions workaround: Pin Python 3.7.16

ulya-tkch:


ulya-tkch:
Can we keep windows at `3.7` as that was the original test? Otherwise LGTM. Reminder to change it back once https://github.com/actions/setup-python/issues/682 is addressed

jwmueller (approved):
@ulya-tkch @sanjanag  Don't forget to undo this once github fixes their actions

PR #750, opened 2023-06-22
lowering swap score for multiple labels associated with a box

aditya1503:
This PR lowers the scores when there might be 2 or more overlapping labels of different classes associated with an object: 
```
bboxes = np.array(
    [
        [359.0, 146.0, 472.0, 360.0],
        [340.0, 22.0, 494.0, 323.0],
        [472.0, 173.0, 508.0, 221.0],
        [486.0, 183.0, 517.0, 218.0],
        [359.0, 144.0, 470.0, 358.0],
        [340.0, 22.0, 494.0, 323.0],
    ]
)
label_classes = [0, 1, 1, 1, 1, 1]
label = {"bboxes": bboxes, "labels": label_classes}
predicion = np.array(
    [np.array([[340.0, 22.0, 494.0, 323.0, 0.9]]), np.empty(shape=(0, 2))], dtype=object
)
is_issue = find_label_issues(
    [label], [predicion], overlapping_label_check=overlapping_label_check
)[0]
if overlapping_label_check:
    assert is_issue == True
else:
    assert is_issue == False
```


cleanlab/object_detection/rank.py
```diff
@@ -369,6 +370,27 @@ def _get_iou(bb1: Dict[str, Any], bb2: Dict[str, Any]) -> float:
     return iou
 
 
+def _has_overlap(label_boxes, labels):
```
aditya1503, line 373: Can't add dining table case yet since iOU (image_length,image_width) isn't present, this doesn't; affect the thresholding, 
adding image resolution we will need user to pass the parameter of resolutions/ original image paths. 
This can be done in the follow up PR where we'll be considering ratio of areas of boxes and original image as a parameter.

cleanlab/object_detection/rank.py
```diff
@@ -448,7 +467,8 @@ def _get_valid_inputs_for_compute_scores_per_image(
             if 0 in similarity_matrix.shape
             else np.min(similarity_matrix[np.nonzero(similarity_matrix)])
         )
-
+    if has_overlap_label_bboxes is None:
```
aditya1503, line 470: Should the user have an option to disable this?

cleanlab/object_detection/rank.py
jwmueller: ok that's fine, please don't worry about that case

cleanlab/object_detection/rank.py
jwmueller: yes ideally there's way for use to ignore this sort of issue

cleanlab/object_detection/rank.py
```diff
@@ -820,22 +846,25 @@ def _compute_swap_box_scores_for_image(
...
+    has_overlap_label_bboxes = auxiliary_input_dict["has_overlap_label_bboxes"]
     scores_swap = np.empty(
         shape=[
             len(lab_labels),
         ]
     )  # same length as number of labeled boxes
     for iid, k in enumerate(lab_labels):
         not_k_idx = pred_labels != k
-
+        if has_overlap_label_bboxes[iid]:
```
jwmueller, line 865: need unit tests

cleanlab/internal/constants.py
```diff
@@ -31,6 +31,7 @@
 LOW_PROBABILITY_THRESHOLD = 0.5  # Param for get_label_quality_score, lowest predicted class probability threshold allowed when considering predicted boxes to identify badly located label boxes.
 HIGH_PROBABILITY_THRESHOLD = 0.95  # Param for objectlab, high probability threshold for considering predicted boxes to identify overlooked and swapped label boxes
 TEMPERATURE = 0.1  # Param for objectlab, temperature of the softmin function used to pool the per-box quality scores for an error subtype across all boxes into a single subtype score for the image. With a lower temperature, softmin pooling acts more like minimum pooling, alternatively it acts more like mean pooling with high temperature.
+LABEL_OVERLAP_THRESHOLD = 0.95  # Param for objectlab, overlap threshold for considering a box for having more than one annotation
```
jwmueller, line 34:
```suggestion
LABEL_OVERLAP_THRESHOLD = 0.95  # Param for objectlab, overlap threshold for deciding that a box has multiple conflicting annotations
```

cleanlab/object_detection/rank.py
```diff
@@ -369,6 +370,23 @@ def _get_iou(bb1: Dict[str, Any], bb2: Dict[str, Any]) -> float:
     return iou
 
 
+def _has_overlap(label_boxes, labels):
+    """This function determines whether each labeled box overlaps with another box of a different class, returning a boolean array."""
```
jwmueller, line 374:
```suggestion
    """This function determines whether each labeled box overlaps with another box of a different class (i.e. virtually the same box having multiple conflicting annotations). It returns a boolean array."""
```

cleanlab/object_detection/rank.py
aditya1503: Added : https://github.com/cleanlab/cleanlab/blob/401743a946885e7a29a2cea4df48e761b60c05a1/cleanlab/object_detection/rank.py#L109

cleanlab/object_detection/rank.py
aditya1503: added: https://github.com/cleanlab/cleanlab/blob/401743a946885e7a29a2cea4df48e761b60c05a1/tests/test_object_detection.py#L626

cleanlab/object_detection/rank.py
```diff
@@ -369,6 +378,23 @@ def _get_iou(bb1: Dict[str, Any], bb2: Dict[str, Any]) -> float:
     return iou
 
 
+def _has_overlap(label_boxes, labels):
+    """This function determines whether each labeled box overlaps with another box of a different class (i.e. virtually the same box having multiple conflicting annotations). It returns a boolean array."""
```
ulya-tkch, line 382: `label_boxes` is called `lab_bboxes` in other functions. Regardless, I think function params should be generalized to (`bbox_list`, `labels`) as the function can calculate overlap matrix for any set of boxes/labels.

cleanlab/object_detection/rank.py
```diff
@@ -369,6 +378,23 @@ def _get_iou(bb1: Dict[str, Any], bb2: Dict[str, Any]) -> float:
...
+    iou_matrix = _get_overlap_matrix(label_boxes, label_boxes)
+    res_overlap = []
```
ulya-tkch, line 384: `result_overlap`, avoid shortening param names

cleanlab/internal/constants.py
```diff
@@ -31,6 +31,7 @@
 LOW_PROBABILITY_THRESHOLD = 0.5  # Param for get_label_quality_score, lowest predicted class probability threshold allowed when considering predicted boxes to identify badly located label boxes.
 HIGH_PROBABILITY_THRESHOLD = 0.95  # Param for objectlab, high probability threshold for considering predicted boxes to identify overlooked and swapped label boxes
 TEMPERATURE = 0.1  # Param for objectlab, temperature of the softmin function used to pool the per-box quality scores for an error subtype across all boxes into a single subtype score for the image. With a lower temperature, softmin pooling acts more like minimum pooling, alternatively it acts more like mean pooling with high temperature.
+LABEL_OVERLAP_THRESHOLD = 0.95  # Param for objectlab, overlap threshold for deciding that a box has multiple conflicting annotations
```
ulya-tkch, line 34:
```suggestion
LABEL_OVERLAP_THRESHOLD = 0.95  # Param for objectlab, minimum IoU threshold for deciding when two boxes overlap used for deciding which objects have multiple conflicting annotations.
```

cleanlab/object_detection/rank.py
```diff
@@ -103,6 +106,9 @@ def get_label_quality_scores(
        swapped examples, bad location examples, and overlooked examples.
        It is important to ensure that the weights are non-negative values and that their sum equals 1.0.
 
+    overlapping_label_check:
+       A boolean that keeps a low swap score for an object that has more than 1 label associated with it.
```
ulya-tkch, line 110: If ``True``, annotated examples with more than one label associated with them are intentionally given a low swap score.

(or some variant of this)

cleanlab/object_detection/rank.py
```diff
@@ -543,6 +578,7 @@ def _compute_overlooked_box_scores_for_image(
     lab_bboxes: Optional[np.ndarray] = None,
     similarity_matrix: Optional[np.ndarray] = None,
     min_possible_similarity: Optional[float] = None,
+    has_overlap_label_bboxes: Optional[np.ndarray] = None,
```
ulya-tkch, line 581: [nit] `has_overlap_label_bboxes` sounds a bit more like a boolean flag instead of a vector. Super nit so feel free to ignore but maybe changing name of this variable to `overlap_label_bboxes` might be clearer. (I always confuse it with `overlapping_label_check` which sounds less like a boolean flag)

cleanlab/object_detection/rank.py
```diff
@@ -675,6 +712,7 @@ def _compute_badloc_box_scores_for_image(
...
```
ulya-tkch, line 715: Does badloc use "has_overlap_label_bboxes"? We should avoid passing this param into functions that don't use it (notice how high_probability_threshold is not passed into this function).

cleanlab/object_detection/rank.py
ulya-tkch: Does badloc use "has_overlap_label_bboxes"? We should avoid passing this param into functions that don't use it (notice how high_probability_threshold is not passed into this function).

cleanlab/object_detection/rank.py
```diff
@@ -980,6 +1023,7 @@ def _get_subtype_label_quality_scores(
     high_probability_threshold: Optional[float] = None,
     temperature: Optional[float] = None,
     aggregation_weights: Optional[Dict[str, float]] = None,
+    overlapping_label_check: bool = True,
```
ulya-tkch, line 1026: Add param to docstring here

tests/test_object_detection.py
```diff
@@ -601,3 +604,34 @@ def test_visualize(monkeypatch, generate_single_image_file):
...
+@pytest.mark.parametrize("overlapping_label_check", [True, False])
+def test_swap_overlap_labels(overlapping_label_check):
+    prediction = predictions[3].copy()
+    label = labels[3].copy()
+    label["bboxes"] = np.append(label["bboxes"], [label["bboxes"][-1]], axis=0)
+    label["labels"] = np.append(label["labels"], (label["labels"][-1] + 1) % 10)
+    score = get_label_quality_scores(
+        [label], [prediction], overlapping_label_check=overlapping_label_check
+    )[0]
+    if overlapping_label_check:
```
ulya-tkch, line 634: Write one more unit test for user facing `compute_swap_box_scores()` function to make sure expected boxes get low overlap score if overlapping_label_check() is on.

cleanlab/object_detection/rank.py
aditya1503: the reason I passed it is: 
_get_valid_inputs_for_compute_scores_per_image is called in all of the functions, so 
has_overlap_label_bboxes is computed over and over again (3 times)

cleanlab/object_detection/rank.py
aditya1503: Fixed here: https://github.com/cleanlab/cleanlab/blob/1b4059b2a99b3e2ebed1c24a1ba3a2e5cfa50142/cleanlab/object_detection/rank.py#L380

tests/test_object_detection.py
aditya1503: Added here: https://github.com/cleanlab/cleanlab/blob/96b10ac0087edd5f90bf30961408a2bed7f75564/tests/test_object_detection.py#L640

cleanlab/object_detection/rank.py
```diff
@@ -103,6 +105,9 @@ def get_label_quality_scores(
        swapped examples, bad location examples, and overlooked examples.
        It is important to ensure that the weights are non-negative values and that their sum equals 1.0.
 
+    overlapping_label_check : bool, default = True
```
ulya-tkch, line 108: Nice job integrating this param into rest of code. Maybe for the 3 docstring, say what happens if set to false. Something like:
If True, annotated examples with more than one label associated with them are intentionally given a low swap score. Set to ``False`` to assign annotated examples with more than one label associated with them their expected scores instead.

(remember to do this in all 3 docstrings)

tests/test_object_detection.py
```diff
@@ -601,3 +604,76 @@ def test_visualize(monkeypatch, generate_single_image_file):
...
+def test_badloc_high_low_probability_threshold():
+    prediction = predictions[3].copy()
+    label = labels[3].copy()
+    label["bboxes"] = np.append(label["bboxes"], [label["bboxes"][-1]], axis=0)
+    label["labels"] = np.append(label["labels"], (label["labels"][-1] + 1) % 10)
+    score = compute_badloc_box_scores([label], [prediction], low_probability_threshold=1.0)[0]
+    assert np.allclose(score, np.ones_like(score), atol=1e-2)
+
+
+def test_overlooked_high_high_probability_threshold():
```
ulya-tkch, line 665: Why are these functions called "test_swap_high_high" or "test_badloc_high_low". Shouldn't it just be "test_badloc_low_probability_threshold()" and "test_swap_high_probability_threshold()"?

ulya-tkch (approved):
A little more info on the flag in the docstring, otherwise LGTM!

cleanlab/object_detection/rank.py
```diff
@@ -66,6 +67,7 @@ def get_label_quality_scores(
     labels: List[Dict[str, Any]],
     predictions: List[np.ndarray],
     aggregation_weights: Optional[Dict[str, float]] = None,
+    overlapping_label_check: bool = True,
```
jwmueller, line 70: why is this appearing before the *? It should be after as an optional argument that must be specified using keyword only.

cleanlab/object_detection/rank.py
```diff
@@ -66,6 +67,7 @@ def get_label_quality_scores(
     labels: List[Dict[str, Any]],
     predictions: List[np.ndarray],
     aggregation_weights: Optional[Dict[str, float]] = None,
```
jwmueller, line 69: This `aggregation_weights` argument should also come after the * (as with all optional arguments).

@ulya-tkch why was this argument put before the *

cleanlab/object_detection/rank.py
```diff
@@ -103,6 +105,9 @@ def get_label_quality_scores(
        swapped examples, bad location examples, and overlooked examples.
        It is important to ensure that the weights are non-negative values and that their sum equals 1.0.
 
+    overlapping_label_check : bool, default = True
+       If True, annotated examples with more than one label associated with them are intentionally given a low swap score.
```
jwmueller, line 109:
```suggestion
       If True, boxes annotated with more than one class label  have their swap score penalized.  Set this to False if you are not concerned when two very similar boxes exist with different class labels in their given annotations.
```

cleanlab/object_detection/rank.py
jwmueller: I added suggestion of this in one of the docstrings, you can copy it for the others

cleanlab/object_detection/rank.py
ulya-tkch: I did not add `aggregation_weights` param to the repo, but I should've caught that during this review. Good point!

cleanlab/object_detection/rank.py
aditya1503: Added

cleanlab/object_detection/filter.py
```diff
@@ -42,8 +42,9 @@
 def find_label_issues(
     labels: List[Dict[str, Any]],
     predictions: List[np.ndarray],
-    *,
     return_indices_ranked_by_score: bool = False,
```
ulya-tkch, line 45: `return_indices_ranked_by_score` param should be optional as well.

cleanlab/object_detection/filter.py
```diff
@@ -84,6 +85,10 @@ def find_label_issues(
     return_indices_ranked_by_score:
         Determines what is returned by this method (see description of return value for details).
 
+    overlapping_label_check : bool, default = True
+       If True, boxes annotated with more than one class label  have their swap score penalized.  Set this to False if you are not concerned when two very similar boxes exist with different class labels in the given annotations.
```
ulya-tkch, line 89:
```suggestion
       If True, boxes annotated with more than one class label have their swap score penalized.  Set this to False if you are not concerned when two very similar boxes exist with different class labels in the given annotations.
```

cleanlab/object_detection/filter.py
```diff
@@ -114,9 +120,10 @@ def find_label_issues(
 def _find_label_issues(
     labels: List[Dict[str, Any]],
     predictions: List[np.ndarray],
-    *,
     return_indices_ranked_by_score: bool = True,
```
ulya-tkch, line 123: Param `return_indices_ranked_by_score` and `scoring_method` should be after * as well

cleanlab/object_detection/rank.py
```diff
@@ -65,9 +66,10 @@
 def get_label_quality_scores(
     labels: List[Dict[str, Any]],
     predictions: List[np.ndarray],
-    aggregation_weights: Optional[Dict[str, float]] = None,
-    *,
     verbose: bool = True,
```
ulya-tkch, line 69: verbose should come after `*`

cleanlab/object_detection/rank.py
```diff
@@ -169,11 +175,12 @@ def issues_from_scores(label_quality_scores: np.ndarray, *, threshold: float = 0
 def _compute_label_quality_scores(
     labels: List[Dict[str, Any]],
     predictions: List[np.ndarray],
-    aggregation_weights: Optional[Dict[str, float]] = None,
-    *,
     method: str = "objectlab",
```
ulya-tkch, line 178: method param should go below `*`

cleanlab/object_detection/rank.py
```diff
@@ -192,13 +199,14 @@ def _compute_label_quality_scores(
...
-            predictions,
+            labels=labels,
+            predictions=predictions,
             alpha=ALPHA,
             low_probability_threshold=LOW_PROBABILITY_THRESHOLD,
             high_probability_threshold=HIGH_PROBABILITY_THRESHOLD,
             temperature=TEMPERATURE,
             aggregation_weights=aggregation_weights,
+            overlapping_label_check=overlapping_label_check,
         )
```
ulya-tkch, line 210: Not relevant to this PR but can you add an ValueError if the method is not `objectlab`?

cleanlab/object_detection/rank.py
```diff
@@ -410,16 +435,16 @@ def _get_min_possible_similarity(
 
 
 def _get_valid_inputs_for_compute_scores_per_image(
-    *,
     alpha: float,
+    *,
```
ulya-tkch, line 439: why was this shifted below alpha?

cleanlab/object_detection/rank.py
```diff
@@ -464,8 +489,8 @@ def _get_valid_inputs_for_compute_scores_per_image(
 
 def _get_valid_inputs_for_compute_scores(
     alpha: float,
-    labels: Optional[List[Dict[str, Any]]] = None,
-    predictions: Optional[List[np.ndarray]] = None,
+    labels: Optional[List[Dict[str, Any]]],
```
ulya-tkch, line 492: Why are labels and predictions no longer set to None if the param is not passed in?

cleanlab/object_detection/rank.py
```diff
@@ -1010,6 +1054,9 @@ def _get_subtype_label_quality_scores(
     temperature:
         Optional temperature of the softmin function where a lower score suggests softmin acts closer to min. If not provided, a good default is used.
 
+    overlapping_label_check : bool, default = True
+       If True, boxes annotated with more than one class label  have their swap score penalized.  Set this to False if you are not concerned when two very similar boxes exist with different class labels in the given annotations.
```
ulya-tkch, line 1058:
```suggestion
       If True, boxes annotated with more than one class label have their swap score penalized.  Set this to False if you are not concerned when two very similar boxes exist with different class labels in the given annotations.
```

tests/test_object_detection.py
```diff
@@ -601,3 +604,109 @@ def test_visualize(monkeypatch, generate_single_image_file):
...
+
+def test_swap_high_probability_threshold():
+    prediction = predictions[3].copy()
+    label = labels[3].copy()
+    label["bboxes"] = np.append(label["bboxes"], [label["bboxes"][-1]], axis=0)
+    label["labels"] = np.append(label["labels"], (label["labels"][-1] + 1) % 10)
+    score = compute_swap_box_scores(
+        labels=[label], predictions=[prediction], high_probability_threshold=1.0
+    )[0]
+    assert np.allclose(score, np.array([1.0, 1.0, 1.0, 1.0, 1.0, 0.0, 0.0]), atol=1e-2)
```
ulya-tkch, line 712: Don't forget to add test a for find_label_issues() with this new flag (results are different if the flag is turned on and there are overlapping boxes in an image). That test can be seperate or go into the general find_label_issues tests above.

cleanlab/object_detection/rank.py
```diff
@@ -169,11 +175,12 @@ def issues_from_scores(label_quality_scores: np.ndarray, *, threshold: float = 0
 def _compute_label_quality_scores(
     labels: List[Dict[str, Any]],
     predictions: List[np.ndarray],
-    aggregation_weights: Optional[Dict[str, float]] = None,
-    *,
     method: str = "objectlab",
-    threshold: Optional[float] = None,
     verbose: bool = True,
```
jwmueller, line 179: same w verbose arg. Please go through ALL of the object detection code and move all optional args below the * in user-facing methods

cleanlab/object_detection/filter.py
jwmueller: @aditya1503 why do you keep moving the * arguments? please never move them up above optional args

cleanlab/object_detection/rank.py
aditya1503: alpha wasn't optional in the code

cleanlab/object_detection/rank.py
aditya1503: new mypy throws an error, (even though you're checking for None later in this function)

aditya1503: I'll try resolving this

tests/test_object_detection.py
aditya1503: It was added here: https://github.com/cleanlab/cleanlab/blob/1924bf3ad6759b848a203698380cceec3f6e493c/tests/test_object_detection.py#L657

cleanlab/object_detection/rank.py
aditya1503: Added: https://github.com/cleanlab/cleanlab/blob/420e69bed2a1e64c6bf282b7da54359465361694/cleanlab/object_detection/rank.py#L212

cleanlab/object_detection/rank.py
aditya1503: https://github.com/cleanlab/cleanlab/blob/420e69bed2a1e64c6bf282b7da54359465361694/cleanlab/object_detection/rank.py#L175

cleanlab/object_detection/rank.py
jwmueller: SGTM. Note we also don't really need to worry about the usage of * and positional vs non-positional arguments for internal methods FYI. That distinction is just important for user-facing methods.

aditya1503: aah thanks, I was about to ask that question

ulya-tkch (approved):
LGTM!

PR #757, opened 2023-06-28
Make sure CI flags any notebook cell exceptions

huiwengoh:
`nbsphinx_allow_errors` should not be `True` as it will not flag any potential errors in the notebooks and build/push docs regardless 

ref: https://nbsphinx.readthedocs.io/en/0.8.8/allow-errors.html


jwmueller (approved):
good catch!

PR #758, opened 2023-06-29
Add imbalance issue to datalab

tataganesh:
Imbalance data issue added to datalab. It is not part of the default issue types and it can be run by assigning `{"class_imbalance": {}}`  to the `issue_types` parameter of `find_issues`. 

Example Usage
```python
import numpy as np

from cleanlab import Datalab

X = np.random.rand(200, 3)
y = np.zeros(X.shape[0])
y[:5] = 1 # Add rare class
data = {"X": X, "y": y}
pred_probs = np.random.rand(100, 2)
pred_probs = pred_probs / pred_probs.sum(axis=1, keepdims=True)

lab = Datalab(data=data, label_name="y")
lab.find_issues(pred_probs=pred_probs, features=data["X"], issue_types={"class_imbalance": {}})
lab.report()

``` 

<details><summary>See Report</summary>

```
Finding class_imbalance issues ...
Audit complete. 5 issues found in the dataset.

     issue_type  num_issues
class_imbalance           5

Dataset Information: num_examples: 200, num_classes: 2


------------------ class_imbalance issues ------------------

About this issue:
        Examples belonging to the most under-represented class in the dataset.

    Parameters
    ----------
    datalab :
        The Datalab instance that this issue manager searches for issues in.

    fraction:
        Minimum fraction of samples of each class that are present in a dataset without class imbalance.
    

Number of examples with this issue: 5
Overall dataset quality in terms of this issue: 0.9756

   is_class_imbalance_issue  class_imbalance_score
0                      True                  0.025
1                      True                  0.025
2                      True                  0.025
3                      True                  0.025
4                      True                  0.025
```

</details>


cleanlab/datalab/issue_manager/imbalance.py
```diff
@@ -0,0 +1,83 @@
...
+        The Datalab instance that this issue manager searches for issues in.
+
+    fraction:
+        Minimum fraction of samples of each class that are present in a dataset without class imbalance.
+    """
+    issue_name: ClassVar[str] = "class_imbalance"
+    verbosity_levels = {
+        0: [],
+        1: [],
+        2: ["threshold"],
```
elisno, line 49: So I understand correctly, the threshold is related to the `fraction: float = 0.1` parameter?

Do you intend to include it in the report?

cleanlab/datalab/issue_manager/imbalance.py
```diff
@@ -0,0 +1,83 @@
...
+        labels = self.datalab.labels
+        K = len(self.datalab.class_names)
+        class_probs = np.bincount(labels) / len(labels)
+        imbalance_exists = class_probs.min() < self.fraction * (1 / K)
+        is_issue_column = np.full(len(labels), False)
+        scores = np.ones(len(labels))
+        if imbalance_exists:
+            rarest_class = np.argmin(class_probs)
+            is_issue_column[labels == rarest_class] = True
+            scores[labels == rarest_class] = class_probs[rarest_class]
```
elisno, line 75: I think the `if` block is unnecessary. We're assuming that the `self.datalab.labels` are integers in the range [0, K-1].
We can construct `is_issue_column` and `scores` in one line each.

```suggestion
        rarest_class = np.argmin(class_probs) if imbalance_exists else -1
        is_issue_column = labels == rarest_class
        scores = np.where(is_issue_column, class_probs[rarest_class], 1)
```

tests/datalab/issue_manager/test_imbalance.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def test_find_issues(self, embeddings, issue_manager):
+        issue_manager.find_issues(features=embeddings)
+        issues, summary = issue_manager.issues, issue_manager.summary
+        expected_issue_mask = np.array([True] * 1 + [False] * 49)
+        assert np.all(
+            issues["is_class_imbalance_issue"] == expected_issue_mask
+        ), "Issue mask should be correct"
+        assert summary["issue_type"][0] == "class_imbalance"
+        assert summary["score"][0] == pytest.approx(expected=0.9804, abs=1e-7)
```
elisno, line 36: I'm thinking of two similar tests for `find_issues`?

1. What if there is no rare class. I expect that the issue_mask will be all falses. Will the scores just be ones? I think the test case will look similar to `test_find_issues`
2. What if there is more than one "rare" classe? Let's make `K-1` still the _most_ rare class, but what do you suggest we do about the "other rare class" in the test suite for now? Suppose `K-2` is the second rarest class, then we could check that it's not flagged as having an issue and it's score is still set to 1?

elisno (commented):
This is a great start @tataganesh!

Just leaving some suggestions for the source code and some requests for the tests.

elisno:
While the `ClassImbalanceIssueManager` is not run by default, it is available to `Datalab` when registered in `factory.py`.
This is why we're getting some failing unit tests, like here ([link](https://github.com/cleanlab/cleanlab/actions/runs/5407245137/jobs/9825037298?pr=758)).

cleanlab/datalab/issue_manager/imbalance.py
tataganesh: Good point, I am actually not sure. Would you have any suggestions for this? (When I pushed the code I didn't completely understand the function of verbosity levels).

tests/datalab/issue_manager/test_imbalance.py
tataganesh: 1. That makes sense, yes all the scores will be one. 
2. I am assuming that in the presence of another rare class with `K-1` still being the rarest class, tests for both issue_mask and scores could be exactly same as the one in `test_find_issues`? Do we explicitly need to check for the `K-2` class having an issue, and it's score being 1?

cleanlab/datalab/issue_manager/imbalance.py
elisno: For now, you can just 


1. Rename `fraction` -> `threshold`
2. Update the class docstring accordingly.
3. Remove `threshold` from the verbosity_levels for now. This only matters when running `Datalab.report()`.
4. `find_issues` should set the `self.info` dictionary. You could add a `{"threshold": threshold}` to `self.info`. We usually extract the code to the `collect_info` method ([see docs](https://docs.cleanlab.ai/stable/cleanlab/datalab/issue_manager/issue_manager.html#cleanlab.datalab.issue_manager.issue_manager.IssueManager.collect_info)). Below is an example of how this is done in the `NearDuplicatesIssueManager`.

https://github.com/cleanlab/cleanlab/blob/1575ba743231dc741076cfc6c37f195f27e80442/cleanlab/datalab/issue_manager/duplicate.py#L113
https://github.com/cleanlab/cleanlab/blob/1575ba743231dc741076cfc6c37f195f27e80442/cleanlab/datalab/issue_manager/duplicate.py#L159-L188

tests/datalab/issue_manager/test_imbalance.py
elisno: Re 2: I think we can just reuse the issue mask, but I'd like an explicit test that only _one_ class has an issue.

tataganesh: I see. So something like `assert np.sum(issues["is_class_imbalance_issue"]) == 1` should do?

elisno: Yeah, in fact, you can do this in the existing test as well!

tataganesh:
@elisno I have pushed the requested changes, along with the incremental search test in `test_datalab.py`.

tests/datalab/test_datalab.py
```diff
@@ -932,3 +932,41 @@ def test_find_issues_features_works_with_and_without_labels(self, features, labe
...
+        lab.find_issues(pred_probs=pred_probs, issue_types={"label": {}})
+        summary = lab.get_issue_summary()
+        assert len(summary) == 1
+        assert "class_imbalance" not in summary["issue_type"].values
+        lab.find_issues(features=random_embeddings, issue_types={"class_imbalance": {}})
+        summary = lab.get_issue_summary()
+        assert len(summary) == 2
+        assert "class_imbalance" in summary["issue_type"].values
+        class_imbalance_summary = lab.get_issue_summary("class_imbalance")
+        assert class_imbalance_summary["num_issues"].values[0] > 0
```
elisno, line 972: This is a very comprehensive test.

It seems to me that `ClassImbalanceIssues.find_issues` doesn't require any arguments to be run (it just needs the labels provided in the associated `Datalab` instance.

Can you make a similar test that just runs

```python
 lab = Datalab(data=data, label_name="labels")
lab.find_issues( issue_types={"class_imbalance": {}})
```

I expect the resulting class imbalance issues to be the same as in this test.

elisno:
> What would I need to do to fix the failing unit tests?

Ahh, you just have to add the issue type to this test:

https://github.com/cleanlab/cleanlab/blob/1575ba743231dc741076cfc6c37f195f27e80442/tests/datalab/test_factory.py#L16

Now that you bring this up, the variable name in the test is not accurate. It's just the issue types that are expected to be "possible", not defaults.

tests/datalab/test_datalab.py
tataganesh: Got it. Is this part of the incremental search or a separate test?

elisno: A separate test. Some code repetition is fine here.

tataganesh:
Test for class imbalance find issues without args (i.e. without features, pred_probs and knn_graph) has been added and the variable name in `test_factory` has been fixed.

tests/datalab/test_datalab.py
tataganesh: Related question: For `ClassImbalanceIssues.find_issues` tests (i.e. test_imbalance.py), I do pass the features to the function. Should I remove them from those tests, and subsequently remove `kwargs` from the `find_issues` function of `ClassImbalanceIssueManager`? Since I can add `"class_imbalance":{}`  here - 

https://github.com/cleanlab/cleanlab/blob/1575ba743231dc741076cfc6c37f195f27e80442/cleanlab/datalab/issue_finder.py#L217C1-L221

jwmueller:
@tataganesh It seems some unit tests are still not passing. Could you fix those before our team does another review of the PR?  Thanks!

tataganesh:
@Elías Snorrason @jwmueller I am looking at the[ failing test](https://github.com/cleanlab/cleanlab/actions/runs/5439088625/jobs/9905753216?pr=758) and I was able to reproduce it locally, although I am not sure if my reasoning is right (since I am getting a similar issue for python 3.8 in the master branch as well).


The issue seems to be arising in python 3.8. On creating a venv with python 3.8 locally, I noticed that the test fails. For python 3.9 it passes. While performing a diff between versions of packages install for 3.8 and 3.9, I noticed that 3.9 installs `scikit-learn==1.2.2` and 3.8 installs scikit-learn==1.3.0. (I am not sure if this is just in my local system). 

Now, the failing test is `test_aux_inputs` in `test_classification.py`, for this assert
https://github.com/cleanlab/cleanlab/blob/1575ba743231dc741076cfc6c37f195f27e80442/tests/test_classification.py#L259

Looking at closer at `classification.py` (the fit method), I noticed these lines:
https://github.com/cleanlab/cleanlab/blob/1575ba743231dc741076cfc6c37f195f27e80442/cleanlab/classification.py#L531-L533

Strangely, for scikit-learn 1.2.2, `"sample_weight" in inspect.getfullargspec(self.clf.fit).args` gives `True` whereas for scikit-learn 1.3.0, it gives `False`.


Confused, I ran the following snippet for both python envs :

```python
import numpy as np
import inspect


X = np.random.randn(19,1)
y = np.random.choice([0,1], 19)

clf=LogisticRegression(multi_class="auto", solver="lbfgs", random_state=0)
clf.fit(X, y)

print(inspect.getfullargspec(clf.fit).args)
```
`scikit-learn 1.2.2` - ['self', 'X', 'y', 'sample_weight']
`scikit-learn 1.3.0` - ['estimator']

And the reason for this seems to be the inclusion of a decorator for the fit function in scikit-learn [here](https://github.com/scikit-learn/scikit-learn/blob/7f9bad99d6e0a3e8ddf92a7e5561245224dab102/sklearn/linear_model/_logistic.py#L1138-L1140). The inspection function is returns the args of the decorator and not the fit function. (Again, not completely sure about this). 
I believe it is the same issue for the regression as well.
https://github.com/cleanlab/cleanlab/blob/1575ba743231dc741076cfc6c37f195f27e80442/cleanlab/regression/learn.py#L266-L270

I might have gone into the wrong rabbit-hole to resolve the unit-test, but this is the only reason I could find for the failing test.

jwmueller:
@tataganesh Don't forget to add a brief description of this new issue type in the Datalab Issue Type guide:

https://docs.cleanlab.ai/stable/cleanlab/datalab/guide/issue_type_description.html

https://github.com/cleanlab/cleanlab/blob/master/docs/source/cleanlab/datalab/guide/issue_type_description.rst

tests/datalab/test_datalab.py
elisno: @tataganesh Good question!

Removing `features` from the issue manager tests would be good, but I believe keeping `kwargs` is safer for now.

I don't think it's necessary to update the `args_dict` in the linked code.

tataganesh:
@elisno I am not sure why there is a decrease in coverage (it is the only check not passing). Can you help?

docs/source/cleanlab/datalab/guide/issue_type_description.rst
```diff
@@ -110,6 +110,17 @@ The assumption that examples in a dataset are Independent and Identically Distri
...
 
+Class-Imbalance Issue
+---------------------
+
+Class Imbalance issues are determined using the `labels` provided as part of the dataset. First, the proportion of samples belonging to each class in the dataset is calculated.
+If `q` is the proportion of samples of the rarest class and `q` is lesser than `threshold` `c,` where `0 < c < 1`, then the dataset is imbalanced, and the class imbalance examples are all samples
+belonging to the rarest class. For such examples, the class imbalance quality score is `q`, and it is `1.0` for all other examples.
+
+Class Imbalance issues in a dataset can lead to sub-par model performance for the under-represented class. To handle such issues,
+you can increase the number of samples belonging to the under-represented class.
```
jwmueller, line 121:
```suggestion
Class imbalance is diagnosed just using the `labels` provided as part of the dataset. The overall class imbalance quality score of a dataset is the proportion of examples belonging to the rarest class `q`. If this proportion `q` falls below a threshold, then we say this dataset suffers from the class imbalance issue.  

In a dataset identified as having class imbalance, the class imbalance quality score for each example is set equal to `q` if it is labeled as the rarest class, and is equal to 1 for all other examples.

Class imbalance in a dataset can lead to subpar model performance for the under-represented class. Consider collecting more data from the under-represented class, or at least take special care while modeling via techniques like over/under-sampling, SMOTE, asymmetric class weighting, etc.
```

elisno (approved):
This looks great @tataganesh! 
This PR is an excellent example for future contributors.
Impressive work!

elisno:
> @elisno I am not sure why there is a decrease in coverage (it is the only check not passing). Can you help?

This check sometimes reports a decrease in coverage without a good reason.
In this case, since all the other checks have passed and the code has been thoroughly reviewed and tested, I wouldn't consider this a failing check.

The currently failing CI checks are unrelated to this PR.

PR #760, opened 2023-07-04
Label Issue Detection Extention

### LINKED ISSUE #735: Extend label issue detection in Datalab to work even without pred_probs input
Goal: extend the label issue check in Datalab to work even if user only provided: `features`, `labels` to `Datalab.find_issues()`.

There are multiple ways this can be achieved:

Option 1 (easiest): Use sklearn `KNNclassifier` (or `LogisticRegression`) applied to `X=features, y=labels` in order to produce out-of-sample `pred_probs` and then continue as usual.

Option 2: Use methods from other papers like these (requires benchmarking them first):
- [SelfClean: A Self-Supervised Data Cleaning Strategy](https://arxiv.org/abs/2305.17048)
- [Detecting Corrupted Labels Without Training a Model to Predict](https://arxiv.org/abs/2110.06283)

Steven-Yiran:
Closes #735 to extend the label issue checks in Datalab to work even if users only provided: `features`, `labels` to `Datalab.find_issues()`.


jwmueller:
Thanks for another awesome contribution!  
Our team can review it once you get all unit tests passing, and remove the unused imports.

Steven-Yiran:
I am not sure why the CI tests would fail on Ubuntu, the tests all passed on my local MacOS machine.

huiwengoh:
The unittests are failing on something unrelated, don't worry about it for now I'll take a look at that

Steven-Yiran:
Thanks a lot for the comments. I just fixed the unit test issues.

cleanlab/datalab/issue_manager/label.py
```diff
@@ -63,6 +64,7 @@ def __init__(
         datalab: Datalab,
         clean_learning_kwargs: Optional[Dict[str, Any]] = None,
         health_summary_parameters: Optional[Dict[str, Any]] = None,
+        k: int = 10,
```
huiwengoh, line 67: Could you add this new argument to the docstring above explaining what it is for?

cleanlab/datalab/issue_manager/label.py
```diff
@@ -118,9 +121,19 @@ def _reset(self) -> None:
...
         **kwargs,
     ) -> None:
+        if pred_probs is None and features is None:
+            raise ValueError("Either pred_probs or features must be provided to find label issues.")
+        if pred_probs is None:
+            # produce out-of-sample pred_probs from features
+            knn = KNeighborsClassifier(n_neighbors=self._k)
+            pred_probs = cross_val_predict(
+                knn, features, self.datalab.labels, cv=5, method="predict_proba"
+            )
```
huiwengoh, line 135: [nit] these checks could be condensed a little nicer:

```suggestion
        if pred_probs is None:
            if features is None:
                raise ValueError("Either pred_probs or features must be provided to find label issues.")
            # produce out-of-sample pred_probs from features
            knn = KNeighborsClassifier(n_neighbors=self._k)
            pred_probs = cross_val_predict(
                knn, features, self.datalab.labels, cv=5, method="predict_proba"
            )
```

cleanlab/datalab/issue_manager/label.py
huiwengoh: Could you also add quick unittest to test passing `k` in as an arg in the `Datalab.find_issues` call?

tests/datalab/test_datalab.py
```diff
@@ -809,14 +809,25 @@ def test_incremental_search(self, pred_probs, random_embeddings):
...
+            features=random_embeddings, pred_probs=pred_probs, issue_types={"label": {}}
+        )
+        summary = lab.get_issue_summary()
+        assert len(summary) == 4
+        assert "label" in summary["issue_type"].values
+        label_summary_both = lab.get_issue_summary("label")
+        assert (
+            label_summary_both["num_issues"].values[0]
+            == label_summary_pred_probs["num_issues"].values[0]
+        )
```
huiwengoh, line 830: [nit] I'm not entirely sure if these tests fall under `test_incremental_search`, I think checking that executing `find_issues` with pred_probs and executing `find_issues` with both returns the same result falls under a different test

cleanlab/datalab/issue_manager/label.py
```diff
@@ -118,9 +121,19 @@ def _reset(self) -> None:
...
         **kwargs,
     ) -> None:
+        if pred_probs is None:
+            if features is None:
+                raise ValueError("Either pred_probs or features must be provided to find label issues.")
+            # produce out-of-sample pred_probs from features
+            knn = KNeighborsClassifier(n_neighbors=self._k)
+            pred_probs = cross_val_predict(
+                knn, features, self.datalab.labels, cv=5, method="predict_proba"
+            )
```
jwmueller, line 135: For efficiency and better accuracy, let's not use cross-validation here. You can instead do:

```
knn.fit(features, self.datalab.labels)
pred_probs = knn.predict_proba(features)
```

Then finally you can adjust these `pred_probs` to account for the fact that the datapoint itself was included amongst its neighbors by doing this adjustment for each example `i`:
```
pred_probs[i] = (pred_probs[i] - 1/self._k * onehotlabel[i]) * self._k/(self._k-1)
```
where `onehotlabel` = zeros vector except for a 1 in the index corresponding to the given label of example `i`. Make sure to test the adjustment (aka my math) by verifying the probs still sum to 1 for each example `i`.

Ideally you should vectorize the above adjustment to `pred_probs` across all examples via numpy. The above code is also 2x slower because I call `fit()` and `predict_proba()` separately even though it is technically not required, but I think we can tolerate this inefficiency for ease of implementation.

Steven-Yiran: Thanks for the detailed explanation!

cleanlab/datalab/issue_manager/label.py
Steven-Yiran: I am a bit confused by the expected behavior of passing `k` in as an arg in the Datalab.find_issues(). Should the expected behavior be an unexpected keyword argument? Since k is only defined in __init__().

cleanlab/datalab/issue_manager/label.py
```diff
@@ -118,9 +124,36 @@ def _reset(self) -> None:
...
+            if features is None:
+                raise ValueError(
+                    "Either pred_probs or features must be provided to find label issues."
+                )
+            # produce out-of-sample pred_probs from features
+            knn = KNeighborsClassifier(n_neighbors=self.k)
+            knn.fit(features, self.datalab.labels)
+            pred_probs = knn.predict_proba(features)
+
+            encoder = OneHotEncoder(sparse_output=False)
```
huiwengoh, line 151:
```suggestion
            encoder = OneHotEncoder()
```

This `sparse_output` arg is causing the unittests to fail for `python 3.7` as this arg was named something different in `sklearn==1.0` (which is the latest version supported on `python 3.7`, ref [here](https://scikit-learn.org/1.0/modules/generated/sklearn.preprocessing.OneHotEncoder.html)).

I also believe it is not required to force a non-sparse matrix here as the multiplication below will work just fine, so could just remove entirely.

cleanlab/datalab/issue_manager/label.py
```diff
@@ -118,9 +124,36 @@ def _reset(self) -> None:
...
+            # produce out-of-sample pred_probs from features
+            knn = KNeighborsClassifier(n_neighbors=self.k)
+            knn.fit(features, self.datalab.labels)
+            pred_probs = knn.predict_proba(features)
+
+            encoder = OneHotEncoder()
+            label_transform = self.datalab.labels.reshape(-1, 1)
+            one_hot_label = encoder.fit_transform(label_transform)
+
+            pred_probs = (pred_probs - 1 / self.k * one_hot_label) * self.k / (self.k - 1)
```
huiwengoh, line 155:
```suggestion
            pred_probs = np.asarray((pred_probs - 1 / self.k * one_hot_label) * self.k / (self.k - 1))
```

My suggested edit yesterday would make `pred_probs` be of type matrix (as we multiplied it with `one_hot_label` which is a sparse matrix), to make it compatible with the future functions we need to convert it back into an array.

cleanlab/datalab/issue_manager/label.py
Steven-Yiran: Thanks for the clarification! I was confused about how to pass `k`

huiwengoh: No worries! For more details you can also check out [this tutorial](https://docs.cleanlab.ai/stable/tutorials/datalab/datalab_advanced.html#Functionality-2:-Specifying-nondefault-arguments)

cleanlab/datalab/issue_manager/label.py
```diff
@@ -118,9 +126,38 @@ def _reset(self) -> None:
...
+
+        features :
+            The features for each example.
+        """
+        if pred_probs is None:
+            if features is None:
+                raise ValueError(
+                    "Either pred_probs or features must be provided to find label issues."
+                )
+            # produce out-of-sample pred_probs from features
```
huiwengoh, line 148:
```suggestion
            # computing pred_probs from the given features
```

We are not getting out-of-sample pred_probs anymore

tests/datalab/test_datalab.py
```diff
@@ -809,15 +809,46 @@ def test_incremental_search(self, pred_probs, random_embeddings):
...
+
+    def test_pred_probs_precedence(self, pred_probs, random_embeddings):
+        data = {"labels": np.random.randint(0, 2, 100)}
+        lab = Datalab(data=data, label_name="labels")
+        lab.find_issues(pred_probs=pred_probs, issue_types={"label": {}})
+        summary = lab.get_issue_summary()
+        assert "label" in summary["issue_type"].values
+        label_summary_pred_probs = lab.get_issue_summary("label")
+        assert label_summary_pred_probs["num_issues"].values[0] > 0
+        lab.find_issues(
```
huiwengoh, line 841: I would instantiate a new `Datalab` instance here just to make sure it isn't just returning something that was computed previously (it shouldn't, each `find_issues` call should theoretically rewrite the existing results, but just to make sure)

cleanlab/datalab/issue_manager/label.py
```diff
@@ -37,6 +40,9 @@ class LabelIssueManager(IssueManager):
     datalab :
         A Datalab instance.
 
+    k :
+        The number of nearest neighbors to consider when computing pred_probs from features.
```
huiwengoh, line 44:
```suggestion
    k :
        The number of nearest neighbors to consider when computing pred_probs from features.
        Only applicable if features are provided and pred_probs are not.
```

huiwengoh (commented):
Nearly there! Just requested a few minor changes, otherwise it's looking good!

cleanlab/datalab/issue_manager/label.py
```diff
@@ -118,9 +126,38 @@ def _reset(self) -> None:
...
+            knn.fit(features, self.datalab.labels)
+            pred_probs = knn.predict_proba(features)
+
+            encoder = OneHotEncoder()
+            label_transform = self.datalab.labels.reshape(-1, 1)
+            one_hot_label = encoder.fit_transform(label_transform)
+
+            pred_probs = np.asarray(
+                (pred_probs - 1 / self.k * one_hot_label) * self.k / (self.k - 1)
+            )
```
Steven-Yiran, line 159: I think I am doing the out-of-sample `pred_probs` here according to the method described in the earlier comment https://github.com/cleanlab/cleanlab/pull/760#discussion_r1256490110

cleanlab/datalab/issue_manager/label.py
jwmueller: @huiwengoh They are still out-of-sample pred_probs, just computed via leave-one-out cross-validation in a much more efficient matter.

@huiwengoh  Please check the math carefully, I wrote down the formula very hurriedly. Ensure the math/code is equivalent to doing leave-one-out cross-validation with a KNN model (you can even run a mini code test on toy dataset)

Steven-Yiran: This is the script I used to check the results. It gives about 5% differences between the class predicted by leave-one-out and by the KNN method. 
https://github.com/Steven-Yiran/cleanlab/blob/test/check_equal.ipynb

jwmueller: Sorry there was a bug in my math if the current code does not agree with cross_val_predict. More strongly than the predictions agreeing is the pred_probs output by both methods should agree too.

However I spot many issues in your script so I'm not sure it is conclusive:
1. you are setting `k=5` and `n_neighbors=10` but these have to agree (or perhaps `k` should actually be `n_neighbors` + or - 1, need to check math closely to determine).
2. you set `cv=5` in the cross_val_predict, but it should be equal to the amount of data (or minus 1) to mimic leave-one-out cross-validation.

@huiwengoh please test this too and confirm, thanks!

Steven-Yiran: Thanks for pointing out my errors. I have updated the script above just for reference.

cleanlab/datalab/issue_manager/label.py
huiwengoh: Ah yes my bad sorry, thanks for pointing out! Can just keep the comment as is.

cleanlab/datalab/issue_manager/label.py
huiwengoh: @jwmueller your math is right, by omitting the datapoint itself it is equivalent to doing a KNN with `k-1` neighbours. 

Perhaps the something we want to change is to add one to the `n_neighbors` arg to the KNN initializer (ie. `knn = KNeighborsClassifier(n_neighbors=self.k + 1)`) so we adjust for the correct number of neighbours after omitting one datapoint later.

cleanlab/datalab/issue_manager/label.py
```diff
@@ -118,9 +127,38 @@ def _reset(self) -> None:
...
+        features :
+            The features for each example.
+        """
+        if pred_probs is None:
+            if features is None:
+                raise ValueError(
+                    "Either pred_probs or features must be provided to find label issues."
+                )
+            # produce out-of-sample pred_probs from features
+            knn = KNeighborsClassifier(n_neighbors=self.k)
```
huiwengoh, line 150:
```suggestion
            knn = KNeighborsClassifier(n_neighbors=self.k + 1)
```

To adjust for omitting one datapoint later (to get out-of-sample pred_probs). Ref: https://github.com/cleanlab/cleanlab/pull/760#discussion_r1264730502

cleanlab/datalab/issue_manager/label.py
```diff
@@ -118,9 +127,38 @@ def _reset(self) -> None:
...
+            knn.fit(features, self.datalab.labels)
+            pred_probs = knn.predict_proba(features)
+
+            encoder = OneHotEncoder()
+            label_transform = self.datalab.labels.reshape(-1, 1)
+            one_hot_label = encoder.fit_transform(label_transform)
+
+            pred_probs = np.asarray(
+                (pred_probs - 1 / self.k * one_hot_label) * self.k / (self.k - 1)
+            )
```
huiwengoh, line 160:
```suggestion
            # adjust pred_probs so it is out-of-sample
            pred_probs = np.asarray(
                (pred_probs - 1 / (self.k + 1) * one_hot_label) * (self.k + 1) / self.k
            )
```

huiwengoh (approved):
I made some small math edits to ensure the number of neighbours for the KNN classifier is correct and the `pred_probs` are properly adjusted (ref https://github.com/cleanlab/cleanlab/pull/760#discussion_r1264730502).

Otherwise this PR looks great! Thanks @Steven-Yiran for another amazing contribution!

Steven-Yiran:
Thanks a lot for the helpful edits and comments!

PR #761, opened 2023-07-05
change inspect.getfullargspec to signature

huiwengoh:
sklearn's recent [v1.3 release](https://github.com/scikit-learn/scikit-learn/releases/tag/1.3.0) added decorators in some of the model's fit function (eg. [here](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/linear_model/_base.py#L650) in `linear_model`).

The `inspect.getfullargspec()` function we use to get the arguments of a function does not handle decorated functions properly (ref: https://github.com/joblib/joblib/issues/1164) and hence some checks are not working as intended in some `CleanLeaning` methods. Updating these methods to call `inspect.signature()` instead to better handle getting args from these decorated functions.


jwmueller (approved):
Thank you for fixing!

PR #763, opened 2023-07-06
Improve hidden block check in tabular tutorial

huiwengoh:
The outputs of the tabular tutorial has been slightly inconsistent lately, I believe due to the lack of setting the random seed in the ExtraTrees model (noticed this briefly when debugging the nbsphinx CI checks earlier on, but due to nbsphinx suppressing exceptions I believe it was not being flagged properly). 

It recently did not pass the CI check, so setting the seeds more strictly now to make the outputs consistent.

### DIFF
--- a/docs/source/tutorials/tabular.ipynb
+++ b/docs/source/tutorials/tabular.ipynb
@@ -153,7 +153,7 @@
     "from cleanlab.filter import find_label_issues\n",
     "from cleanlab.classification import CleanLearning\n",
     "\n",
-    "SEED = 1234 \n",
+    "SEED = 100 \n",
     "\n",
     "np.random.seed(SEED)\n",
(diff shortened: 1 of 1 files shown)


jwmueller (approved):
Good catch! feel free to merge whenever

huiwengoh:
Seems like it's still a little flaky (it passed locally + on Colab but seems to still be failing on CI), I'll look into it deeper tomorrow and might just remove the specific example mentions and make the hidden cell checks a little less strict if it's still an issue

jwmueller:
if flaky you can instead refrain from referring to any specific examples in the markdown cells. Also switch the test from using exact set inclusion to a jaccard similarity between the set of label issues and true errors being above some threshold.

PR #768, opened 2023-07-09
Fix links and link checker for Markdown files

anishathalye:
This patch improves the regex so it matches paths starting with "../" and paths with hashes.


huiwengoh (approved):
LGTM, thanks for fixing!

PR #769, opened 2023-07-12
Add analytics for docs website

anishathalye:
Note that this won't backfill the analytics code onto old pages, but that's probably not worth the effort at the moment (and we can always do it later/separately if we want).


jwmueller (approved):
thanks!

anishathalye:
Confirmed that analytics code is deployed at https://docs.cleanlab.ai/master/.

PR #775, opened 2023-07-14
Segmentation docs fix

elisno:
- Some internal links were broken. Should be fixed now.

- Use special block for tips:

image

- Format class_names example as a code block:
image

- Put note in a block
image

### DIFF
--- a/cleanlab/segmentation/summary.py
+++ b/cleanlab/segmentation/summary.py
@@ -55,12 +55,12 @@ def display_issues(
       Optional discrete array of noisy labels for a segmantic segmentation dataset, in the shape``(N,H,W,)``.
       where each pixel must be integer in 0, 1, ..., K-1.
       If `labels` is provided, this function also displays given label of the pixel identified with issue.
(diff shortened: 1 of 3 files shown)


cleanlab/segmentation/summary.py
```diff
@@ -215,7 +222,7 @@ def common_label_issues(
     Returns
     -------
     issues_df:
-      DataFrame `issues_df` contains columns ``['given_label', 'predicted_label', 'num_label_issues']`` and each row contains information for a
+      DataFrame containing columns ``['given_label', 'predicted_label', 'num_label_issues']`` and each row contains information for a
       given/predicted label swap, ordered by the number of label issues inferred for this type of label swap.
```
jwmueller, line 226:
```suggestion
      DataFrame with columns ``['given_label', 'predicted_label', 'num_label_issues']``
      where each row contains information about a particular given/predicted label swap.
      Rows are ordered by the number of label issues inferred to exhibit this type of label swap.
```

jwmueller (approved):
Great catches!  Just had one suggestion but go ahead and merge after considering it

PR #778, opened 2023-07-21
Add test checking that find_label_issues does not flag examples where given_label==prediction

huiwengoh:


jwmueller (approved):
Thanks this will be useful to have!  
I was referring to add a stronger test of this via [property-based testing](https://hypothesis.readthedocs.io/en/latest/), but this will suffice for now I think.

PR #781, opened 2023-07-25
Update near-duplicate sets

elisno:
After this PR:
- Examples that are not flagged with `is_near_duplicate_issue = True` will have an empty near-duplicate set.
- The near-duplicates are now symmetric.

### DIFF
--- a/cleanlab/datalab/issue_manager/duplicate.py
+++ b/cleanlab/datalab/issue_manager/duplicate.py
@@ -135,6 +135,15 @@ class NearDuplicateIssueManager(IssueManager):
         indices = knn_graph.indices[mask.ravel()]
         near_duplicate_sets = [indices[indptr[i] : indptr[i + 1]] for i in range(N)]
 
+        # Second pass over the data is required to ensure each item is included in the near-duplicate sets of its own near-duplicates.
+        # This is important because a "near-duplicate" relationship is reciprocal.
+        # For example, if item A is a near-duplicate of item B, then item B should also be considered a near-duplicate of item A.
+        # NOTE: This approach does not assure that the sets are ordered by increasing distance.
+        for i, near_duplicates in enumerate(near_duplicate_sets):
+            for j in near_duplicates:
+                if i not in near_duplicate_sets[j]:
+                    near_duplicate_sets[j] = np.append(near_duplicate_sets[j], i)
+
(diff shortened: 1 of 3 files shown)


cleanlab/datalab/issue_manager/duplicate.py
```diff
@@ -125,7 +125,7 @@ def _neighbors_within_radius(knn_graph: csr_matrix, radius: float):
         N = knn_graph.shape[0]
         distances = knn_graph.data.reshape(N, -1)
         # Create a mask for the threshold
-        mask = distances < radius
```
jwmueller, line 128: can you confirm the set of examples flagged as near duplicates has not changed, i.e. this PR only updates the sets of near_duplicates in the info object?

tests/datalab/issue_manager/test_duplicate.py
```diff
@@ -80,3 +104,49 @@ def test_report(self, issue_manager, embeddings):
...
+        assert all(
+            len(near_duplicate_set) == 0
+            for i, near_duplicate_set in enumerate(near_duplicate_sets)
+            if not issues[i]
+        ), "Non-issue examples should not have near duplicate sets"
+        assert all(
+            len(near_duplicate_set) > 0
+            for i, near_duplicate_set in enumerate(near_duplicate_sets)
+            if issues[i]
+        ), "Issue examples should have near duplicate sets"
```
jwmueller, line 152: think you could easily combine these into one test for runtime efficiency, but not important if it's not easy to do

jwmueller (approved):
Note that `black` format is not passing in CI, so update that first.

Otherwise LGTM, feel free to merge once you confirm the questions I had!

tests/datalab/issue_manager/test_duplicate.py
elisno: Combined tests as they share the same setup code.

cleanlab/datalab/issue_manager/duplicate.py
elisno: The `issues` dataframe that flags near duplicates is set only once in the `find_issues` method. This occurs before the call to the `_neighbors_within_radius` which computes the sets of near-duplicates (which ends up in the info object).

I can confirm: The flagged examples have **not changed**.

PR #784, opened 2023-07-27
Add more descriptive error messages for multiannotator modules

huiwengoh:
Adds an explicit check + error message to ensure `pred_probs` are numpy arrays

### DIFF
--- a/cleanlab/internal/multiannotator_utils.py
+++ b/cleanlab/internal/multiannotator_utils.py
@@ -154,7 +154,7 @@ def assert_valid_pred_probs(
 
         if pred_probs_unlabeled is not None:
             if not isinstance(pred_probs_unlabeled, np.ndarray):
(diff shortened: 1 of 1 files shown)


ulya-tkch:
I would also add this check directly into `get_majority_vote_label()` since it seems the "assert_valid_inputs" tests are not triggered if `verbose=False`. Otherwise LGTM!

ulya-tkch (approved):
Error message clear and triggered in most user facing functions (see comment above). Lgtm

huiwengoh:
I think codecov is alright (we don't usually check for defensive/error checking code).

For `get_majority_vote_label()`, imo we don't need to add another explicit check because if users set `verbose=False` a whole bunch of other checks are not being done anyways. Also we set `verbose=True` by default and also warn users to not set `verbose=False` here: https://github.com/cleanlab/cleanlab/blob/master/cleanlab/multiannotator.py#L869

PR #791, opened 2023-08-02
Expose low memory option for finding label issues in classification

### LINKED ISSUE #762: integrate mini-batch find_label_issues_batched as an option for methods that internally call find_label_issues
Many methods throughout the package internally call [cleanlab.filter.find_label_issues](https://docs.cleanlab.ai/stable/cleanlab/filter.html#cleanlab.filter.find_label_issues). This method requires a significant amount of memory, so another version of this method has been created with much lower memory requirements that also runs faster: [cleanlab.experimental.label_issues_batched.find_label_issues_batched](https://docs.cleanlab.ai/stable/cleanlab/experimental/label_issues_batched.html#cleanlab.experimental.label_issues_batched.find_label_issues_batched)

This is achieved by computing the estimates via iterative mini-batching and certain approximations.

**Goal**: For each method that internally calls `filter.find_label_issues()`, we should offer an optional argument: `low_memory = False` by default, which if set to `True` internally calls `experimental.label_issues_batched.find_label_issues_batched()` instead.  Thus users who need to run things with reduced memory/runtime can simply specify `low_memory=True` in whichever  method they are using and internally it will run things differently to help them.

tataganesh:
#762 - find_label_issues in cleanlab classification. 

### Example Usage
```python
from cleanlab import Datalab
X = np.random.rand(200, 3)
y = np.random.choice([0,1,2], X.shape[0])
data = {"X": X, "y": y}
pred_probs = np.random.rand(X.shape[0], 3)
pred_probs = pred_probs / pred_probs.sum(axis=1, keepdims=True)

lab = Datalab(data=data, label_name="y")
issue_types = {"label": {"clean_learning_kwargs": {"low_memory": True}}}
lab.find_issues(pred_probs=pred_probs, issue_types=issue_types)
```
The following results were obtained using the [MNIST Classification example](https://github.com/cleanlab/cleanlab/blob/master/docs/source/tutorials/image.ipynb).

### Runtime Comparison
The testing was performed on a MAC M1 Machine (CPU). 

Without Low Memory 
```
Audit complete. 141 issues found in the dataset.
3.3 s ± 1.29 s per loop (mean ± std. dev. of 3 runs, 1 loop each)
```

With Low Memory
```
Audit complete. 123 issues found in the dataset.
132 ms ± 11 ms per loop (mean ± std. dev. of 3 runs, 1 loop each)
```

### Results Comparison

Filter Dataframes to only consider samples with issue. 
```python
get_df_with_issue = lambda df: df[df["is_label_issue"]==True]
ori_issues = get_df_with_issue(issue_results) 
lm_issues = get_df_with_issue(issue_results_lm) # With low_memory=True
```

Calculate Metrics
```python
lm_index = lm_issues.index
ori_index = ori_issues.index
intersection = len(set(lm_index).intersection(set(ori_index)))
union = len(lm_index) + len(ori_index) - intersection
print(f"Jaccard Similarity - {intersection / union}")
print(f"Intersection - {intersection} ({intersection*100.0/len(ori_index):.2f}%)")
print(f"Mean Scores - {lm_issues.label_score.mean(), ori_issues.label_score.mean()}")
```

|                                            | Original  | Low Memory |
|--------------------------------------------|:---------:|:----------:|
|              Number of Issues              |    141    |     123    |
| Mean Score for<br>samples with label issue | <br>0.015 | <br>0.0039 |
|               Issues Overlap               |     -     | 67 (47.2%) |
|             Jaccard Similarity             |     -     |    0.34    |

###  Plotting Samples with label issue and low quality
```python
ranked_label_issues = ori_issues.sort_values("label_score").index
plot_examples(ranked_label_issues[:15], 3, 5)
```
image

```python
# with low memory
ranked_label_issues_lm = lm_issues.sort_values("label_score").index
plot_examples(ranked_label_issues_lm[:15], 3, 5)
```

Plotting Samples with label issue (low memory)
image


cleanlab/classification.py
```diff
@@ -218,6 +219,10 @@ class CleanLearning(BaseEstimator):  # Inherits sklearn classifier
     verbose : bool, default=False
       Controls how much output is printed. Set to ``False`` to suppress print
       statements.
+
+    low_memory: bool, default=False
+      Uses :py:func:`experimental.label_issues_batched.find_label_issues_batched <cleanlab.experimental.label_issues_batched>`
```
huiwengoh, line 224:
```suggestion
      Set as ``True`` if you have a big dataset with limited memory.
      Uses :py:func:`experimental.label_issues_batched.find_label_issues_batched <cleanlab.experimental.label_issues_batched>`
```

cleanlab/classification.py
```diff
@@ -113,6 +113,7 @@ def score(self, X, y, sample_weight=None):
 the :math:`\\tilde{y}` used in confident learning paper.
 """
 
+from cleanlab.experimental.label_issues_batched import find_label_issues_batched
```
huiwengoh, line 116: [nit] I would move this import a little lower (perhaps line 152) to group it together with the rest of the cleanlab imports - makes it a little neater

cleanlab/classification.py
```diff
@@ -894,11 +901,16 @@ def find_label_issues(
         labels = labels_to_array(labels)
         if self.verbose:
             print("Using predicted probabilities to identify label issues ...")
-        label_issues_mask = filter.find_label_issues(
-            labels,
-            pred_probs,
-            **self.find_label_issues_kwargs,
-        )
+        if self.low_memory:
```
huiwengoh, line 904: Since `find_label_issues_kwargs` is not being used  when `low_memory=True`, perhaps we want to raise a warning (or print statement) informing users that `find_label_issues_kwargs` is not used in this scenario (tho only raised if `find_label_issues_kwargs != {}`).

`find_label_issues_batched` also has additional kwargs but supporting that could be messy/a little complicated so I think it is not the most important to add, for now I think the warning is good enough

huiwengoh (commented):
Thanks for the contribution! 
This looks super good, just had some small comments/suggested improvements

cleanlab/classification.py
tataganesh: https://github.com/cleanlab/cleanlab/blob/0cd40abbb28f4a92b2cea3813b806ebd968cf9ac/cleanlab/classification.py#L785
This function seems to be adding some default keyword arguments like `min_examples_per_class` to `find_label_issues_kwargs`. Does thie mean that the warning should always be shown when `low_memory=True` or is it only applicable for non-default (explicit) kwargs passed via `find_label_issues_kwargs`?

huiwengoh: Yes the latter is right, warning should be raised when a user explicitly passes in `find_label_issues_kwargs`.
ie. on [line 792](https://github.com/cleanlab/cleanlab/blob/c8ad4fd589f14369cf2cc36eb16b395dcc809c79/cleanlab/classification.py#L792) here you can do something like
```
if low_memory == True and len(self.find_label_issues_kwargs) != 0:
    # print some warning message
else:
    self._process_label_issues_kwargs(self.find_label_issues_kwargs)
```

huiwengoh: In fact, this method computes many values such as `confident_joint` and more which are eventually added to `find_label_issue_kwargs` to be passed to `filter.find_label_issues`. 

However when `low_memory=True` they are not being used, hence I believe these rows here are completely irrelevant and should not be executed (they are just wasted computes): 
https://github.com/cleanlab/cleanlab/blob/c8ad4fd589f14369cf2cc36eb16b395dcc809c79/cleanlab/classification.py#L792-L899

My suggestion would be to move the `if self.low_memory` check much earlier and skip any computes that are not necessary (should be just the ones I mentioned above, but would be good to double check if I accidentally ignored too many things, or missed any).

After making that change, if you could rerun your runtime and performance benchmarks (if not too much work) and report back that would be super helpful (and of course make sure this doesn't make thing worse than it was before - ML metrics wise it should not change from what you already have). Thanks!

tataganesh: That makes sense!  When you say "move the if self.low_memory check much earlier and skip any computes that are not necessary", do you mean something like this? (i.e. throw all computation for `filter.find_label_issues` in else). 

```
if self.low_memory:
  # find_label_issues_batched
else:
   # Compute noise matrix, confident joint etc.
   # filter.find label issues
```

huiwengoh: Yup that's right! You can place all the code that is unrelated to `find_label_issues_batched` in the else block.

And yes good catch! `pred_probs` is necessary for either case so we would need to make sure that is checked and computed if necessary.

tataganesh: Thank you! I have two follow-up questions based on initial testing.

**Regarding `pred_probs`**
If `pred_probs` is not passed and it is calculated before checking `low_memory`, then the following code snippet, where both `noise_matrix` and `inverse_noise_matrix` are not passed and `pred_probs is None`, will never be executed. 
https://github.com/cleanlab/cleanlab/blob/c8ad4fd589f14369cf2cc36eb16b395dcc809c79/cleanlab/classification.py#L820-L842
I should also mention that  at this point, the `pred_probs` calculation is being performed _after_ the above snippet, hence it is not dead code. So if this is **not** the intended consequence, then one solution would be to check and calculate `pred_probs` separately within the `low_memory` condition, and otherwise proceed as before.

```python
if low_memory:
    if pred_probs is None:
           # calculate
else:
    # Follow same steps as before
 ```
 
 **Unexpected failing unit-test**
When I was testing the above-mentioned changes, I noticed that the following unit-test is failing.
https://github.com/cleanlab/cleanlab/blob/c8ad4fd589f14369cf2cc36eb16b395dcc809c79/tests/test_classification.py#L742-L744
It is failing for the following code snippet
https://github.com/cleanlab/cleanlab/blob/0cd40abbb28f4a92b2cea3813b806ebd968cf9ac/cleanlab/classification.py#L837-L850
I believe this is happening because when computing the `confident_joint`, the dataframe to array conversion is not being done properly. 
https://github.com/cleanlab/cleanlab/blob/0cd40abbb28f4a92b2cea3813b806ebd968cf9ac/cleanlab/count.py#L578
It converts the labels to `[[[0], [0], [0], [0], [0]]]` instead of `[0 0 0 0 0 0 0 0 0 0 2]`.
Changing this to `labels = labels_to_array(labels)` seems to pass the test. 

If this makes sense, I can make this change as well (and maybe also add a unit-test for this specific case).

Comment (huiwengoh) [reply cleanlab/classification.py:904]: Yes you will have to make sure the `pred_probs` are computed in either case (even if you might have some repeated code). There is also this function here that just extracts the `pred_probs` and not the noise matrix and other things: 
https://github.com/cleanlab/cleanlab/blob/master/cleanlab/count.py#L1070

It computes the same number of things under the hood but it might help with storing less information.

I'm not entire sure why the unittest error you mentioned is popping up after the code change? Currently the unittests are passing and I don't believe any changes you make should alter the shape of `labels`? It's tough for me to diagnose as well without seeing the code change/error, so if you're not sure how to resolve it/need my help you can push the failing commits and let me know if you want me to take a look at it.

**Comment by tataganesh** [comment]
Requested changes - 
* Avoid extra computation when `low_memory=True`.
* Warning for arguments passed to `find_label_issues` that are not used when `low_memory=True`. 
* Add unit-test to show that when cleanlearning's `fit` function is called with a dataframe label and pred_probs,  there is an error. 
* `pred_probs` is calculated for both conditions separately.

File: cleanlab/count.py [hunk 4]
```diff
@@ -576,6 +576,7 @@ def compute_confident_joint(
 
     # labels needs to be a numpy array
     labels = np.asarray(labels)
+    # replacing with labels = labels_to_array(labels) resolves unit-test
```
huiwengoh, line 579: We should not do this as we intentionally want to preserve the nested array format in `labels` for some of the following function calls (eg. `get_confident_thresholds`).

Instead you can call `labels_to_array(labels)` in `classification.find_label_issues` when intended

cleanlab/classification.py
```diff
@@ -782,123 +790,160 @@ def find_label_issues(
...
+            arg_values = {
+                "thresholds": thresholds,
+                "noise_matrix": noise_matrix,
+                "inverst_noise_matrix": inverse_noise_matrix,
+                "find_label_issues_kwargs": self.find_label_issues_kwargs,
+            }
+            for arg_name, arg_val in arg_values.items():
+                if arg_val:
+                    warnings.warn(f"`{arg_name}` is not used when `low_memory=True`.")
+            label_issues_indices = find_label_issues_batched(labels, pred_probs)
```
huiwengoh, line 824: Can add `labels = labels_to_array(labels)` above this line

tataganesh: Sorry, I am a little confused. 
*  In the [first comment](https://github.com/cleanlab/cleanlab/pull/791#discussion_r1286601954), you mention that we want to preserve the nested array format in labels. Doesn't this mean `labels = labels_to_array(labels)` cannot be used anytime before calling `compute_confident_joint`, because it doesn't preserve the nested structure?  If so, how else could I resolve this issue?
* Should `labels = labels_to_array(labels)` be added only for `find_label_issues_batched`? The unit test would still fail when `low_memory` is not True.

huiwengoh: Apologies if my comment was unclear, I took a closer look at the source code and here are some notes:

- The unittest you added actually caught a bug that was present even before the `low_memory` stuff was added - thanks for the great catch! 
  I was initially confused because I thought the `low_memory` addition introduced the bug hence I proposed the solution above.

- When I mentioned this above -
  > We should not do this as we intentionally want to preserve the nested array format in labels for some of the following function calls (eg. `get_confident_thresholds`).

  I mainly meant that we should not alter than function in `count.py` as it is used by many other functions in the library. In general, we should always be cautious about changing the data formats in these commonly used functions. 
  The biggest reason behind not adding this line of code in `count.py` is because that function also handles labels for multi-label classification which are formatted in nested list format. However, `CleanLearning.find_label_issues` currently only supports multi-class classification hence we do have more liberty to reformat these labels. Because we don't have the multi-label case to worry about, we can probably now move this `labels = labels_to_array(labels)` line before the if-else statements (perhaps as high as L770) to ensure it gets called in both cases, and also before the function calls that expect labels to be formatted properly.

tataganesh: Apologies for the confusion regarding the unit-test! I should have been clearer about it.
Thank you for the detailed explanation regarding `labels = labels_to_array(labels)`, I will make the change in `CleanLearning.find_label_issues`.

tests/datalab/issue_manager/test_label.py
```diff
@@ -103,3 +103,10 @@ def test_get_summary_parameters(self, issue_manager, monkeypatch):
         summary_parameters.pop("pred_probs")
         expected_parameters.pop("pred_probs")
         assert summary_parameters == expected_parameters
+
+    def test_find_issues_low_memory(self, pred_probs, lab):
+        clean_learning_kwargs = {"low_memory": True}
+        issue_manager_lm = LabelIssueManager(
+            datalab=lab, clean_learning_kwargs=clean_learning_kwargs
+        )
+        issue_manager_lm.find_issues(pred_probs=pred_probs)
```
jwmueller, line 112: IMO this block should instead be added at the end of an existing unit test that does the find_issues without low_memory, and then you should finally check the jaccard similarity between the 2 runs.

tests/datalab/test_datalab.py
```diff
@@ -237,6 +237,10 @@ def test_find_issues_without_values_in_issue_types_raises_warning(self, lab, pre
             in warning_message
         )
 
+    def test_find_issues_with_low_memory(self, lab, pred_probs):
+        issue_types = {"label": {"clean_learning_kwargs": {"low_memory": True}}}
+        lab.find_issues(pred_probs=pred_probs, issue_types=issue_types)
```
jwmueller, line 242: IMO this block should instead be added at the end of an existing unit test that does the find_issues without low_memory, and then you should finally check the jaccard similarity between the 2 runs.

cleanlab/classification.py
```diff
@@ -782,123 +791,160 @@ def find_label_issues(
...
+            if self.verbose:
+                print("Using predicted probabilities to identify label issues ...")
+
+            arg_values = {
+                "thresholds": thresholds,
+                "noise_matrix": noise_matrix,
+                "inverst_noise_matrix": inverse_noise_matrix,
+                "find_label_issues_kwargs": self.find_label_issues_kwargs,
+            }
+            for arg_name, arg_val in arg_values.items():
```
huiwengoh, line 822: This check for `self.find_label_issues_kwargs` is invalid because by default `self.find_label_issues_kwargs = {}` and not `None`, so this warning will always trigger.

tataganesh: That makes sense. So should I use a separate if-else statement just for `find_label_issues_kwargs`?

huiwengoh: Yeah you could, or you could also isolate `self.find_label_issues_kwargs` from that loop - whichever is easier, does not really matter too much

huiwengoh (approved):
PR looks good to me! Thanks @tataganesh for the great contribution!

(note: the line failing on codecov is a print statement, we are good to ignore it)

Issue #765, opened 2023-07-06
[enhancement] DataLab for regression


mglowacki100:
Initiated by discussion https://github.com/cleanlab/cleanlab/discussions/137 @jwmueller 
I think it'd be nice to have `DataLab` functionality also for regression task when it will be oficially released.
I've looked at codebase and as far I see, `DataLab` is closely coupled with classification (e.g. method signature `find_issues(features=feature_embeddings, pred_probs=pred_probs)`, see `pred_probs`), so at the moment I'm not sure if this class could be easily and properly extended.
I've also noticed that besides `regression` task that is a little odd-ball, there are `object detection`, `segmentation`, `token_classification` and I don't know if `DataLab` supports them already? If no, maybe a better approach is to provide separate `DataLabs` for every task (with reusage of factory registration mechanism)?
Please let me know, what you think about it?

jwmueller:
Thanks for starting off the discussion!  CCing @elisno whose thoughts I'm curious to hear.

elisno:
Hi @mglowacki100, thank you for your suggestion. This is a crucial topic that we want to address.

I agree that `Datalab`'s method signatures are currently more oriented towards classification tasks. As we continue to add features that cater to various tasks, it's essential to consider how `Datalab` can be adapted to support these.

Starting with regression tasks is a good idea, but we should also approach this to handle future tasks.

However, instead of creating separate `Datalab` classes for each task type, we should introduce an argument to the `Datalab` constructor. This way, most of the work can be managed by a new issue mnager, keeping things clean and efficient.


---

To set the wheels in motion, I've opened [this issue](https://github.com/cleanlab/cleanlab/issues/774). It outlines the main steps required to kick-start regression support in `Datalab`.

How would you feel about contributing to this next step, @mglowacki100?

mglowacki100:
@elisno thanks for action plan! I'd work on this in next couple of days and let you know if I encounter any blockers or difficulties.
Btw. have you consider moving `classification.py` into new folder `classification` and renaming it into `learn.py` to make it more consistent with regression and other tasks? Or is it make on purpose and other tasks use "classification", as I didn't find for them `CleanLearnig` class?

elisno:
>Btw. have you consider moving `classification.py` into new folder `classification` and renaming it into `learn.py` to make it more consistent with regression and other tasks? Or is it make on purpose and other tasks use "classification", as I didn't find for them `CleanLearnig` class?

[...]

Issue #774, opened 2023-07-14
Supporting task-specific issues in Datalab


elisno:
While cleanlab has added support for various types of ML tasks, The only task-specific check supported by `Datalab` is label error detection for classification.

## Proposed Changes

The most straightforward approach to supporting new tasks, such as regression, is to write a new issue manager.

We propose several changes to extend `Datalab`’s functionality for different tasks (focusing on supporting classification and regression for now).

- [ ]  Define a `RegressionLabelIssueManager` for checking for label issues in regression tasks.
    - Add this into a separate `label.py` module in a new directory, like shown below.
        
        ```diff
        cleanlab/cleanlab/datalab/issue_manager/
        + ├── regression
        + │   └── label.py
        ├── __init__.py
        ├── duplicate.py
        ├── imbalance.py
        ├── issue_manager.py
        ├── label.py
        ├── noniid.py
        └── outlier.py
        ```
        
    - For simplicity, the issue manager and its `find_issues` method should take in whatever necessary arguments. Don’t worry about supporting different `model`s for now.
        - Take a close look at `regression.CleanLearning.find_label_issues`.
- [ ]  Add an argument to the Datlab constructor for specifying a task
    
    ```bash
    # Add argument and store as an instance variable
    Datalab.__init__(..., ml_task: str = "classification")
    ```
    
- [ ]  Make` _IssueManagerFactory` methods accept an argument for tasks
- [ ]  Update classes that use `_IssueManagerFactory` to pass task-argument from `Datalab` to the factory methods.

mglowacki100:
Hi @elisno 

To be on the same page regarding `Datalab` class modification, do you mean something like this:
```
class Datalab:
    def __init__(self, option):
        self.task = option

    def find_issues(self, *args, **kwargs):
        if self.task == 'classification':
            self._find_issues_classification(*args, **kwargs)
        elif self.task == 'regression':
            self._find_issues_regression(*args, **kwargs)
        else:
            print("Invalid option!")

    def _find_issues_classification(self, arg1=None, arg2=None):
        ...

    def _find_issues_regression(self, arg3=None):
        ...
```

elisno:
Hey @mglowacki100.

Thank you for your thoughtful input. I appreciate the effort you've put into solving this issue.

I've considered your approach of having separate task-specific registries for classification and regression tasks, and while this could work, I believe it may introduce some complexity, especially when we need to add more tasks in the future.

[...]

PR #796, opened 2023-08-07
datalab_regression

mglowacki100:
Based on discussion https://github.com/cleanlab/cleanlab/issues/774
Preleminary work to ensure that default task  (classification) will work smoothly. Still to do: `label.py` for regression, test for Datalab regression, docstring updates.

## Usage example

```python
from cleanlab import Datalab
import numpy as np

# Generate some toy-examples
N = 200
X = np.random.rand(N, 3)
y = 1 + 2 * X[:,0] + 3 * X[:,1] + 4 * X[:,2] + 0.05*np.random.randn(N)

# Run defaults
lab = Datalab(data={"y": y}, label_name='y', task="regression")
lab.find_issues(features=X)

# Run with any scikit-learn compatible model
from sklearn.neighbors import KNeighborsRegressor
lab = Datalab(data={"y": y}, label_name='y', task="regression")
model = KNeighborsRegressor(n_neighbors=3)
lab.find_issues(features=X, issue_types={"label": {"clean_learning_kwargs": {"model": model}}})
```


cleanlab/datalab/internal/issue_manager_factory.py
```diff
@@ -79,23 +88,40 @@ class _IssueManagerFactory:
...
+            if task not in TASK_SPECIFIC_REGISTRY:
+                raise ValueError(
+                    f"Invalid task type: {task}, must be in {list(TASK_SPECIFIC_REGISTRY.keys())}"
+                )
+
+            if issue_type not in TASK_SPECIFIC_REGISTRY[task] and issue_type not in REGISTRY:
+                raise ValueError(f"Invalid issue type: {issue_type} for task {task}")
+
+            if issue_type in TASK_SPECIFIC_REGISTRY[task]:
+                REGISTRY[issue_type] = TASK_SPECIFIC_REGISTRY[task][issue_type]
```
elisno, line 112:
```suggestion
        if task is None and issue_type not in REGISTRY:
            raise ValueError(f"Invalid issue type: {issue_type}")
        if task not in TASK_SPECIFIC_REGISTRY:
            raise ValueError(
                f"Invalid task type: {task}, must be in {list(TASK_SPECIFIC_REGISTRY.keys())}"
            )
        if issue_type not in TASK_SPECIFIC_REGISTRY[task] and issue_type not in REGISTRY:
            raise ValueError(f"Invalid issue type: {issue_type} for task {task}")
```


You can remove most of the nesting this way. If you feel these are too many if-statements, consider extracting all of them into an internal method, like

```python
@classmethod
def _validate_arguments(cls, issue_type: str, task: str) -> None:
    ...
```

Also, we shouldn't mutate the `REGISTRY` with `from_str`. We'd just return the `TASK_SPECIFIC_REGISTRY[task][issue_type]` instead.

cleanlab/datalab/internal/issue_manager_factory.py
```diff
@@ -79,23 +88,40 @@ class _IssueManagerFactory:
     """Factory class for constructing concrete issue managers."""
 
     @classmethod
-    def from_str(cls, issue_type: str) -> Type[IssueManager]:
+    def from_str(cls, issue_type: str, task: str = "classification") -> Type[IssueManager]:
```
elisno, line 91: The default `task = "classification"` should be set at one place in the code base, in the constructor of Datalab:
https://github.com/cleanlab/cleanlab/blob/c5af24340d6cdccc5caf1d76b11219016482d900/cleanlab/datalab/datalab.py#L103


```suggestion
    def from_str(cls, issue_type: str, task: str) -> Type[IssueManager]:
```

cleanlab/datalab/internal/issue_manager_factory.py
```diff
@@ -79,23 +88,40 @@ class _IssueManagerFactory:
...
+            if issue_type in TASK_SPECIFIC_REGISTRY[task]:
+                REGISTRY[issue_type] = TASK_SPECIFIC_REGISTRY[task][issue_type]
+
         return REGISTRY[issue_type]
 
     @classmethod
-    def from_list(cls, issue_types: List[str]) -> List[Type[IssueManager]]:
+    def from_list(
+        cls, issue_types: List[str], task: str = "classification"
+    ) -> List[Type[IssueManager]]:
```
elisno, line 119:
```suggestion
    def from_list(cls, issue_types: List[str], task: str) -> List[Type[IssueManager]]:
```

cleanlab/datalab/internal/issue_manager_factory.py
```diff
@@ -79,23 +88,40 @@ class _IssueManagerFactory:
...
         """Constructs a list of concrete issue manager classes from a list of strings."""
-        return [cls.from_str(issue_type) for issue_type in issue_types]
+        return [cls.from_str(issue_type, task) for issue_type in issue_types]
 
 
-def register(cls: Type[IssueManager]) -> Type[IssueManager]:
+def register(cls: Type[IssueManager], task: str = "classification") -> Type[IssueManager]:
```
elisno, line 124:
```suggestion
def register(cls: Type[IssueManager], task: str) -> Type[IssueManager]:
```

cleanlab/datalab/internal/issue_manager_factory.py
```diff
@@ -143,14 +169,30 @@ def find_issues(self, **kwargs):
...
+        if task not in TASK_SPECIFIC_REGISTRY:
+            raise ValueError(
+                f"Invalid task type: {task}, must be in {list(TASK_SPECIFIC_REGISTRY.keys())}"
+            )
+
+        if name in TASK_SPECIFIC_REGISTRY[task]:
+            print(
+                f"Warning: Overwriting existing issue manager {name} with {cls} for task {task}."
+                "This may cause unexpected behavior."
+            )
```
elisno, line 195: You can remove one level of nesting.

cleanlab/datalab/internal/issue_manager_factory.py
```diff
@@ -143,14 +169,30 @@ def find_issues(self, **kwargs):
...
+            print(
+                f"Warning: Overwriting existing issue manager {name} with {cls}. "
+                "This may cause unexpected behavior."
+            )
+
+    else:
+        if task not in TASK_SPECIFIC_REGISTRY:
+            raise ValueError(
+                f"Invalid task type: {task}, must be in {list(TASK_SPECIFIC_REGISTRY.keys())}"
+            )
```
elisno, line 189: Consider moving this to the top to raise an error early in the function.

cleanlab/datalab/internal/report.py
```diff
@@ -56,12 +56,14 @@ class Reporter:
     def __init__(
         self,
         data_issues: "DataIssues",
+        task: str = "classification",
```
elisno, line 59:
```suggestion
        task: str,
```

cleanlab/datalab/internal/issue_finder.py
```diff
@@ -70,8 +70,9 @@ class IssueFinder:
     `Datalab.find_issues` method which internally utilizes an IssueFinder instance.
     """
 
-    def __init__(self, datalab: "Datalab", verbosity=1):
+    def __init__(self, datalab: "Datalab", task: str = "classification", verbosity=1):
```
elisno, line 73:
```suggestion
    def __init__(self, datalab: "Datalab", task: str, verbosity=1):
```

elisno (commented):
Great start @mglowacki100!

Here are a few comments.

elisno:
You may want to merge master into your branch before moving on with development.

cleanlab/datalab/internal/issue_manager_factory.py
mglowacki100: Totally agree, refactoring was on todo list :)

elisno: Just a reminder to handle the case where we return the task-specific issue manager:

```python
TASK_SPECIFIC_REGISTRY[task][issue_type]
```

cleanlab/datalab/datalab.py
```diff
@@ -339,14 +342,25 @@ def report(
...
+        else:
+            reporter = Reporter(
+                data_issues=self.data_issues,
+                task=self.task,
+                verbosity=verbosity,
+                include_description=include_description,
+                show_summary_score=show_summary_score,
+                imagelab=self._imagelab,
+            )
+
```
elisno, line 363: It's okay for you to update the factory so that it won't have to pass the task to `ImagelabReporterAdapter`.

Let the factory be responsible for deciding if the Reporter needs to know about the task.

cleanlab/datalab/internal/issue_finder.py
mglowacki100: I think `IssueFinder` doesn't have to directly passed `task` as it could be extracted from `Datalab`. 
Btw. in few other places removal of default value of classification causes failure of test (missing value for positional argument `task`), I've updated those tests.  `ImageLab` was also updated.

cleanlab/datalab/datalab.py
mglowacki100: I've missed your comment and fixed it in this way:
```
class ImagelabReporterAdapter(Reporter):
    def __init__(
        self,
        data_issues: "DataIssues",
        imagelab: "Imagelab",
        task: str,
        verbosity: int = 1,
        include_description: bool = True,
        show_summary_score: bool = False,
    ):
        super().__init__(
            data_issues=data_issues,
            task=task,
            verbosity=verbosity,
            include_description=include_description,
            show_summary_score=show_summary_score,
        )
        self.imagelab = imagelab
```
but I'll reconsider your remark, especially that `Reporter` itself could be impacted by regression.

cleanlab/datalab/internal/issue_manager_factory.py
mglowacki100: Sure, will be covered be nesting refactoring.

cleanlab/datalab/internal/issue_manager_factory.py
mglowacki100: Below are proposed changes. I'm not sure if this is a coincidence but after this test on Mac started to fail due to unreliable test timings.
```
    if not issubclass(cls, IssueManager):
        raise ValueError(f"Class {cls} must be a subclass of IssueManager")

    name: str = str(cls.issue_name)

    if task is not None and task not in TASK_SPECIFIC_REGISTRY:
        raise ValueError(
            f"Invalid task type: {task}, must be in {list(TASK_SPECIFIC_REGISTRY.keys())}"
        )

    if task is not None and name in TASK_SPECIFIC_REGISTRY[task]:
        print(
            f"Warning: Overwriting existing issue manager {name} with {cls} for task {task}."
            "This may cause unexpected behavior."
        )

    if task is None and name in REGISTRY:
        print(
            f"Warning: Overwriting existing issue manager {name} with {cls}."
            "This may cause unexpected behavior."
        )
```

mglowacki100:
Hi @elisno 

Thanks for feedback :)
1. I've updated my branch to be on the sampe page with cleanlab/cleanlab:master
2. All tests are passing, but for some of them I've to specify `task`. Also I've added one test-case to increase test coverage for factory.
3. If everything is ok, I'll start to implement `label.py` for regression.
4. I think here docstring is wrong:
https://github.com/cleanlab/cleanlab/blob/master/cleanlab/datalab/internal/issue_manager/label.py
```
 @staticmethod
    def _process_find_label_issues_kwargs(kwargs: Dict[str, Any]) -> Dict[str, Any]:
        """Searches for keyword arguments that are meant for the
        CleanLearning.find_label_issues method call

        Examples
        --------
        >>> from cleanlab.datalab.internal.issue_manager.label import LabelIssueManager
        >>> LabelIssueManager._process_clean_learning_kwargs(thresholds=[0.1, 0.9])
        {'thresholds': [0.1, 0.9]}
```
should it be `LabelIssueManager._process_find_label_issues_kwargs( {'thresholds': [0.1, 0.9]})` ?

cleanlab/datalab/internal/issue_manager_factory.py
elisno: Yeah, the unreliable test timings are unrelated to this but recently fixed.

The Linux platform has always been faster, so that's expected behavior.

tests/datalab/test_issue_manager.py
```diff
@@ -68,3 +73,25 @@ def find_issues(self):
...
+    assert all(
+        [
+            text in sys.stdout.getvalue()
+            for text in [
+                "Warning: Overwriting existing issue manager label with ",
+                "NewerFoo",
+                " for task classification.",
+            ]
+        ]
+    ), "Should print a warning"
```
elisno, line 97: For consistency in the test, let the `NeweFoo.issue_name = "foo"`.

- Wouldn't this issue manager be found in the `TASK_SPECIFIC_REGISTRY`?
- The way the test is written, I understand that it wouldn't overwrite an existing issue manager . I recommend you use `@pytest.mark.parametrize` to test all possible values of `task`.

elisno (commented):
Everything is shaping up nicely, so more comments.

elisno:
> 3. If everything is ok, I'll start to implement `label.py` for regression.

That's great, feel free to start that when you're ready!
It's wonderful to see the iterative approach you've taken in this PR with frequent updates.

One small suggestion for making the review process smoother: 

- Consider providing more descriptive commit messages. Instead of "Update x.py", maybe something like:
  - "Adjusted function Y in x.py"
  - "Changed value Z in Y"
  - "Move block of code in A"
  - "rearrange imports"

  This will help give a clearer snapshot of the changes, especially when there are multiple commits for the same file.

- If you find yourself updating the same file multiple times in rapid succession, you might consider squashing those commits locally into one with a descriptive message before pushing.

Keep up the fantastic work!

elisno:
> 4. I think here docstring is wrong:
>    https://github.com/cleanlab/cleanlab/blob/master/cleanlab/datalab/internal/issue_manager/label.py
>         >>> LabelIssueManager._process_clean_learning_kwargs(thresholds=[0.1, 0.9])
>         {'thresholds': [0.1, 0.9]}
> ```
> 
> should it be `LabelIssueManager._process_find_label_issues_kwargs( {'thresholds': [0.1, 0.9]})` ? 

Nice catch! The example should use a dictionary as input, not a keyword argument.
I'll open a separate PR for these kinds of docstring revisions.

elisno:
>5. In `label.py` there is functionality regarding health summary from dataset, for multilabel classification there is a separate `dataset.py` in multilabel classification folder, should I create separate `dataset.py` for regression? Have you thought about some template for `dataset.py` (probably implementing datasets will be a first step for `Datalab` for other tasks) ?

That's ambitious, but I think we can manage without another `dataset.py` module for now.

tests/datalab/test_factory.py
```diff
@@ -13,15 +13,16 @@ def registry():
 def test_list_possible_issue_types(registry):
     issue_types = Datalab.list_possible_issue_types()
     assert isinstance(issue_types, list)
-    possible_issues = ["label", "outlier", "near_duplicate", "non_iid", "class_imbalance"]
+    possible_issues = ["outlier", "near_duplicate", "non_iid"]
```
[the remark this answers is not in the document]
mglowacki100: @elisno  I think it could be done in this way:
```
def list_possible_issue_types() -> List[str]:
        """... """
        return list(REGISTRY.keys()) + list(TASK_SPECIFIC_REGISTRY[self.datalab.task].keys())
```
or by mutating REGISTRY, but we want to avoid it.

Btw. I've started to test locally regression and noticed one small potential issue:
```
dl_regression = Datalab(data=df, label_name="target", task="regression")
datalab_regression.class_names
datalab_regression.get_info() 
```
This is problematic as for regression "number of classes" could be equal to number of datapoints, also for "extreme classification" - this could flood the screen ;) 
Sorry for spamming with commits, that should be squashed on my side.

jwmueller:
Closes https://github.com/cleanlab/cleanlab/issues/765

mglowacki100:
Hi @elisno , @jwmueller 

I had little time recently because of my regular job and family, but I had few hours today and I make it work end to end for regression and classification (basic test):
https://colab.research.google.com/drive/1xaOn6N_Jf4EBNZA-v6k7f8dCVkOB9a0b?usp=sharing
This is a draft and there is no code coverage, some tests are failing etc. but here are main changes/problems:
1. `DataIssues` class now has `self.task` attribute to workaround issue with regression and `get_info` method, namely `_label_map` for regression is "defined" but with it doesn't make much sense and with vectorize returns None
```
         for key in ["given_label", "predicted_label"]:
                labels = info.get(key, None)
                # ugly workaround:
                if labels is not None and self.task != "regression":
                    info[key] = np.vectorize(self._label_map.get)(labels)
                if labels is not None and self.task == "regression":
                    info[key] = labels
```
additionally in `DataLab` I pass task to `data_issues`.
There is no default for `task` in `DataIssues` so current tests are failing (missing) argument.
I don't like this workaround, but I didn't come up with alternative solution that is concise and elegant. Probably, assumptation that every issue has unique name could fix it or separate DataIssue class for regression. I've also tried to put `_label_map` into None for regression in init, but somehow it was restored later in the flow.
2. issue_finder.py: 
- list_possible_issue_types, list_default_issue_types methods are no longer static, because they depend on task. Regular methods have different signature than static ones, so I had to change them in few places.. 
- get_available_issue_types in the end is a little hacky:
```
if self.datalab.task == "regression":
       issue_types_copy = {**issue_types_copy, **{"label": {}}}

```
3. `label.py` for regression, I think this is a minimum viable implementation ;) To expose more I need to understand more about inner working of CleanLearning, epistemic and aleatoric uncertainty etc.
4. I'm not sure why tests for ImageLab fails but I suppose it is cause by change in DataIssues. I'll investigate it later, if we agree that "empty" task for ImageLab to have the same interface is a good solution.
5. Git, I think problem is with my Github Desktop configuration, when I push to my repo it also goes to cleanlab repo. I'll stick to command line git. Btw. is somewhere detailed instruction like this: https://xgboost.readthedocs.io/en/stable/contrib/git_guide.html?

What do you think about it?

elisno:
> 1. DataIssues class now has self.task attribute to workaround issue with regression and get_info method, namely _label_map for regression is "defined" but with it doesn't make much sense and with vectorize returns None
> ```
>          for key in ["given_label", "predicted_label"]:
>                 labels = info.get(key, None)
>                 # ugly workaround:
>                 if labels is not None and self.task != "regression":
>                     info[key] = np.vectorize(self._label_map.get)(labels)
>                 if labels is not None and self.task == "regression":
>                     info[key] = labels
> ```
> additionally in DataLab I pass task to data_issues.
> There is no default for task in DataIssues so current tests are failing (missing) argument.
> I don't like this workaround, but I didn't come up with alternative solution that is concise and elegant. Probably, assumptation that every issue has unique name could fix it or separate DataIssue class for regression. I've also tried to put _label_map into None for regression in init, but somehow it was restored later in the flow.

Hey @mglowacki100 

First and foremost, thank you for the dedication and time you've invested in this feature. I genuinely appreciate your commitment to enhancing `Datalab`.

Having gone through your proposal regarding `DataIssues`, I agree that the handling of `get_info` for various tasks is necessary.

Here's a suggested approach that I believe might streamline our current architecture

1. Leverage the strategy pattern for the `DataIssues.get_info` method:

    A scalable way to manage specific tasks without burdening the `DataIssues` class that's currently managing "issues", "info", and "statistics", is to inject a strategy (class) that handles task-specific behavior for each of the attributes. Right now, we're just worried about `get_info` depending on a label_map that only makes sense for classification. But the first part of the method is more generic that should work for all kinds of tasks. Simply returning the info dictionary may be enough.
    
    ```python
    class _InfoStrategy:

        def _get_info_for_task(self, data: Data, info: Dict[str, Dict[str, Any]], issue_name: Optional[str] = None) -> Dict[str, Any]:
            """Get the info for the issue_name key, handling task-specific logic."""
            return info.copy() # Maybe unnecessary to copy.

        def get_info(self, data: Data, info: Dict[str, Dict[str, Any]], issue_name: Optional[str] = None) -> Dict[str, Any]:
            info = info.get(issue_name, None) if issue_name else self.info
            if info is None:
                raise ValueError(
                    f"issue_name {issue_name} not found in self.info. These have not been computed yet."
                )
            assert info is not None
            
            # Handle task-specific logic.
            info = self._get_info_for_task(data=data, info=info, issue_name=issue_name)
            return info

    class _ClassificationInfoStrategy(_InfoStrategy):
    
        def _get_info_for_task(self, data: Data, info: Dict[str, Dict[str, Any]], issue_name: Optional[str] = None) -> Dict[str, Any]:
            info = info.copy()
            label_map = data.labels.label_map
            if issue_name == "label":
                if label_map is None:
                    raise ValueError(
                        "The label map is not available. "
                        "Most likely, no label column was provided when creating the Data object."
                    )
                # Labels that are stored as integers may need to be converted to strings.
                for key in ["given_label", "predicted_label"]:
                    labels = info.get(key, None)
                    if labels is not None:
                        info[key] = np.vectorize(label_map.get)(labels)

                info["class_names"] = info["statistics"]["class_names"]
            return info

    class DataIssues:
    
        def __init__(self, data: Data, strategy: _InfoStrategy) -> None:
            # ... same initialization, drop the label_map attribute...
            self._data = data
            self._strategy = strategy

        def get_info(self, issue_name: Optional[str] = None) -> Dict[str, Any]:
            return self._strategy.get_info(data=self._data,info=self.info, issue_name=issue_name)
    ```

   I don't think it's necessary to worry about strategies for `issues` and `statistics` just yet, so something like this should be enough. You could define a `_RegressionInfoStrategy(_InfoStrategy)` class that doesn't have to use a label_map. It's totally up to that class how it will handle the case `issue_name == "label"`.
   Of course, coming up with more descriptive argument names is very important. This is just to give an idea of how to extract task-specific behavior out of classes like `DataIssues`.

2. Switch from factory to builder pattern for `DataIssues` creation within `Datalab`:

    Considering the factors involved in constructing the `Datalab.data_issues` attribute, (i.e. `imagelab` and `task`), a builder pattern should make this process more controllable and intuitive.
    
    It would look something like this:
    
    ```python
    class _DataIssuesBuilder:
    
        def __init__(self, data: Data):
            self.data = data
            self.imagelab = None
            self.task = None

        def set_imagelab(self, imagelab):
            self.imagelab = imagelab
            return self  # Returning self for chaining

        def set_task(self, task: str):
            self.task = task
            return self  # Returning self for chaining

        def build(self) -> DataIssues:
            DataIssuesClass = self._data_issues_factory()

            # Create a _InfoStrategy based on self.task, maybe with another factory
            # Inject the strategy into a DataIssuesClass constructor and return it    
            return DataIssuesClass(...)

        def _data_issues_factory(self) -> Type[DataIssues]:
            if self.imagelab:
                return ImagelabDataIssuesAdapter
            else:
                return DataIssues
    ```

    On the `Datalab`-side, we have to make a builder object and specify the necessary steps before building the DataIssues object:
    
    
    ```python
    class Datalab:

        def __init__(...):
            # ... existing code ...

            builder = _DataIssuesBuilder(data=self._data)
            self.data_issues = (
                builder
                .set_imagelab(self._imagelab)
                # Task can be set here too if required in the future
                .build()
            )
    ```

mglowacki100:
Hi @elisno 
Thanks for kind feedback and change proposals :) I'll try to incorporate it in available time, but feel free to incorporate parts of this PR if this speeds up developments of Datalabs for other tasks by @jwmueller.
Btw. I've a question regarding regression for CleanLearn itself, now `find_label_issues` returns dataframe with columns like: is_label_issue, label_quality, given_label, predicted_label, so information what is a source (aleatoric, uncertainty or both) of the problem is not provided. I think, it'd be great to have that information, to know if issue is fixable (more data in that "region") or to know if we try to predict a random thing. I've looked at classification `find_label_issues` and it also could return sample_weight, so this dataframe format doesn't have to be completly unified. What are your thoughts on it?

jwmueller:
> I've a question regarding regression for CleanLearn itself, now find_label_issues returns dataframe with columns like: is_label_issue, label_quality, given_label, predicted_label, so information what is a source (aleatoric, uncertainty or both) of the problem is not provided. I think, it'd be great to have that information, to know if issue is fixable (more data in that "region") or to know if we try to predict a random thing

Yes that sounds like a nice idea. You can store this information in the `datalab.info` object which stores misc extra information computed anyway during the issue detection of each issue-type.

mglowacki100:
Hi @elisno , @jwmueller 
I've implemented also builder pattern. For `DataIssues` I've to add 
```
    def collect_issues_from_imagelab(self, imagelab: "Imagelab", issue_types: List[str]) -> None:
        pass  # pragma: no cover
```
to overcame type checker complaints. Other way is to unify it (name and signature) it with method `collect_issues_from_issue_manager`.
Here is basic test of branch https://colab.research.google.com/drive/1xaOn6N_Jf4EBNZA-v6k7f8dCVkOB9a0b?usp=sharing
All tests are passing (except macos - I think this is an infrastracture issue) and there is full code coverage, but I'm new to `pytest` (I use doctest ;) ) so probably a lot of things could be done better.
What do you think about it?

cleanlab/datalab/internal/issue_manager/regression/label.py
```diff
@@ -43,16 +34,98 @@ class RegressionLabelIssueManager(IssueManager):
...
+        ]
+        return {k: v for k, v in kwargs.items() if k in accepted_kwargs and v is not None}
+
+    def find_issues(
+        self,
+        **kwargs,
+    ) -> None:
+        """Find label issues in the datalab."""
+
+        RegressionLabelIssueManager.issue_score_key = "label_score"
```
elisno, line 91: Doesn't the [metaclass `IssueManagerMeta`](https://github.com/cleanlab/cleanlab/blob/14d1a86d4375beb17168eafdea88c84fd73c9a65/cleanlab/datalab/internal/issue_manager/issue_manager.py#L78) already handle this properly? Can you add a test case to confirm this?

It should work without this line added explicitly.

tests/datalab/test_factory.py
```diff
@@ -23,7 +24,7 @@ class TestIssueManager(IssueManager):
 
     TestIssueManager = register(TestIssueManager, task=None)
 
-    issue_types = Datalab.list_possible_issue_types()
+    issue_types = lab.list_possible_issue_types()
```
elisno, line 27: Reminder to self:

Revisit this line after the issue type listing methods have been refactored.

elisno (commented):
Thanks for your work on this, @mglowacki100!

A few thoughts based on [your latest update](https://github.com/cleanlab/cleanlab/pull/796#issuecomment-1706060121).

- I had a feeling the `collect_issues_from_issue_manager` vs `collect_issues_from_imagelab` resolution would have its rough edges. Because the methods are fundamentally differernt, I’m considering using more composition. If your quick-fix works, we can address that later.
- `[datalab.py](http://datalab.py)` should only have public-facing functionality. Can you move the `_DataIssuesBuilder` into the `helper_factory.py` for now? We can worry about module naming later.
- After reviewing the structure and responsibilities of the **`IssueFinder`** class, I believe there is an opportunity for refactoring to make our codebase more maintainable and intuitive.

The **`IssueFinder`** class, as it stands, seems to have multiple responsibilities: not only does it handle the actual task of finding issues, but it also determines what types of issues can be found, by referencing **`REGISTRY`** and **`TASK_SPECIFIC_REGISTRY`**. The latter responsibility might fit more naturally in the module where the registries are defined.

## Proposed Changes

1. **Moving Listing Methods**: I suggest we move `list_possible_issue_types` and `list_default_issue_types` from the `IssueFinder`  class to `cleanlab/datalab/internal/issue_manager_factory.py` where the registries are defined. Something like:
        
```python
    def list_possible_issue_types(task: str) -> List[str]:
        ...
    def list_default_issue_types(task: str) -> List[str]:
        ...
```

2. **Refactoring** `IssueFinder`**:** The `IssueFinder` class would retain its `find_issues` method (and the remaining helper methods required). When it needs to call the listing methods, it can simply reference the aforementioned utility functions by passing the necessary `task`.
    For now, I think it’s fine to set a `task` instance variable so we don’t need to inject `self.datalab.task` everywhere. In the long run, it would be nice if we can minimize the bidirectional coupling between `Datalab` and the internal classes like `IssueFinder`.

Looking forward to the updates. Let me know if you need any clarifications!

cleanlab/datalab/datalab.py
```diff
@@ -50,9 +51,42 @@
...
+    def set_task(self, task):
+        self.task = task
+        return self
+
+    def build(self) -> DataIssues:
+        data_issues_class = self._data_issues_factory()
+        strategy = self._create_info_strategy()
+        return data_issues_class(self.data, strategy)
+
+    def _data_issues_factory(self) -> Union[Type[DataIssues], Type[ImagelabDataIssuesAdapter]]:
```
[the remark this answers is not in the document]
mglowacki100: Yeah, it was my attempt to suppress type checker error about missing `collect_issues_from_imagelab` in `DataIssues`. Fixed in https://github.com/cleanlab/cleanlab/pull/796/commits/0d389173a354d8b92255921135121083b2c6866e

cleanlab/datalab/datalab.py
```diff
@@ -117,7 +151,12 @@ def __init__(
         self.cleanlab_version = cleanlab.version.__version__
         self.verbosity = verbosity
         self._imagelab = create_imagelab(dataset=self.data, image_key=image_key)
-        self.data_issues = data_issues_factory(self._imagelab)(self._data)
+
+        # Create the builder for DataIssues
+        builder = _DataIssuesBuilder(self._data)
+        builder.set_imagelab(self._imagelab)
+        builder.set_task(task)
```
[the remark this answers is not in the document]
mglowacki100: OK. If understand correclty, do you want to in this style:
```
builder = _DataIssuesBuilder(self._data)
builder.set_imagelab(self._imagelab).set_task(task)
self.data_issues = builder.build()
```
ore even more compact:
```
self.data_issues = _DataIssuesBuilder(self._data).set_imagelab(self._imagelab).set_task(task).build()
```

cleanlab/datalab/internal/issue_finder.py
```diff
@@ -358,4 +372,7 @@ def get_available_issue_types(self, **kwargs):
                 )
                 issue_types_copy.pop("outlier")
 
+        if self.datalab.task == "regression":
+            issue_types_copy = {**issue_types_copy, **{"label": {}}}
+
```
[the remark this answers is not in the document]
mglowacki100: The problem is in `find_issues` method in `IssueFinder`:
```
    def find_issues(
        self,
        *,
        pred_probs: Optional[np.ndarray] = None,
        features: Optional[npt.NDArray] = None,
        knn_graph: Optional[csr_matrix] = None,
        issue_types: Optional[Dict[str, Any]] = None,
    ) -> None:
        issue_types_copy = self.get_available_issue_types(
            pred_probs=pred_probs,
            features=features,
            knn_graph=knn_graph,
            issue_types=issue_types,
        )
        if not issue_types_copy:
            return None
```
For regression it could be called without arguments `find_issues()`, wihtout this "if" in `get_available_issue_types` result will be `None`. `issue_types_copy = {**issue_types_copy, **{"label": {}}}` ensures that we keep e.g. issue types from  `knn_graph` if provided in issues_types_copy  and adds `label` - for regression we can't dectect it from `pred_probs`. I hope this makes sense, but I could miss something obvious.

cleanlab/datalab/datalab.py
```diff
@@ -50,9 +51,42 @@
...
+        if self.imagelab:
+            return ImagelabDataIssuesAdapter
+        else:
+            return DataIssues
+
+    def _create_info_strategy(self):
+        if self.task == "regression":
+            return _RegressionInfoStrategy()
+        else:
+            return _ClassificationInfoStrategy()
```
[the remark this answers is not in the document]
mglowacki100: Good catch, fixed in https://github.com/cleanlab/cleanlab/pull/796/commits/76e546bd54223682e4e2742f9aae35db4a189496

cleanlab/datalab/internal/issue_manager/regression/label.py
mglowacki100: Great catch! Fixed in https://github.com/cleanlab/cleanlab/pull/796/commits/49635a785fd0e301fe7b6869f30f708b6b2514ad

cleanlab/datalab/internal/helper_factory.py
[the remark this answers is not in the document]
mglowacki100: Fixed in https://github.com/cleanlab/cleanlab/pull/796/commits/20c455b60262adf2e560b7da08a5fd5ce71fb9a3

mglowacki100:
Hey @elisno,
I've made changes that you suggested, I think everything is commited except conversation about method chaining in builder. Please resolve conversations if issue is resolved.

Note that, I've removed  test from test_issue_finder.py:
```
       @pytest.mark.parametrize(
        "defaults_dict",
        [
            {"issue_type_1": {"arg_1": "default_value_1"}},
        ],
    )
    @pytest.mark.parametrize(
        "issue_types",
        [{"issue_type_1": {"arg_1": "value_1", "arg_2": "value_2"}}, {"issue_type_1": {}}],
    )
    def test_set_issue_types(self, issue_finder, issue_types, defaults_dict, monkeypatch):
        """Test that the issue_types dict is set correctly."""
        with monkeypatch.context() as m:
            # Mock the validation method to do nothing
            m.setattr(issue_finder, "_validate_issue_types_dict", lambda x, y: None)
            m.setattr(issue_finder, "list_possible_issue_types", lambda *_: ["issue_type_1"])
            issue_types_copy = issue_finder._set_issue_types(issue_types, defaults_dict)

            # For each argument in issue_types missing from defaults_dict, it should be added to the defaults dict
            for issue_type, args in issue_types.items():
                missing_args = set(args.keys()) - set(defaults_dict[issue_type].keys())
                for arg in missing_args:
                    assert issue_types_copy[issue_type][arg] == args[arg]
 
``` 
also in test_datalab.py, I've changed:
@pytest.fixture(scope="function")
def list_possible_issue_types(monkeypatch, request):
    monkeypatch.setattr(IssueFinder, "list_possible_issue_types", lambda *_: request.param)
into:
```
@pytest.fixture(scope="function")
def list_possible_issue_types(monkeypatch, request):
    return lambda *_: request.param
```
probably, there is more obvious way to do it.

Thanks for review and have a good weekend :)

mglowacki100:
Hi @elisno  @jwmueller ,

I hope you're doing well. I wanted to touch base regarding the pull request I submitted recently. I understand everyone is busy, but I'd appreciate it if you could spare some time to review it when you have a chance. Your feedback is always valuable.

On another note, congratulations to the team on the successful release! 

I also noticed that there are some type check test failures after merge to new release, but they seem unrelated to the changes I made in my pull request. I'll be happy to assist in debugging and resolving these issues if needed. Please let me know if you would like me to take a closer look or if there's anything specific I should focus on.

Thank you for your time, and I look forward to your feedback on the pull request.

elisno:
Hi @mglowacki100,

I trust you're well. I wanted to reach out and let you know that I've seen your comments. I must admit, that your first message slipped by me, and I apologize for not responding sooner.

I’m currently working through a busy schedule, but I have allocated time to address your comments and concerns in the next couple of days.

I appreciate your understanding and value your insight in this PR.
At a quick glance, you're right that the type checks are not related to this PR.

cleanlab/datalab/internal/issue_finder.py
```diff
@@ -72,6 +76,7 @@ class IssueFinder:
 
     def __init__(self, datalab: "Datalab", verbosity=1):
         self.datalab = datalab
+        self.task = self.datalab.task
```
elisno, line 79: Can we just inject the task into this constructor, instead of increasing the coupling between Datalab and IssueFinder?

```suggestion
    def __init__(self, datalab: "Datalab", task:str, verbosity=1):
        self.datalab = datalab
        self.task = task
```

Update docstring accordingly!

cleanlab/datalab/internal/issue_finder.py
```diff
@@ -356,4 +342,7 @@ def get_available_issue_types(self, **kwargs):
                 )
                 issue_types_copy.pop("outlier")
 
+        if self.task == "regression":
+            issue_types_copy = {**issue_types_copy, **{"label": {}}}
```
elisno, line 346: Maybe add a comment above here why `"label"` is being "reset".

Can you confirm that this line won't run if `"label"` is not in the issue_types dictionary?

elisno (commented):
Here are a couple of suggested changes for now.

cleanlab/datalab/internal/data_issues.py
```diff
@@ -40,6 +41,73 @@
...
 
 
+class _InfoStrategy(ABC):
+    @staticmethod
+    @abstractmethod
+    def get_info(
+        data: Data,
+        info: Dict[str, Dict[str, Any]],
+        issue_name: Optional[str] = None,
+    ) -> Dict[str, Any]:
```
elisno, line 51:
```suggestion
    def get_info(
        data: Data,
        info: Dict[str, Dict[str, Any]],
        issue_name: Optional[str] = None,
    ) -> Dict[str, Any]:
        """
        Get information about a data issue from an information dictionary.

        Parameters
        ----------
        info : dict
            A dictionary of information about data issues.
        issue_name : str or None, optional (default=None)
            The name of the issue to get information about. If `None`, this method returns `None`.

        Returns
        -------
        dict or None
            A copy of the information dictionary for the specified issue, or `None` if the issue
            name is `None`.

        Raises
        ------
        ValueError
            If the specified issue name is not found in the information dictionary.
        """
```

cleanlab/datalab/datalab.py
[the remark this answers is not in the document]
mglowacki100: I've implemented first variant

cleanlab/datalab/internal/issue_finder.py
mglowacki100: Hi @elisno,

I'm not entirely certain about this suggestion. For example, I'm unsure about how IssueFinder should behave if I initialize it with a Datalab that has the task set to classification and then provide the task "regression" in a separate argument as you suggest. Currently, issueFinder obtains task from datalab.

cleanlab/datalab/internal/issue_finder.py
mglowacki100: Detailed explanation is here: https://github.com/cleanlab/cleanlab/pull/796#discussion_r1316764067
I've added addition comment in code

cleanlab/datalab/datalab.py
```diff
@@ -499,10 +509,9 @@ def list_possible_issue_types() -> List[str]:
         --------
         :py:class:`REGISTRY <cleanlab.datalab.internal.issue_manager_factory.REGISTRY>` : All available issue types and their corresponding issue managers can be found here.
         """
-        return IssueFinder.list_possible_issue_types()
+        return list_possible_issue_types(task=self.task)
```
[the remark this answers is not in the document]
mglowacki100: Code updated.

cleanlab/datalab/datalab.py
```diff
@@ -514,7 +523,7 @@ def list_default_issue_types() -> List[str]:
         --------
         :py:class:`REGISTRY <cleanlab.datalab.internal.issue_manager_factory.REGISTRY>` : All available issue types and their corresponding issue managers can be found here.
         """
-        return IssueFinder.list_default_issue_types()
+        return list_default_issue_types(task=self.task)
```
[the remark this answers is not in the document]
mglowacki100: Hi @elisno 
Fixed in the same way as list_possible_issue_types (change to _list_possible_issue_types)

cleanlab/datalab/internal/data_issues.py
```diff
@@ -265,6 +304,9 @@ def collect_issues_from_issue_manager(self, issue_manager: IssueManager) -> None
         )
         self._update_issue_info(issue_manager.issue_name, issue_manager.info)
 
+    def collect_issues_from_imagelab(self, imagelab: "Imagelab", issue_types: List[str]) -> None:
+        pass  # pragma: no cover
+
```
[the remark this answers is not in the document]
mglowacki100: Hi @elisno 
Suprisingly, there is no complaint from type-checker now, so it looks like it was weird issue (maybe there was some changes in type-checker config/version in the meanwhile that caused it?)

cleanlab/datalab/internal/issue_finder.py
jwmueller: I don't think we should allow changing the task type in a constructed Datalab. It should be fixed forever IMO

elisno (approved):
Wonderful work @mglowacki100!

I've made some adjustments to the registry and related functionality, as well as handled labels are numerical values (for regression).

mglowacki100:
Hi @elisno 
Thank you for your kind words! I really appreciate your acknowledgment of the work. 
I look forward to continuing to contribute to cleanlab.
Best regards,
Marek

PR #821, opened 2023-08-16
Multiannotator module improvements + speedup

huiwengoh:
Optimization of the multiannotator module functions that will provide computational speedups.

At it’s best (ie. the case where the multiannotator labels dataframe has string indices for annotators), this PR provides up to ~50x speedup, perhaps even more as the datasets gets larger as pd.loc/iloc in the old implementation scales far worse.


cleanlab/multiannotator.py
```diff
@@ -743,13 +792,23 @@ def get_active_learning_scores_ensemble(
...
-        if isinstance(labels_multiannotator, np.ndarray):
-            labels_multiannotator = pd.DataFrame(labels_multiannotator)
+        if isinstance(labels_multiannotator, pd.DataFrame):
+            labels_multiannotator = (
+                labels_multiannotator.replace({pd.NA: np.NaN}).astype(float).to_numpy()
+            )
+        elif isinstance(labels_multiannotator, np.ndarray):
+            # make sure it is a column array
+            if labels_multiannotator.ndim == 1:
+                labels_multiannotator = labels_multiannotator.reshape(-1, 1)
```
jwmueller, line 802: should this line be unit-tested?

cleanlab/multiannotator.py
```diff
@@ -1689,10 +1823,17 @@ def _get_annotator_quality(
...
-            lambda s: np.mean(s[pd.notna(s)] == consensus_label_subset[pd.notna(s)]),
-            axis=0,
-        )
+        annotator_quality = np.zeros(labels_multiannotator_subset.shape[1])
+        for i in range(len(annotator_quality)):
+            labels = labels_multiannotator_subset[:, i]
+            labels_mask = ~np.isnan(labels)
+            # case where annotator does not annotate any examples with any other annotators
+            if np.sum(labels_mask) == 0:
+                annotator_quality[i] = np.NaN
```
jwmueller, line 1832: should this line be unit tested?

cleanlab/multiannotator.py
```diff
@@ -172,19 +172,30 @@ def get_label_quality_multiannotator(
 
     """
 
-    if isinstance(labels_multiannotator, np.ndarray):
-        labels_multiannotator = pd.DataFrame(labels_multiannotator)
+    if isinstance(labels_multiannotator, pd.DataFrame):
```
jwmueller, line 175: can you confirm none of the user-facing methods' APIs change in this PR?

cleanlab/multiannotator.py
huiwengoh: This is non-intended, have edited this line

cleanlab/multiannotator.py
huiwengoh: Yes, user facing API has not changed. All the multiannotator functions still allow `labels_multiannotator` to be either a np array or pd dataframe. 

The difference here is that we used to convert everything to a dataframe before performing internal operations, whereas now we convert everything to a np array internally since the optimized functions operate on np arrays.

jwmueller (approved):
50x speedup is amazing!

PR #822, opened 2023-08-16
Expose low memory option for token and multi-label classification

tataganesh:
https://github.com/cleanlab/cleanlab/issues/762 - find_label_issues in cleanlab token classification and  mutl-label classification.

## Summary of Additonal Changes
* `return_mask`  parameter added to `find_label_issues_batched` to ensure that its return type is consistent with `filter.find_label_issues`. 
* Use f-string for all print statements for the `display_issues` function in token classification.
* Modify print output of `display_issues` to show mislabelled token as well as its ID. 

##  Token Classification Results 
(using the Token Classification [tutorial](https://github.com/cleanlab/cleanlab/blob/master/docs/source/tutorials/token_classification.ipynb))

### Runtime Comparison
Original
```
2.17 s ± 76.2 ms per loop (mean ± std. dev. of 3 runs, 1 loop each)
```

With Low Memory
```
252 ms ± 1.41 ms per loop (mean ± std. dev. of 3 runs, 1 loop each)
```
### Compare Metrics

|                                            | Original  | Low Memory |
|--------------------------------------------|:---------:|:----------:|
|              Number of Issues              |    2254    |     2175    |
|               Issues Overlap               |     -     | 1991 (88.33%) |
|             Jaccard Similarity             |     -     |    0.82    |

##  Multi-Label Classification Results 
(using the Multi-Label Classification [tutorial](https://github.com/cleanlab/cleanlab/blob/master/docs/source/tutorials/multilabel_classification.ipynb))

### Runtime Comparison
Original
```
3.47 s ± 78.5 ms per loop (mean ± std. dev. of 3 runs, 1 loop each)
# (This seems to be higher than I expected for a small dataset)
```

With Low Memory
```
3.47 ms ± 903 µs per loop (mean ± std. dev. of 3 runs, 1 loop each)
```
### Compare Metrics

|                                            | Original  | Low Memory |
|--------------------------------------------|:---------:|:----------:|
|              Number of Issues              |    30    |     23    |
|               Issues Overlap               |     -     | 23 (76.67%) |
|             Jaccard Similarity             |     -     |    0.77    |

### `display_issues` changes for token classification

Before
```
Sentence 2907, token 0:
Given label: PER, predicted label according to provided pred_probs: O
----
Little change from today's weather expected.
```

After
```
Sentence index: 2907, Token index: 0
Token: Little
Given label: PER, predicted label according to provided pred_probs: O
----
Little change from today's weather expected.
```


cleanlab/multilabel_classification/filter.py
```diff
@@ -241,27 +265,39 @@ def find_multilabel_issues_per_class(
             raise ValueError("confident_joint should be of shape (num_classes, 2, 2)")
     for class_num, (label, pred_prob_for_class) in enumerate(zip(y_one.T, pred_probs.T)):
         pred_probs_binary = stack_complement(pred_prob_for_class)
```
huiwengoh, line 267: (can't comment on the exact line so will comment here)

I believe the `confident_joint` computation (starts a few line above) can also be skipped if `low_memory=True`

huiwengoh (commented):
Small comments, overall I think it looks pretty good (in general if you could double check that. 
Super great job incorporating feedback from your previous PR into this one as well!

cleanlab/multilabel_classification/filter.py
```diff
@@ -241,27 +265,39 @@ def find_multilabel_issues_per_class(
             raise ValueError("confident_joint should be of shape (num_classes, 2, 2)")
```
huiwengoh, line 265: (can't comment on the exact line so will comment here)

I believe the confident_joint computation (starts a few line above) can also be skipped if low_memory=True

cleanlab/multilabel_classification/filter.py
```diff
@@ -134,10 +134,8 @@ def find_label_issues(
             warnings.warn(f"`rank_by_kwargs` is not used when `low_memory=True`.")
         arg_values = {
             "filter_by": filter_by,
-            "frac_noise": frac_noise,
```
huiwengoh: I think we should still check if users pass in non-default args for this, my previous comment was just pointing out that the default arg is not `None`, so we need to check for it slightly differently

cleanlab/multilabel_classification/filter.py
```diff
@@ -134,10 +134,8 @@ def find_label_issues(
...
             "num_to_remove_per_class": num_to_remove_per_class,
             "confident_joint": confident_joint,
-            "min_examples_per_class": min_examples_per_class,
```
huiwengoh: I think we should still check if users pass in non-default args for this (see comment on `frac_noise` above)

cleanlab/multilabel_classification/filter.py
tataganesh: Got it. How would we check if the argument has a default or non-default value? The only way I can think of is

```
if frac_noise ! = 1.0:
    warnings.warn("..")
if min_examples_per_class != 1:
   warnings.warn("..")
```
This doesn't seem right though, because if the default values are changed, then they would need to be changed in these conditions as well.  But if this works, I can make the relevant changes.

tataganesh: Since cleanlab does use the `inspect` module at certain places, this might also work.
```python
import inspect
signature = inspect.signature(find_label_issues)
default_args =  {
    k: v.default
    for k, v in signature.parameters.items()
    if v.default is not inspect.Parameter.empty
}
arg_values = {
    "filter_by": filter_by,
    "num_to_remove_per_class": num_to_remove_per_class,
    "confident_joint": confident_joint,
    "n_jobs": n_jobs,
    "num_to_remove_per_class": num_to_remove_per_class,
    "frac_noise": frac_noise,
    "min_examples_per_class": min_examples_per_class
}
# If an argument does not have a default value or if 
# passed value differs from the default, raise warning, 
for arg_name, arg_val in arg_values.items():
    if arg_name not in default_args or arg_val != default_args[arg_name]:
        warnings.warn(f"`{arg_name}` is not used when `low_memory=True`.")
```
Adapted from this [SO answer](https://stackoverflow.com/a/12627202/8505509).

huiwengoh: I like the `inspect` approach! 

One suggestion - I believe you can simplify the last if statement to just this?

```
for arg_name, arg_val in arg_values.items():
    if arg_val != default_args[arg_name]:
        print(f"`{arg_name}` is not used when `low_memory=True`.")
```

I'm not entirely sure what `arg_name not in default_args` is checking for since I think arguments with `None` values will still be in the `default_args` dict. Please double check this tho and correct me if I'm wrong / am missing something.

tataganesh: You are right! I will simplify the condition, thanks!
As a side note, I observed that I made similar changes in other places where the `low_memory` option has been introduced. In the future, it might make sense to make a common function `check_args(arg_values, func)` that uses the signature of the function and specified arguments to prints warnings.

huiwengoh (approved):
Code additions LGTM - amazing job working on exposing the `low_memory` options throughout the code base! 

Thanks for another great PR @tataganesh 🙂

PR #825, opened 2023-08-22
Catch specific exception for knn in datalab issue managers

tataganesh:
Mimic exception handling performed in [outlier.py](https://github.com/cleanlab/cleanlab/blob/0adce8290fbe7527cc9c7d3652da6ed68cc5afc4/cleanlab/outlier.py#L434-L437) for duplicate and non-iid issue managers


elisno (approved):
Polished touches, @tataganesh!

PR #828, opened 2023-08-24
Fix Imbalance Score: Use rare class proportion instead of mean score

tataganesh:
Use proportion of rarest class for overall dataset score for the class imbalance issue type. Further, small changes have been to remove function call to `min()` and instead  only use `argmin` for finding the minimum value. Using the example mentioned in https://github.com/cleanlab/cleanlab/pull/758, the output of summary score is as follows:

**Before Fix**
```
Overall dataset quality in terms of this issue: 0.9756
```

**After Fix**
```
Overall dataset quality in terms of this issue: 0.0250
```


elisno (approved):
That's a bright idea @tataganesh - well done!

Issue #776, opened 2023-07-18
new Datalab issue type: underperforming group (data slice receiving bad predictions)


jwmueller:
Goal is to auto-detect a coherent cluster of “hard” examples (ie. data slice) where models predictions are poor. Cf:
https://dcai.csail.mit.edu/lectures/data-centric-evaluation/

This should be a [new Datalab issue type ](https://docs.cleanlab.ai/master/cleanlab/datalab/guide/custom_issue_manager.html) called something like: `underperforming_group`.

Detecting this issue type will require these inputs: `features`, `pred_probs`, `labels`.

Steps to detect this issue: 
1. Cluster the data
2. For each cluster `c`, compute the average loss of all examples belonging to that cluster: `L_c`. Loss is computed between `pred_probs` and `labels` and can be any loss-function, say [self_confidence](https://docs.cleanlab.ai/stable/cleanlab/rank.html#cleanlab.rank.get_self_confidence_for_each_label). Let `L_w` be the the worst cluster's loss: `L_w = min_c { L_c }`
3. Compute overall prediction loss across the full dataset too: `L`.
4. If `L_w / L < c`, then we say this dataset suffers from the `underperforming_group` issue. 0 < c < 0.5 here is a hyperparameter.
5. If the dataset suffers from this issue, then we set: 
- the overall dataset quality score for the `underperforming_group` issue to be = `L_w / L` (lower values = more severe issue).
- the number of examples suffering from this issue in the dataset = the size of the worst-performing cluster.
- the quality score for each example for the `underperforming_group` issue to be = `L_w / L` for examples that belong to the worst-performing cluster, otherwise = 1.0 for other examples.

If the dataset does not suffer from this issue, then we set:
- the overall dataset quality score for the `underperforming_group` issue to be = `L_w / L` (lower values = more severe issue).
- the number of examples suffering from this issue in the dataset = 0
- the quality score for each example for the `underperforming_group` issue to be = 1.0

6. Set `Datalab.info` to keep track of other useful information computed during these steps.

**Implementation Notes**: 
Datalab already computes a KNN graph of the examples based on provided `features`. 
We should just use these pre-computed distances as the sole input to clustering algorithm. 
Recommend to start with [DBSCAN](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.DBSCAN.html#sklearn.cluster.DBSCAN) to do the clustering, using `metric='precomputed'` and passing in the already-computed KNN graph as a sparse array.

[...]

PR #838, opened 2023-09-04
Underperforming Group Issue Type

tataganesh:
## Summary

 🎯 **Purpose**: Implementation of the Underperforming Group Issue Manager. 

📜 **Example Usage**:
```python
import numpy as np
from cleanlab import Datalab
from sklearn.model_selection import cross_val_predict

# Set seed for reproducibility
np.random.seed(0)

# Generate toy dataset
X1A, X1B, X2 = (
    np.random.normal(loc=2, scale=0.4, size=(100, 2)),
    np.random.normal(loc=0.3, scale=0.1, size=(50, 2)) + [2.1, 0.2],
    np.random.normal(loc=4.5, scale=0.5, size=(150, 2))
)
X = np.concatenate((X1A, X1B, X2), axis=0)

y1A, y1B, y2 = (
    np.full(100, 0),
    np.random.choice([0, 1], size=50, p=[0.25, 0.75]),
    np.full(150, 1)
)
y = np.concatenate((y1A, y1B, y2), axis=0)

data = {"X": X, "y": y}

# Train classifier and generate out-of-sample probabilities
model = LogisticRegression()
pred_probs = cross_val_predict(model, X, y, method="predict_proba")

# Use Datalab to extract Underperforming Group
lab = Datalab(data={"y": y}, label_name="y")
lab.find_issues(features=X, pred_probs=pred_probs, issue_types={"underperforming_group": {}})
lab.report()

```

<details><summary>See Report</summary>

```
Finding underperforming_group issues ...

Audit complete. 50 issues found in the dataset.

           issue_type  num_issues
underperforming_group          50

Dataset Information: num_examples: 300, num_classes: 2


--------------- underperforming_group issues ---------------

About this issue:
        An underperforming group refers to a collection of “hard” examples
    for which the model predictions are poor. The quality of predictions is
    computed using the :py:func:`get_self_confidence_for_each_label <cleanlab.rank.get_self_confidence_for_each_label>` function.
    

Number of examples with this issue: 50
Overall dataset quality in terms of this issue: 0.0978

     is_underperforming_group_issue  underperforming_group_score
149                            True                      0.09782
119                            True                      0.09782
120                            True                      0.09782
121                            True                      0.09782
122                            True                      0.09782
```

Cluster Plot
image

</details>

## Impact

🌐 Areas Affected: New Issue Manager added to Datalab. It will not be run by default. Updated documentation for issue types supported by Datalab. 

## Testing

 🔍 Testing Done: Unit-tests for issue manager and basic end-to-end testing with Datalab added.
🔗 Test Case Link: <To be Filled>

Partially addresses #776


cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,222 @@
...
+            cur_cluster_labels = labels[cluster_mask]
+            cur_cluster_pred_probs = pred_probs[cluster_mask]
+            cluster_loss = get_self_confidence_for_each_label(
+                cur_cluster_labels, cur_cluster_pred_probs
+            ).mean()
+            if cluster_loss < min_loss:
+                min_loss = cluster_loss
+                min_loss_clusterid = i
+        loss_ratio = min(min_loss / full_loss, 1.0)
+        print(min_loss, full_loss)
```
elisno, line 126: Reminder to remove this print statement later.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,222 @@
...
+        min_loss_clusterid = min_loss_clusterid if loss_ratio < self.threshold else n_clusters
```
elisno, line 127: I'm wondering how to make the `else n_clusters` bit work, so the chosen ID doesn't conflict with any existing cluster.

My first thought was to replace it with `max(unique_labels) + 1`, but I'm not sure.

What do you think @tataganesh?

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+        if prefer_new_graph:
+            statistics_dict["statistics"][graph_key] = knn_graph
+            if self.metric is not None:
+                statistics_dict["statistics"]["knn_metric"] = self.metric
+
+        return statistics_dict
+
+    def _get_cluster_statistics(self, n_clusters: int) -> Dict[str, Dict[str, Any]]:
+        cluster_stats = {"HDBSCAN": {"n_clusters": n_clusters}}
```
elisno, line 206: I'd like to suggest a slight modification in how we structure the clustering information. We can make the dictionary more intuitive and maintainable, especially if we decide to expand on it or integrate other clustering methods in the future.

```suggestion
        cluster_stats = {
            "clustering": {
                "algorithm": "HDBSCAN",  # or any other method we might use in the future
                "params": {
                    # Specific parameters related to the chosen clustering algorithm
                    "min_cluster_size": 5,
                    "min_samples": None,
                    "metric": "precomputed",
                    # ... any other parameters that may be useful ...
                },
                "stats": {
                    # Derived statistics or post-clustering results
                    "n_clusters": n_clusters,
                    "cluster_labels": labels_,
                    "silhouette_score": sklearn.metrics.silhouette_score(...)
                    # ... any other statistics/results ...
                }
            }
        }
```

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,222 @@
...
+
+from cleanlab.datalab.internal.issue_manager import IssueManager
+from cleanlab.rank import get_self_confidence_for_each_label
+
+if TYPE_CHECKING:  # pragma: no cover
+    import numpy.typing as npt
+    from cleanlab.datalab.datalab import Datalab
+
+
+class UnderperfGroupIssueManager(IssueManager):
```
elisno, line 38: I don't think the `Underperf` abbreviation here is necessary.

```suggestion
class UnderperformingGroupIssueManager(IssueManager):
```

cleanlab/datalab/internal/issue_manager/underperf_group.py
tataganesh: I think `max(unique_labels) + 1` makes sense. Initially, I had incorrectly assumed that the cluster labels would always be in the range `(0, n_clusters - 1)`.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,222 @@
...
+        hdb = HDBSCAN(metric="precomputed")
+        hdb.fit(knn_graph)
+        cluster_labels = hdb.labels_
+
+        unique_labels = set(cluster_labels)
+        n_outlier_clusters = sum(l in (-1, -2, -3) for l in unique_labels)
+        n_clusters = len(unique_labels) - n_outlier_clusters
+
+        min_loss = 10  # Assign some other value
+        min_loss_clusterid = -1
```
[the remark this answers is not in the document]
tataganesh: Regarding `(-1, -2, -3)`, here is the excerpt from HDBSCAN [documentation](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.HDBSCAN.html): 

Cluster labels for each point in the dataset given to fit. Outliers are labeled as follows:
* Noisy samples are given the label -1.
* Samples with infinite elements (+/- np.inf) are given the label -2.
* Samples with missing data are given the label -3, even if they also have infinite elements.

I am currently using `OUTLIER_LABELS = {"HDBSCAN": (-1, -2, -3)}` to store the outlier labels. The dictionary can be altered if other clustering techniques are added.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,222 @@
...
+        self.threshold = self._set_threshold(threshold)
+        self.k = k
+        self.near_duplicate_sets: List[List[int]] = []
+
+    def find_issues(
+        self,
+        features,
+        pred_probs,
+        **kwargs,
+    ) -> None:
```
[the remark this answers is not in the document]
tataganesh: Thank you, this does improve the code considerably. 
Also, the function for setting the KNN graph could be placed in a common module if this functionality is used by more than issue manager.

cleanlab/datalab/internal/issue_manager/underperf_group.py
tataganesh: The abbreviation has been replaced with the full version in most cases. Would you also prefer the file name to be `underperforming_group.py`?

tataganesh:
Major pending TODOs after latest commit

* Add `silhouette_score` and report time taken by its computation. 
* DataLab Tests.
* Description of Underperforming Issue in  Datalab Issue Type guide.

elisno (commented):
A few extra comments.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,281 @@
...
+                    f"Metric {self.metric} does not match metric {knn.metric} used to fit knn. "
+                    "Most likely an existing NearestNeighbors object was passed in, but a different "
+                    "metric was specified."
+                )
+            self.metric = knn.metric
+
+            try:
+                check_is_fitted(knn)
+            except NotFittedError:
+                knn.fit(features)
```
elisno, line 127: This only stores the `features`, no distances are stored.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,281 @@
...
+
+            knn_graph = knn.kneighbors_graph(mode="distance", n_neighbors=features.shape[0] - 1)
```
elisno, line 129: Make sure we use a sensible default value of `k` here. Computing the full-connectivity is not sensible for large datasets.

jwmueller: the default should just be the same as for the other issue types right? I don't think there should be a bunch of different `K` defaults

elisno: Agreed, `k=10` is the consistent default we've set across other issue types. It's a reasonable starting point for this context as well.

If our experiments show significant improvements in clustering accuracy with a higher value, we can consider adjusting it.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,281 @@
...
+            if self.metric is not None:
+                statistics_dict["statistics"]["knn_metric"] = self.metric
+
+        return statistics_dict
+
+    def _get_cluster_statistics(
+        self,
+        n_clusters: int,
+        cluster_labels: npt.NDArray[np.int_],
+        cluster_obj: ClusterMixin,
```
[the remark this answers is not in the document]
tataganesh: As per our previous discussion, I am going to decouple the cluster obj and only use `n_clusters`, `cluster_labels` and `knn_graph` for now.

cleanlab/datalab/internal/issue_manager/underperf_group.py
tataganesh: Some context  here - I had initially computed the knn graph with `k` as the number of data points, since HDBSCAN expects a symmetric distance matrix when `metric="precomputed"`.  Since this is computationally expensive, we have discussed to try and create a symmetric matrix using a smaller value of `k`, and as @elisno mentioned, experiments with different `k` can provide a better idea regarding clustering performance.

tataganesh:
**Update**: HDBSCAN fails if the KNN graph is disconnected. 
**Issue**: It seems like HDBSCAN [cannot work with a disconnected graph](https://github.com/scikit-learn-contrib/hdbscan/issues/82), i.e., if our knn graph is disjoint, then the mutual reachability graph might also be disjoint, which causes an exception. With a smaller k, the knn graph is likely to be disconnected.
**Report**: Here is a [colab notebook](https://colab.research.google.com/drive/1syMrTwPa9hkytLyV-TtzOlAXkMluTLhr?usp=sharing) to reproduce the error with explanation for the exception (which seems misleading from scikit-learn's side).
**Potential Solution** (by @elisno) - Perform clustering on each connected component and then aggregate all labels.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -126,24 +127,84 @@ def set_knn_graph(self, features: npt.NDArray, find_issues_kwargs: Dict) -> csr_
             except NotFittedError:
                 knn.fit(features)
 
-            knn_graph = knn.kneighbors_graph(mode="distance", n_neighbors=features.shape[0] - 1)
-        return knn_graph
+            knn_graph = knn.kneighbors_graph(mode="distance")
+            symmetric_knn_graph = self._symmetrize_knn_graph(knn_graph)
+        return symmetric_knn_graph, knn_graph
```
elisno, line 132: Don't change the return value of this method `set_knn_graph`. It should just return the `knn_graph`.

It should not be responsible for the symmetrization effort because it's sometimes an unnecessary step.

```suggestion
        return knn_graph
```

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -126,24 +127,84 @@ def set_knn_graph(self, features: npt.NDArray, find_issues_kwargs: Dict) -> csr_
...
+        return symmetric_distances
+
+    def perform_clustering(self, knn_graph: csr_matrix) -> npt.NDArray[np.int_]:
+        # hdb = HDBSCAN(metric="precomputed")
+        # hdb.fit(knn_graph)
+        # cluster_labels = hdb.labels_
+        # ALTERNATIVE FORMULATION: Connect the graph using a dummy node before HDBSCAN.
+        # Adapted from - https://github.com/scikit-learn-contrib/hdbscan/issues/82#issuecomment-1242604639
+        # knn_graph = self.connect_knn_graph(knn_graph)
+        cluster_labels = self.run_hdbscan_on_components(knn_graph)
```
elisno, line 149: Could we consider using a different variable name than `knn_graph` when transformations such as symmetrizing or adding any kind of bridge edges are anticipated?

Additionally, would it be possible to set [DBSCAN](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.DBSCAN.html#sklearn-cluster-dbscan) as the default clustering algorithm for the time being? This would ensure we have an algorithm that is guaranteed to function with an existing knn graph without necessitating excessive computation.

Note that it should operate with the knn_graph once it has been made symmetric, but it shouldn’t require a connected knn graph.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -126,24 +127,84 @@ def set_knn_graph(self, features: npt.NDArray, find_issues_kwargs: Dict) -> csr_
...
+            component_mask = component_labels == component_idx
+            component_subgraph = knn_graph[component_mask, :][:, component_mask]
+            knn_subgraph = self._symmetrize_knn_graph(component_subgraph)
+            clusterer = HDBSCAN(metric="precomputed")
+            component_clusters = clusterer.fit_predict(knn_subgraph)
+            outliers_mask = np.in1d(component_clusters, self.OUTLIER_LABELS)
+            component_clusters[~outliers_mask] += offset
+            all_clusters[component_mask] = component_clusters
+            offset = np.max(component_clusters) + 1
+        return all_clusters
```
elisno, line 171: Good to keep this. I'd suggest you make this an "internal" method. If we extract _symmetrize_knn_graph into a function, then this could be a function as well that we move to a separate utility module.

For now, just rename it to `_run_hdbscan_on_components`.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -126,24 +127,84 @@ def set_knn_graph(self, features: npt.NDArray, find_issues_kwargs: Dict) -> csr_
...
+                Sparse graph with only one component.
+        Source: https://github.com/scikit-learn-contrib/hdbscan/issues/82#issuecomment-1242604639
+        """
+        N = knn_graph.shape[0]
+        if new_node_dist is None:
+            new_node_dist = (knn_graph.max() - knn_graph.min()) * 1000
+        connected_knn_graph = knn_graph.copy()
+        connected_knn_graph = sp.vstack((connected_knn_graph, np.full((1, N), new_node_dist)))
+        connected_knn_graph = sp.hstack((connected_knn_graph, np.full((N + 1, 1), new_node_dist)))
+        return connected_knn_graph.tocsr()
```
elisno, line 201: Consider moving this to a utility module as a separate function in the near future.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -242,13 +301,12 @@ def _get_cluster_statistics(
...
         n_clusters: int,
         cluster_labels: npt.NDArray[np.int_],
-        cluster_obj: ClusterMixin,
         knn_graph: csr_matrix,
     ) -> Dict[str, Dict[str, Any]]:
         cluster_stats = {
             "clustering": {
-                "algorithm": cluster_obj.__class__.__name__,
-                "params": {**cluster_obj.get_params()},  # Only works for sklearn.cluster objects
+                "algorithm": "HDBSCAN",
```
elisno, line 308: Can we switch to `"DBSCAN"` as a default algorithm while we try to address the graph connectedness issue?
```suggestion
                "algorithm": "DBSCAN",
```

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -147,8 +147,7 @@ def perform_clustering(self, knn_graph: csr_matrix) -> npt.NDArray[np.int_]:
         }
         dbscan_params["metric"] = "precomputed"
         clusterer = DBSCAN(**dbscan_params)
-        clusterer.fit(knn_graph)
-        knn_graph.eliminate_zeros()  # Reset KNN Graph
+        clusterer.fit(knn_graph.copy())  # Copy to avoid modification
```
elisno, line 153: Does `DBSCAN` modify the `knn_graph`?

tataganesh: Yes. [Here](https://github.com/scikit-learn/scikit-learn/blob/2d8e03f4d5b3c466fb4542360bcef742a3a4e0a1/sklearn/cluster/_dbscan.py#L384-L389) is the code that modifies the `knn_graph`.

elisno: Ok, let's clarify this in the comment.

```suggestion
        clusterer.fit(knn_graph.copy())  # Copy to avoid modification by DBSCAN
```

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -159,6 +161,14 @@ def filter_cluster_ids(self, cluster_ids: npt.NDArray[np.int_]) -> npt.NDArray[n
         unique_cluster_ids = np.array(
             [label for label in set(cluster_ids) if label not in self.OUTLIER_LABELS]
         )
+        frequencies = np.bincount(cluster_ids[cluster_ids != -1])
```
elisno, line 164: make sure the cluster_ids is not in the outlier labels.

cleanlab/datalab/internal/issue_manager/underperf_group.py
jwmueller: If this is doubling the memory requirements for Datalab to run on somebody's machine, I would keep a GH Issue open to see if we can avoid this doubling. 

I wouldn't add this issue-type to the set of issues audited by default until somebody has investigated this question (can do that in a separate PR)

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,317 @@
...
+    computed using the :py:func:`get_self_confidence_for_each_label <cleanlab.rank.get_self_confidence_for_each_label>` function.
+    """
+    issue_name: ClassVar[str] = "underperforming_group"
+    verbosity_levels = {
+        0: [],
+        1: [],
+        2: ["threshold"],
+    }
+    OUTLIER_LABELS = (-1,)
+
```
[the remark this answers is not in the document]
elisno: Find a better name for this.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -67,6 +81,10 @@ def __init__(
         min_cluster_samples: int = 5,
         **_: Any,
     ):
+        """
+            Note: The `min_cluster_samples` argument should not be confused with the
+        `min_samples` argument of sklearn.cluster.DBSCAN.
+        """
```
elisno: This note belongs in the "class" docstring (the docstring that follow the `class UnderperformingGroupIssueManager` definition.
It's used instead of the __init__ method docstring when rendering the docs.

cleanlab/datalab/internal/issue_manager/underperf_group.py
```diff
@@ -0,0 +1,389 @@
...
+        performed_clustering: bool,
+        worst_cluster_id: int,
+    ) -> Dict[str, Dict[str, Any]]:
+        """Get relevant cluster statistics.
+
+        Args:
+            n_clusters (int): Number of clusters
+            cluster_ids (npt.NDArray[np.int_]): Cluster IDs for each datapoint.
+            performed_clustering (bool): If True, clustering was performed on
+            features passed to `find_issues`. If False, `cluster_ids` were explicitly
```
elisno, line 351: Slight change of wording, make sure to adjust this to pass the formatting check.

```suggestion
            performed_clustering (bool): Set to True to indicate that clustering was performed on
            `features` passed to `find_issues`. Set to False to suggest that `cluster_ids` were explicitly
```

elisno (changes requested):
Minor suggestions

elisno (approved):
Great work @tataganesh!

The code works well.
We'll continue work on adding this to the docs and general cleanup in a follow-up PR.

PR #843, opened 2023-09-05
Updated requirements for docs

sanjanag:
- Updated docs requirements.txt

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -19,7 +19,6 @@ tensorflow-io==0.26.0
 sentence-transformers==2.2.2
 speechbrain==0.5.13
 huggingface_hub==0.11.1
-fasttext-wheel==0.9.2
 torch==1.13.1
 torchvision==0.14.1
 torchaudio==0.13.1
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -21,9 +21,7 @@ speechbrain==0.5.13
 huggingface_hub==0.11.1
 fasttext-wheel==0.9.2
 torch==1.13.1
-skorch==0.12.1
 torchvision==0.14.1
 torchaudio==0.13.1
 timm==0.6.12
 datasets>=2.9.0
-cleanvision
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -52,7 +52,7 @@ jobs:
           restore-keys: ${{ runner.os }}-pip-
 
       - name: Install cleanlab
-        run: pip install -e .
+        run: pip install -e ".[all]"
 
       - name: Install dependencies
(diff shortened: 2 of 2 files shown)


docs/requirements.txt
```diff
@@ -21,9 +21,8 @@ speechbrain==0.5.13
 huggingface_hub==0.11.1
 fasttext-wheel==0.9.2
 torch==1.13.1
-skorch==0.12.1
 torchvision==0.14.1
 torchaudio==0.13.1
 timm==0.6.12
 datasets>=2.9.0
-cleanvision
+cleanvision>=0.3.1
```
elisno, line 28: Now we're tracking different version bounds in different places.

https://github.com/cleanlab/cleanlab/blob/14d1a86d4375beb17168eafdea88c84fd73c9a65/setup.py#L37

Can't we make `.[all]` an entry in this file? Maybe the entry can be editable? Something like:

```
-e ".[all]"
```

It would mean we'd have to update the docs/README.md.

docs/source/tutorials/dataset_health.ipynb
elisno, line 1: @sanjanag Maybe something more than the sleep call was committed in this file?

docs/requirements.txt
sanjanag: Sure, that would be easier to maintain.

elisno: I think we can safely remove the "cleanvision" line if the editable install works, but not things like datasets.
The former is never used/imported in the tutorials, while the latter is used for some loading some datasets.

docs/source/tutorials/dataset_health.ipynb
sanjanag: The notebook got formatted by the editor. I'll add it again with the previous formatting

docs/requirements.txt
```diff
@@ -21,9 +21,7 @@ speechbrain==0.5.13
 huggingface_hub==0.11.1
 fasttext-wheel==0.9.2
 torch==1.13.1
-skorch==0.12.1
```
elisno, line 24: Nice catch! Can confirm that `skorch` is [only imported in `tests/test_frameworks.py`](https://github.com/cleanlab/cleanlab/blob/14d1a86d4375beb17168eafdea88c84fd73c9a65/tests/test_frameworks.py#L41).

elisno (approved):
LGTM!

docs/requirements.txt
```diff
@@ -21,9 +21,7 @@ speechbrain==0.5.13
 huggingface_hub==0.11.1
 fasttext-wheel==0.9.2
```
elisno, line 22: I don't see this used anywhere in the docs.

```suggestion
```

jwmueller:
you two should just merge whenever you're good with this

PR #844, opened 2023-09-05
List comprehension to numpy ops for efficiency

tataganesh:
Port list comprehensions to numpy operations. 
Note - The `np.delete` change was derived from [this](https://stackoverflow.com/a/36502927/8505509) SO answer. 

<details><summary>Benchmarks</summary>

Thanks to @elisno for the benchmarking script!

### `np.delete` change.

drawing

### `get_self_confidence_for_each_label` change.
drawing

</details>

### DIFF
--- a/cleanlab/rank.py
+++ b/cleanlab/rank.py
@@ -505,7 +505,7 @@ def get_self_confidence_for_each_label(
     """
 
     # To make this work for multi-label (but it will slow down runtime), return:
-    # np.mean(pred_probs[np.arange(labels.shape[0]), labels], axis=1)
+    # np.array([np.mean(pred_probs[i, l]) for i, l in enumerate(labels)])
     return pred_probs[np.arange(labels.shape[0]), labels]
 
 
--- a/cleanlab/rank.py
+++ b/cleanlab/rank.py
@@ -504,9 +504,9 @@ def get_self_confidence_for_each_label(
       Lower scores indicate more likely mislabeled examples.
     """
 
-    # To make this work for multi-label (but it will slow down runtime), replace:
-    # pred_probs[i, l] -> np.mean(pred_probs[i, l])
-    return np.array([pred_probs[i, l] for i, l in enumerate(labels)])
+    # To make this work for multi-label (but it will slow down runtime), return:
+    # np.mean(pred_probs[np.arange(labels.shape[0]), labels], axis=1)
+    return pred_probs[np.arange(labels.shape[0]), labels]
 
 
 def get_normalized_margin_for_each_label(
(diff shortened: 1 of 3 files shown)


cleanlab/rank.py
```diff
@@ -504,9 +504,9 @@ def get_self_confidence_for_each_label(
       Lower scores indicate more likely mislabeled examples.
     """
 
-    # To make this work for multi-label (but it will slow down runtime), replace:
-    # pred_probs[i, l] -> np.mean(pred_probs[i, l])
-    return np.array([pred_probs[i, l] for i, l in enumerate(labels)])
+    # To make this work for multi-label (but it will slow down runtime), return:
+    # np.mean(pred_probs[np.arange(labels.shape[0]), labels], axis=1)
```
elisno, line 508: This won't work because `labels: List[List[int]]`.

It won't have a `shape` attribute and won't be a valid index.

tataganesh: Gotcha. One possible alternative is - 
```
labels = np.array(labels) # Convert List[List[int]] to np.ndarray
np.mean(pred_probs[np.arange(labels.shape[0]), labels], axis=1)
```
However, if `labels: List[List[int]]` is to be supported, then we might want to specify the parameter type as 
```
labels: Union(List[List[int]], np.ndarray)
```
in the future.

tataganesh: Edit: Deleted my previous comment because I didn't understand the issue properly.
What would you suggest here? I am not seeing a vectorized version.

elisno: Let's just stick to the working version.

```suggestion
    # np.array([np.mean(pred_probs[i, l]) for i, l in enumerate(labels)])
```

cleanlab/rank.py
```diff
@@ -544,8 +544,10 @@ def get_normalized_margin_for_each_label(
     """
 
     self_confidence = get_self_confidence_for_each_label(labels, pred_probs)
-    max_prob_not_label = np.array(
-        [max(np.delete(pred_probs[i], l, -1)) for i, l in enumerate(labels)]
+    N, D = pred_probs.shape
+    del_indices = np.arange(N) * D + labels
+    max_prob_not_label = np.max(
+        np.delete(pred_probs, del_indices, axis=None).reshape(N, D - 1), axis=-1
```
elisno, line 550:
```suggestion
    N, K = pred_probs.shape
    del_indices = np.arange(N) * K + labels
    max_prob_not_label = np.max(
        np.delete(pred_probs, del_indices, axis=None).reshape(N, K - 1), axis=-1
```

elisno (approved):
LGTM, great work @tataganesh!

Issue #820, opened 2023-08-12
Use existing Datalab to audit additional new data (e.g. test set)


jwmueller:
Current Datalab is specific to one dataset only. 
After the Datalab has been used to audit the original dataset, we want the fitted Datalab to be able to audit extra test data that is supposedly from the same underlying distribution. 

Here the audit of test data from the existing Datalab should **re-use statistics** computed from the original dataset (rather than re-computing them on the new dataset).  

The audit of additional test data can look for new types of issues as well, which are specific to discrepancies between the new data and original data:

- out of distribution examples in the new data that not similar to any of the original data
- new types of label issues in the new data (eg. new classes)
- drift between the new data and original data
- higher-loss between labels and predictions in the new data vs original data.

Important: We need a tutorial that makes it clear when it is preferable to do either of the following:

Option 1:  (can be done with existing functionality)

```
all_data = merge(original_data, new_data)
lab = Datalab(...) 
lab.find_issues(all_data, ...)
```

Option 2: (requires the new functionality to be built as specified in this GH issue)

```
lab = Datalab(...)
lab.find_issues(original_data, ...)

lab_new = Datalab(train_datalab = lab, ...)
lab_new.find_issues(new_data, ...)
```

The answer is Option 2 is preferable in settings where:
1. new data is streaming in quickly (and perhaps in small batches)
2. we assume new data to have same distribution as original data, and are specifically interested in issues that capture various failures of this assumption (OOD examples, new classes in new_data, etc.)

`lab_new.find_issues()` should not assume access to the original dataset, only things that have been stored in the original Datalab `lab`. This function should run fairly quickly when the `new_data` is small, regardless if `original_data` was large.

jwmueller:
If working on this, the starting point is: 

1. extend Datalab to be optionally instantiated from an existing Datalab (that was fitted to the original data aka training data)
2. first issue type to work on is detecting OOD examples in the new data (measure their KNN distance to the old data when new data features are provided), as demonstrated in: 
https://docs.cleanlab.ai/stable/tutorials/outliers.html

ChChCh8:
I'd like to work on this!

coding-famer:
I'd like to work on this!

jwmueller:
@coding-famer Amazing! Please sync with @elisno about details before commencing the implementation. This is something where figuring out the exact spec before coding will speed things up greatly.

PR #868, opened 2023-10-06
Use existing Datalab to audit additional new data

coding-famer:
For issue #820 
Datalab now can accept an existing Datalab instance and use its statistics information for finding issues for new data.
This only support label issue and `pred_probs` only.
As mentioned in the issue, I made a simple tutorial [here](https://github.com/coding-famer/cleanlab/blob/datalab-reuse-tutorial/docs/source/tutorials/datalab/datalab_reuse_trained.ipynb).

### DIFF
--- a/cleanlab/datalab/datalab.py
+++ b/cleanlab/datalab/datalab.py
@@ -120,9 +120,8 @@ class Datalab:
             warnings.warn(
                 "An existing Datalab instacne has been passed, the new datalab will use the existing statistics."
             )
-            self.data_issues._update_issue_info(
-                "statistics", trained_datalab.get_info("statistics")
-            )
+            for k in trained_datalab.get_info().keys():
+                self.data_issues._update_issue_info(k, trained_datalab.get_info(k))
             self._trained_statistics = True
         else:
             self._trained_statistics = False
--- a/cleanlab/datalab/datalab.py
+++ b/cleanlab/datalab/datalab.py
@@ -84,6 +84,10 @@ class Datalab:
         If specified, additional image-specific issue types can be detected in the dataset.
         See the CleanVision package `documentation <https://cleanvision.readthedocs.io/en/latest/>`_ for descriptions of these image-specific issue types.
 
+    trained_datalab : Datalab, optional
+        A Trained Datalab instance that contains statistics information from similar data distribution. This means that the `trained_datalab` should already call `find_issues` and collect statistic information.
+        If provided, the new datalab will use the existing statistics.
+
     verbosity : int, optional
         The higher the verbosity level, the more information
         Datalab prints when auditing a dataset.
--- a/cleanlab/datalab/datalab.py
+++ b/cleanlab/datalab/datalab.py
@@ -102,6 +102,7 @@ class Datalab:
         data: "DatasetLike",
         label_name: Optional[str] = None,
         image_key: Optional[str] = None,
+        trained_datalab: Optional[Datalab] = None,
         verbosity: int = 1,
     ) -> None:
         self._data = Data(data, label_name)
@@ -114,6 +115,16 @@ class Datalab:
         self.verbosity = verbosity
         self._imagelab = create_imagelab(dataset=self.data, image_key=image_key)
         self.data_issues = data_issues_factory(self._imagelab)(self._data)
+        if trained_datalab is not None:
+            warnings.warn(
+                "An existing Datalab instacne has been passed, the new datalab will use the existing statistics."
+            )
+            self.data_issues._update_issue_info(
+                "statistics", trained_datalab.get_info("statistics")
+            )
+            self.trained_statistics = True
+        else:
+            self.trained_statistics = False
 
     # todo: check displayer methods
     def __repr__(self) -> str:
@@ -139,6 +150,11 @@ class Datalab:
         If the dataset has no labels, returns an empty list.
         """
         return self._labels.class_names
+
+    @property
+    def has_trained_statistics(self) -> bool:
+        """Whether the dataset has trained statistics."""
+        return self.trained_statistics
 
     def find_issues(
         self,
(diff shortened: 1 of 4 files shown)


coding-famer:
I didn't add the tutorial to this PR, and it's quite simple now, but I'm glad to make a better tutorial to be added if needed.
I will add unit tests later.

cleanlab/datalab/datalab.py
```diff
@@ -114,6 +115,16 @@ def __init__(
...
         self._imagelab = create_imagelab(dataset=self.data, image_key=image_key)
         self.data_issues = data_issues_factory(self._imagelab)(self._data)
+        if trained_datalab is not None:
+            warnings.warn(
+                "An existing Datalab instacne has been passed, the new datalab will use the existing statistics."
+            )
+            self.data_issues._update_issue_info(
+                "statistics", trained_datalab.get_info("statistics")
+            )
+            self.trained_statistics = True
```
elisno, line 125: I noticed you added a `trained_statistics` field to the Datalab class. Considering the complexity and facade pattern of this class, I propose keeping `trained_statistics` as an internal field to minimize potential alterations in the class's external behavior and public API. Utilizing the `@property` decorator, as you’ve wisely done in other instances, offers a clean way to expose necessary data without directly exposing new fields. 

```suggestion
            self._trained_statistics = True
```

With this adjustment, the field is internal and can be manipulated or accessed via methods or properties, safeguarding against unintended external access or modification.

Could we also ensure that adequate inline comments or docstrings are added to clarify its usage and purpose internally, even if it won't be heavily documented in external documentation? This would help maintain clarity and understandability for future development.

Looking forward to your thoughts!

cleanlab/datalab/internal/issue_finder.py
```diff
@@ -219,6 +219,18 @@ def _resolve_required_args(self, pred_probs, features, knn_graph):
...
+        """For label error only now"""
+        confident_joint = kwargs.get("confident_joint", None)
+        py = kwargs.get("py", None)
+        noise_matrix = kwargs.get("noise_matrix", None)
+        inverse_noise_matrix = kwargs.get("inverse_noise_matrix", None)
+        issue_types["label"]["confident_joint"] = confident_joint
+        issue_types["label"]["py"] = py
+        issue_types["label"]["noise_matrix"] = noise_matrix
+        issue_types["label"]["inverse_noise_matrix"] = inverse_noise_matrix
+        return issue_types
```
elisno, line 232: I have a small suggestion to streamline the assignment of values from `kwargs` into `issue_types["label"]`. Something like:

```suggestion
    def _resolve_trained_statistics_args(self, issue_types: Dict[str, Any], **kwargs):
        """For label error only now"""
        keys = ["confident_joint", "py", "noise_matrix", "inverse_noise_matrix"]
        issue_types["label"].update({key: kwargs.get(key, None) for key in keys})
        return issue_types
```

This refactoring utilizes dictionary comprehension to avoid repetitive lines of code, while preserving the clarity of the logic. 


Tell me, should we consider omitting None values during the update to keep issue_types more concise?

```python
issue_types["label"].update({key: kwargs.get(key) for key in keys if kwargs.get(key) is not None})
```

---

Feel free to store some of the expressions as variables for clarity.

It's not just about shortening the code but enhancing readability and reducing potential future maintenance.

---

Could you also expand the docstring with a further explanation of what this helper function expects to find in the `issue_types` input?

cleanlab/datalab/internal/issue_finder.py
```diff
@@ -343,6 +355,17 @@ def get_available_issue_types(self, **kwargs):
...
+        label_use_trained_statistics = (
+            "label" in issue_types_copy
+            and self.datalab.has_trained_statistics
+            and not drop_label_check
+        )
+        if label_use_trained_statistics:
+            warnings.warn("Using precomputed statistics from a previously trained datalab. ")
+            issue_types_copy = self._resolve_trained_statistics_args(
+                issue_types_copy, kwargs=kwargs
+            )
```
elisno, line 367: I may need to test the different scenarios, but this is the first time we're adding a helper function that works with the `**kwargs`.

Maybe document more closely in the function what keyword arguments we want to use here.

cleanlab/datalab/datalab.py
```diff
@@ -102,6 +102,7 @@ def __init__(
         data: "DatasetLike",
         label_name: Optional[str] = None,
         image_key: Optional[str] = None,
+        trained_datalab: Optional[Datalab] = None,
```
elisno, line 105: We'll use this variable name during the development of this PR, but we should brainstorm better variable names for this argument before merging.

elisno (commented):
Thank you so much for your contribution @coding-famer!
This is a great start!
I have a few comments.

cleanlab/datalab/datalab.py
coding-famer: Yes, you are right, also add more comments.

cleanlab/datalab/internal/issue_finder.py
coding-famer: Good point. It seems that all these variables could not be `None` because we can't recalculate it? If so, I will add an assertion.

cleanlab/datalab/datalab.py
coding-famer: Yeah, this is also what I think.

jwmueller:
> I didn't add the tutorial to this PR, and it's quite simple now, but I'm glad to make a better tutorial to be added if needed. I will add unit tests later.

Could you add a tutorial? It make reviewing this work 10x easier with a tutorial. Updating the tutorial should be quite easy with underlying code changes, since the tutorial should only use default arguments. 

For the tutorial, I'd just use the same datalab quickstart tutorial dataset as training data. Then I'd have 1 batch of test data which is large, a 2nd batch of test data that is just a single datapoint, and a 3rd (and later 4th, 5th, 6th, ...) batch of test data that is just a single datapoint too.

coding-famer:
@jwmueller I will sync with @elisno tomorrow and may open a new PR for this feature, moving these to `/experimental`. And for the tutorial, we plan to add the use case to the `datalab_advanced` like [here](https://github.com/coding-famer/cleanlab/blob/datalab-reuse-tutorial/docs/source/tutorials/datalab/datalab_advanced.ipynb). (I used to have a simple tutorial in my local branch but deleted by myself and move those things to the end-to-end unit test.)

coding-famer:
Close and move to #1049

PR #881, opened 2023-11-02
Add image, title, and description for SEO link previews for the docs

elisno:
**Screenshots**

Here's how preview links to the docs would look like (ignore the firebaseapp url, it was just used for testing locally):

image

**Unaddressed Cases**

The image persists across all pages, but the title and description don't.
Is that something that should be done?

## References

- Sphinx extension, [sphinxext-opengraph](https://github.com/wpilibsuite/sphinxext-opengraph), is added in this PR.


cmauck10 (approved):
LGTM! Just add new card when Cris provides.

PR #895, opened 2023-11-13
Test properties of near duplicate sets

elisno:
## Summary

This PR aims to verify the correctness of near-duplicate sets found in knn-graphs by introducing property-based tests. It also modifies the `is_near_duplicate_issue` column to consider examples only within the sets.

## Impact

🌐 Areas Affected: 
- `cleanlab/datalab/internal/issue_manager/duplicate.py`
- Other changes add more tests.

👥 Who’s Affected:

Anyone constructing a `knn_graph: csr_matrix` with an ANN-search algorithm with low-recall is now guaranteed to find only examples in `near_duplicate_sets` that are flagged as `is_near_duplicate_issue == True`.

## Testing

> 🔍 Testing Done: A strategy for generating valid knn-graphs is added to `tests/datalab/issue_manager/test_knn_graph.py`. Its properties are tested in the same module. This strategy is used to generate test inputs to property-based tests for `near_duplicate_sets` found by the `NearDuplicateIssueManager`.
>
> 🔗 Test Case Link: [Main test class](https://github.com/cleanlab/cleanlab/compare/master...elisno:duplicate-sets-properties?expand=1#diff-5ecb578d192001a407b24144094a6c25631fa2cfe24bac88668574343a26170dR172-R199) verifies that the `near_duplicate_sets` satisfy the desired properties for any knn-graph. 

**Unaddressed Cases**

> - sklearn has [deprecated and removed the only known ann-search algorithm](https://scikit-learn.org/0.16/modules/generated/sklearn.neighbors.LSHForest.html) I am aware of.
> - To avoid adding extra test dependencies, the strategy generates only exact knn graphs.
> - These desired properties don't require a brute-force search algorithm and are valid for an approximate search algorithm that finds k_neighbors for every point.

## Reviewer Notes

When reviewing the changes in cleanlab/datalab/internal/issue_manager/duplicate.py, please note that 
```python
is_issue_column = nn_distances < self.threshold * np.median(nn_distances)
```
previously flagged examples whose 1-nearest-neighbor is within the given threshold. With the changes, the same right-hand side of the inequality is used to flag examples in a near-duplicate set. These sets were made symmetric in a [previous PR](https://github.com/cleanlab/cleanlab/pull/781).

### DIFF
--- a/tests/datalab/issue_manager/conftest.py
+++ b/tests/datalab/issue_manager/conftest.py
@@ -14,22 +14,39 @@ def knn_graph_strategy(draw, num_samples, k_neighbors):
 
     # Generate a symmetric distance matrix
     upper_triangle = [
-        draw(st.lists(st.floats(min_value=0, max_value=100, allow_nan=False, allow_infinity=False), 
-                     min_size=i, max_size=i, unique=True))
-        for i in range(1, num_samples+1)
+        draw(
+            st.lists(
+                st.floats(min_value=0, max_value=100, allow_nan=False, allow_infinity=False),
+                min_size=i,
+                max_size=i,
+                unique=True,
+            )
+        )
+        for i in range(1, num_samples + 1)
     ]
 
     distance_matrix = np.zeros((num_samples, num_samples))
     for i, row in enumerate(upper_triangle):
-        distance_matrix[i, :i+1] = row
-        distance_matrix[:i+1, i] = row
+        distance_matrix[i, : i + 1] = row
+        distance_matrix[: i + 1, i] = row
 
-    np.fill_diagonal(distance_matrix, np.inf)  # To ensure we don't select a point as its own neighbor
+    np.fill_diagonal(
+        distance_matrix, np.inf
+    )  # To ensure we don't select a point as its own neighbor
 
     # Compute k-nearest neighbors based on the distance matrix
     sorted_indices = np.argsort(distance_matrix, axis=1)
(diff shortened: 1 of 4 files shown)


tests/datalab/issue_manager/conftest.py
```diff
@@ -0,0 +1,52 @@
+from hypothesis import strategies as st
+import numpy as np
+from scipy.sparse import csr_matrix
+
+
+@st.composite
+def knn_graph_strategy(draw, num_samples, k_neighbors):
```
clu0, line 7: would be helpful to have a high level docstring for this function, describing what properties of the KNN graph you would expect from this function

tests/datalab/issue_manager/test_knn_graph.py
```diff
@@ -0,0 +1,66 @@
+import math
+from hypothesis import HealthCheck, given, settings, strategies as st
+import pytest
+
+from .conftest import knn_graph_strategy
+
+
+class TestKNNGraph:
```
clu0, line 8: question: this test is not for testing cleanlab functionality right? it's just to check that the `knn_graph_strategy` gives KNNs with the right properties?

tests/datalab/issue_manager/conftest.py
clu0: Also the naming of this function is a bit confusing, could we call it something like `create_knn_graph_with_strategy`?

clu0 (commented):
mostly looks good! just left some comments and questions

PR #896, opened 2023-11-16
Refactor Non-IID tests to reuse common code

tataganesh:
## Summary

The Non-IID test class for the datalab interface has been refactored to reuse code that is common across different test methods. This includes using a pytest fixture for the datalab instance and using a function for computing prediction probabilities using features. 

## Impact

Areas Affected: Class `TestDatalabFindNonIIDIssues` in `test_datalab.py`. 

## Testing

Testing Done: Verified that the tests run successfully after the refactor.


elisno (approved):
LGTM. Thanks @tataganesh!

Issue #899, opened 2023-11-20
Extend regression in Datalab to handle predictions as input


elisno:
This issue tracks work for handling `predictions` as inputs when finding label issues in Datlab for regression.

# Details
With #796 now merged, we can run label issue check for regression datasets.

The example shows how to take numerical features like so:

```python
from cleanlab import Datalab
import numpy as np

# Generate some toy-examples
N = 200
X = np.random.rand(N, 3)
y = 1 + 2 * X[:,0] + 3 * X[:,1] + 4 * X[:,2] + 0.05*np.random.randn(N)

# Run defaults
lab = Datalab(data={"y": y}, label_name='y', task="regression")
lab.find_issues(features=X)
```

We need to add a `predictions: Optional[np.ndarray] = None)` argument to https://github.com/cleanlab/cleanlab/blob/204651cf92301aa4d738cc2b5bdb9bd43125aa56/cleanlab/datalab/internal/issue_manager/regression/label.py#L87-L91

The idea is that something like this should work:

```python
from cleanlab import Datalab
import numpy as np

# Generate some toy-examples
N = 200
X = np.random.rand(N, 3)
y = 1 + 2 * X[:,0] + 3 * X[:,1] + 4 * X[:,2] + 0.05*np.random.randn(N)

# Run with existing predictions
from sklearn.neighbors import KNeighborsRegressor
# Not good modeling code, but just for example
y_pred = KNeighborsRegressor(n_neighbors=3).fit(X, y).predict(X)

lab = Datalab(data={"y": y}, label_name='y', task="regression")
lab.find_issues(predictions=y_pred, issue_types= {"label": {}})  # setting issue_types to focus on the issue manager.
```

When passing `predictions` to the LabelIssueManager for regression, the scores for examples should be computed with,

https://github.com/cleanlab/cleanlab/blob/0903c2a3614aa59dea6b826dfacbc995d5bf9c45/cleanlab/regression/rank.py#L38-L43

We still need to resolve how to flag examples with this approach (

## Note to contributors

When the argument is added to the issue manager, a quick sanity check would be to replace the last line in the code snippet above with:

```diff
-lab.find_issues(predictions=y_pred, issue_types= {"label": {}})
+lab.find_issues(issue_types={"label": {"predictions": y_pred}})
```

Defining the new `predictions` argument on the Datalab level will be the next logical step.

Both of these things must be done in the same PR.

[...]

PR #902, opened 2023-11-22
Predictions for regression in Datalab

elisno:
## Summary

Introduce the `predictions` argument to `RegressionLabelIssueManager.find_issues`.
The `pred_probs` argument will still be used on the Datalab-level. Internally, the `predictions` key will pick up the same value as the `pred_probs` key when setting up issue managers (for regression).

```python
# Example code snippet

from cleanlab import Datalab
import numpy as np
from sklearn.datasets import make_regression

# Generate some toy-examples
N = 100
X, y, W = make_regression(n_samples=N, n_features=3, noise=0.1, random_state=0, coef=True, bias=False)

# Get predictions from a regression model. Simplified for illustration.
y_pred = X @ W

# `predictions` used in find_issues
lab = Datalab(data={"y": y}, label_name='y', task="regression")
lab.find_issues(pred_probs=y_pred, issue_types= {"label": {}})

## Outputs
#Finding label issues ...
#
#Audit complete. 7 issues found in the dataset.
```

## Testing

> 🔍 Testing Done: Added specific tests for the RegressionLabelIssueManager. Write additional tests that run Datalab for regression with this new argument.

- This is a follow-up to #796.
- This PR addresses #899.

```python
# How to set a threshold for flagging issues based on predictions
lab.find_issues(pred_probs=..., issue_types={"label":{"threshold": 1.0}}}
```

### DIFF
--- a/cleanlab/datalab/datalab.py
+++ b/cleanlab/datalab/datalab.py
@@ -160,6 +160,7 @@ class Datalab:
         self,
         *,
         pred_probs: Optional[np.ndarray] = None,
+        predictions: Optional[np.ndarray] = None,
         features: Optional[npt.NDArray] = None,
         knn_graph: Optional[csr_matrix] = None,
         issue_types: Optional[Dict[str, Any]] = None,
@@ -188,6 +189,13 @@ class Datalab:
             To best detect label issues, provide this input obtained from the most accurate model you can produce.
 
             If provided, this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
+            Must not be provided if `predictions` is provided.
+
+        predictions :
+            Out-of-sample predicted output variables made by the model for every example in the dataset.
+            To best detect label issues, provide this input obtained from the most accurate model you can produce.
+
+            Must not be provided if `pred_probs` is provided.
 
         features : Optional[np.ndarray]
             Feature embeddings (vector representations) of every example in the dataset.
@@ -301,6 +309,9 @@ class Datalab:
                 >>> # lab.find_issues(pred_probs=pred_probs, issue_types=issue_types)
 
         """
+        # Check for conflicting inputs
+        _ensure_single_prediction_input(pred_probs, predictions, task=self.task)
+
         if issue_types is not None and not issue_types:
             warnings.warn(
                 "No issue types were specified so no issues will be found in the dataset. Set `issue_types` as None to consider a default set of issues."
@@ -311,6 +322,7 @@ class Datalab:
         )
         issue_finder.find_issues(
             pred_probs=pred_probs,
+            predictions=predictions,
             features=features,
             knn_graph=knn_graph,
             issue_types=issue_types,
@@ -576,3 +588,33 @@ class Datalab:
         load_message = f"Datalab loaded from folder: {path}"
         print(load_message)
         return datalab
+
+def _ensure_single_prediction_input(pred_probs: Any, predictions: Any, task: str) -> None:
+    """Helper function for resolving conflicting inputs.
+    
+    For greater coverage, it checks if both pred_probs and predictions are provided, regardless of task.
+    If that's not the case, it checks if the provided input is unsupported for the given task.
+
+    Raises
+    ------
+    ValueError
+        If both `pred_probs` and `predictions` are provided.
+        Or if `pred_probs` is provided for a regression task.
+        Or if `predictions` is provided for a classification task.
+    """
+    # Edge-case, someone passes both pred_probs and predictions
+    error_msg = ""
+    if pred_probs is not None and predictions is not None:
+        error_msg = "Only one of `pred_probs` or `predictions` can be provided."
+        task_tips = {
+            "classification": "pred_probs",
+            "regression": "predictions",
+        }
+        if task in task_tips:
+            error_msg += f" For a {task} task, consider passing `{task_tips[task]}` instead."
+    elif task == "classification" and predictions is not None:
+        error_msg = "`predictions` is not supported for classification tasks. Consider passing `pred_probs` instead."
+    elif task == "regression" and pred_probs is not None:
+        error_msg = "`pred_probs` is not supported for regression tasks. Consider passing `predictions` instead."
+    if error_msg:
+        raise ValueError(error_msg)
--- a/cleanlab/datalab/datalab.py
+++ b/cleanlab/datalab/datalab.py
@@ -589,9 +589,10 @@ class Datalab:
         print(load_message)
         return datalab
 
+
 def _ensure_single_prediction_input(pred_probs: Any, predictions: Any, task: str) -> None:
     """Helper function for resolving conflicting inputs.
-    
+
     For greater coverage, it checks if both pred_probs and predictions are provided, regardless of task.
     If that's not the case, it checks if the provided input is unsupported for the given task.
 
--- a/cleanlab/datalab/datalab.py
+++ b/cleanlab/datalab/datalab.py
@@ -160,7 +160,6 @@ class Datalab:
         self,
         *,
         pred_probs: Optional[np.ndarray] = None,
-        predictions: Optional[np.ndarray] = None,
         features: Optional[npt.NDArray] = None,
         knn_graph: Optional[csr_matrix] = None,
         issue_types: Optional[Dict[str, Any]] = None,
@@ -189,13 +188,6 @@ class Datalab:
             To best detect label issues, provide this input obtained from the most accurate model you can produce.
 
             If provided, this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
-            Must not be provided if `predictions` is provided.
-
-        predictions :
-            Out-of-sample predicted output variables made by the model for every example in the dataset.
-            To best detect label issues, provide this input obtained from the most accurate model you can produce.
-
-            Must not be provided if `pred_probs` is provided.
 
         features : Optional[np.ndarray]
             Feature embeddings (vector representations) of every example in the dataset.
@@ -309,8 +301,6 @@ class Datalab:
                 >>> # lab.find_issues(pred_probs=pred_probs, issue_types=issue_types)
 
         """
-        # Check for conflicting inputs
-        _ensure_single_prediction_input(pred_probs, predictions, task=self.task)
 
         if issue_types is not None and not issue_types:
             warnings.warn(
@@ -322,7 +312,6 @@ class Datalab:
         )
         issue_finder.find_issues(
             pred_probs=pred_probs,
-            predictions=predictions,
             features=features,
             knn_graph=knn_graph,
             issue_types=issue_types,
@@ -588,34 +577,3 @@ class Datalab:
         load_message = f"Datalab loaded from folder: {path}"
         print(load_message)
         return datalab
-
-
-def _ensure_single_prediction_input(pred_probs: Any, predictions: Any, task: str) -> None:
-    """Helper function for resolving conflicting inputs.
-
-    For greater coverage, it checks if both pred_probs and predictions are provided, regardless of task.
-    If that's not the case, it checks if the provided input is unsupported for the given task.
-
-    Raises
-    ------
-    ValueError
-        If both `pred_probs` and `predictions` are provided.
-        Or if `pred_probs` is provided for a regression task.
-        Or if `predictions` is provided for a classification task.
-    """
-    # Edge-case, someone passes both pred_probs and predictions
-    error_msg = ""
-    if pred_probs is not None and predictions is not None:
-        error_msg = "Only one of `pred_probs` or `predictions` can be provided."
-        task_tips = {
-            "classification": "pred_probs",
-            "regression": "predictions",
-        }
-        if task in task_tips:
-            error_msg += f" For a {task} task, consider passing `{task_tips[task]}` instead."
-    elif task == "classification" and predictions is not None:
-        error_msg = "`predictions` is not supported for classification tasks. Consider passing `pred_probs` instead."
-    elif task == "regression" and pred_probs is not None:
-        error_msg = "`pred_probs` is not supported for regression tasks. Consider passing `predictions` instead."
-    if error_msg:
-        raise ValueError(error_msg)
--- a/cleanlab/datalab/datalab.py
+++ b/cleanlab/datalab/datalab.py
@@ -187,8 +187,8 @@ class Datalab:
             Out-of-sample predicted class probabilities made by the model for every example in the dataset.
             To best detect label issues, provide this input obtained from the most accurate model you can produce.
 
-            If provided (for classification), this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
-            If provided (for regression), this must be a 1D array with shape (num_examples,).
+            If provided for classification, this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
+            If provided for regression, this must be a 1D array with shape (num_examples,).
 
         features : Optional[np.ndarray]
             Feature embeddings (vector representations) of every example in the dataset.
--- a/cleanlab/datalab/datalab.py
+++ b/cleanlab/datalab/datalab.py
@@ -187,7 +187,8 @@ class Datalab:
             Out-of-sample predicted class probabilities made by the model for every example in the dataset.
             To best detect label issues, provide this input obtained from the most accurate model you can produce.
 
-            If provided, this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
+            If provided (for classification), this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
+            If provided (for regression), this must be a 1D array with shape (num_examples,).
 
         features : Optional[np.ndarray]
             Feature embeddings (vector representations) of every example in the dataset.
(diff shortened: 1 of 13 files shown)


cleanlab/datalab/internal/issue_manager/regression/label.py
```diff
@@ -38,6 +39,10 @@ class RegressionLabelIssueManager(IssueManager):
     clean_learning_kwargs :
         Keyword arguments to pass to the :py:meth:`CleanLearning <cleanlab.regression.learn.CleanLearning>` constructor.
 
+    threshold :
+        The threshold to use to determine if an example has a label issue. Only used if
```
jwmueller, line 43: this is a multiplier right not an absolute threshold

tests/datalab/test_datalab.py
```diff
@@ -980,53 +980,109 @@ def test_pred_probs_precedence(self, pred_probs, random_embeddings):
...
+        X = regression_data["X"]
+        lab.find_issues(features=X)
+        lab.report()
+        summary = lab.get_issue_summary()
+
+        assert "label" in summary["issue_type"].values
+        assert (summary[summary["issue_type"] == "label"]["num_issues"] == 40).all()
+        assert np.isclose(
+            summary[summary["issue_type"] == "label"]["score"].values[0], 0.672974, atol=1e-5
+        )
```
jwmueller, line 1041: can we assert which of the datapoints have issues, best test is assert that checks for high jaccard similarity between set of flagged issues and set of true issues

tests/datalab/test_datalab.py
```diff
@@ -980,53 +980,109 @@ def test_pred_probs_precedence(self, pred_probs, random_embeddings):
...
+        model = KNeighborsRegressor(n_neighbors=5)
+        X = regression_data["X"]
+        lab.find_issues(
+            features=X, issue_types={"label": {"clean_learning_kwargs": {"model": model}}}
+        )
+        summary = lab.get_issue_summary()
+        assert (summary[summary["issue_type"] == "label"]["num_issues"] == 30).all()
+        assert np.isclose(
+            summary[summary["issue_type"] == "label"]["score"].values[0], 0.777305, atol=1e-5
+        )
```
jwmueller, line 1085: can we assert which of the datapoints have issues, best test is assert that checks for high jaccard similarity between set of flagged issues and set of true issues

cleanlab/datalab/datalab.py
```diff
@@ -188,6 +189,13 @@ def find_issues(
             To best detect label issues, provide this input obtained from the most accurate model you can produce.
 
             If provided, this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
+            Must not be provided if `predictions` is provided.
```
jwmueller, line 192:
```suggestion
```

jwmueller: philosphy-wise, let's focus the current documentation mostly around `pred_probs` still being the primary input to Datalab. 

So keep the explanations of when to use `predictions` vs. pred_probs contained within the docstring of the `predictions` argument.

cleanlab/datalab/datalab.py
```diff
@@ -188,6 +189,13 @@ def find_issues(
...
+
+        predictions :
+            Out-of-sample predicted output variables made by the model for every example in the dataset.
+            To best detect label issues, provide this input obtained from the most accurate model you can produce.
+
+            Must not be provided if `pred_probs` is provided.
```
jwmueller, line 198: needs more explanation about why this exists vs. just pred_probs and when to use one vs. the other

jwmueller: what is long-term plan here? 

deprecate `pred_probs` and have all tasks use `predictions`?

If so, I'd do followup PR where you allow classification tasks to work with `predictions` as well, internally you could just set:
```
if predictions is not None:
    pred_probs = predictions 
```

jwmueller: If ^ this is the long term plan, then will eventually need the `predictions` docstring to be quite involved for each task type. And should soon move the docstring for `pred_probs` into dosctring for `predictions` as well (in a subsection when the task is classification), so that `pred_probs` would start to be de-emphasized.

Alternatively if planning to keep `pred_probs` around for a while, want to think of plan on how to support both it and `predictions` in a nicer way

cleanlab/datalab/datalab.py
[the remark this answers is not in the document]
elisno: While resolving these things, I think it's better to look at pred_probs as the canonical/primary input for all tasks, unless specified otherwise (e.g. predictions for regression).

elisno: The safest approach here is to have a unified argument at the user-level. We'll keep `pred_probs` for all tasks, but don't have to modify the signatures of low-level classes like `RegressionLabelIssueManager`.

What these things have in common is that these are model-outputs that we use as inputs to Datalab. For  `pred_probs` and `predictions` in particular, they are just prediction-based model outputs.
I've added helper classes for handling each type of model output, but that won't change the top-level API for the user that use `Datalab.find_issues(pred_probs=...)`.
It's backwards compatible, easily extensible to other types of tasks and the other argument types like `features`, even `knn_graph`.

These helper classes are just an intermediate representation that won't get passed on to the issue managers directly, so this requires minimal changes.

elisno: This means I've removed the docstring for the `predictions` argument, but clarified the expected layout/structure of `pred_probs` in the relevant classes. That's much easier to maintain.

cleanlab/datalab/datalab.py
elisno: We get the best of both worlds with the approach I briefly mention in https://github.com/cleanlab/cleanlab/pull/902#discussion_r1412323597.

We'll keep only `pred_probs` for backwards compatibility at the user-level.
The `RegressionLabelIssueManager` still uses the `predictions`  argument name, but these implementation details are handled in an intermediate stage when calling `Datalab.find_issues(pred_probs=...)`.
This gives us more time to think if we want to keep two, mutually exclusive arguments in `Datalab.find_issues` when one will suffice.

Docstring maintenance is also much simpler this way.

cleanlab/datalab/datalab.py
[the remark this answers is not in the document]
elisno: Removed.

tests/datalab/test_datalab.py
jwmueller: this is not an end-to-end test though of what matters. What matters is that this code can accurately detect label issues. I'm not sure where that's being tested (maybe I overlooked)

cleanlab/datalab/datalab.py
```diff
@@ -576,3 +588,34 @@ def load(path: str, data: Optional[Dataset] = None) -> "Datalab":
...
+            "regression": "predictions",
+        }
+        if task in task_tips:
+            error_msg += f" For a {task} task, consider passing `{task_tips[task]}` instead."
+    elif task == "classification" and predictions is not None:
+        error_msg = "`predictions` is not supported for classification tasks. Consider passing `pred_probs` instead."
+    elif task == "regression" and pred_probs is not None:
+        error_msg = "`pred_probs` is not supported for regression tasks. Consider passing `predictions` instead."
+    if error_msg:
+        raise ValueError(error_msg)
```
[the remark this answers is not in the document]
elisno: This piece of code has now been removed in favor of some new helper classes for validating the `pred_probs` inputs based on tasks.
Consider this resolved.

tests/datalab/test_datalab.py
elisno: I've updated the dataset for testing and asserted which datapoints have issues. So this should be resolved.

The tricky thing is that the "real issues" might have smaller residuals and are not picked up (leading to a lower Jaccard similarity).

cleanlab/datalab/internal/issue_manager/regression/label.py
elisno: Correct. The docstring has been updated.

tests/datalab/test_datalab.py
elisno: I've updated the dataset for testing and asserted which datapoints have issues. So this should be resolved.

The tricky thing is that the "real issues" might have smaller residuals and are not picked up (leading to a lower Jaccard similarity).

huiwengoh (approved):
lgtm!

cleanlab/datalab/datalab.py
```diff
@@ -187,7 +187,8 @@ def find_issues(
             Out-of-sample predicted class probabilities made by the model for every example in the dataset.
             To best detect label issues, provide this input obtained from the most accurate model you can produce.
 
-            If provided, this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
+            If provided for classification, this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
+            If provided for regression, this must be a 1D array with shape (num_examples,).
```
jwmueller, line 191:
```suggestion
            For classification data, this must be a 2D array with shape ``(num_examples, K)`` where ``K`` is the number of classes in the dataset.
            For regression data, this must be a 1D array with shape ``(num_examples,)`` containing the predicted value for each example.
```

docs/source/cleanlab/datalab/guide/custom_issue_manager.rst
```diff
@@ -18,9 +18,10 @@ to detect a custom-defined type of issue alongside the pre-defined issue types i
         .. code-block:: python
 
             from cleanlab.datalab.internal.issue_manager_factory import register
-            register(MyIssueManager)
+            register(MyIssueManager)  # Defaults to task="classification"
+            # register(MyIssueManagerForRegression, task="regression") # For regression tasks
```
jwmueller, line 22:
```suggestion
            # register(MyIssueManagerForRegression, task="regression")  # Alternative for regression tasks
```

cleanlab/datalab/internal/issue_manager/regression/label.py
```diff
@@ -58,50 +64,53 @@ def __init__(
         self,
         datalab: Datalab,
         clean_learning_kwargs: Optional[Dict[str, Any]] = None,
+        threshold: float = 0.1,
```
jwmueller, line 67:
```suggestion
        threshold: float = 0.13,
```

jwmueller: consider this alternate default value that would match near duplicates

jwmueller: disregard if that's a lot more work, or 0.1 is better default

cleanlab/datalab/internal/issue_manager/regression/label.py
```diff
@@ -58,50 +64,53 @@ def __init__(
...
+        self.threshold = threshold
 
     def find_issues(
         self,
-        features: np.ndarray,
+        features: Optional[np.ndarray] = None,
+        predictions: Optional[np.ndarray] = None,
         **kwargs,
     ) -> None:
         """Find label issues in the datalab."""
```
jwmueller, line 83: you should write in english the priority order that happens inside here, eg:

If model is provided and features, then will use CleanLearning. Otherwise if pred_probs are provided, then will use them, etc etc. 

Otherwise this knowledge will be lost to time, if not explicitly documented

tests/datalab/test_datalab.py
jwmueller: Sorry maybe I'm dense and don't see it. Where is the test that is checking for high jaccard similarity between set of flagged issues and set of true issues?

All I see is:

```
assert (summary[summary["issue_type"] == "label"]["num_issues"] == 40).all()
```

which is only testing how many issues were estimated (which we don't care about testing at all actually, we only care about testing which datapoints were flagged as issues and that the overlap between this set and the set of true erroneous values is sufficiently high).

tests/datalab/test_datalab.py
```diff
@@ -981,53 +981,148 @@ def test_pred_probs_precedence(self, pred_probs, random_embeddings):
...
+    def regression_data(
+        self, num_examples=200, num_features=3, noise=0.01, error_frac=0.05, error_noise=10
+    ):
+        np.random.seed(SEED)
+        X = np.random.random(size=(num_examples, num_features))
+        coefficients = np.random.uniform(-1, 1, size=num_features)
+        label_noise = np.random.normal(scale=noise, size=num_examples)
+
+        true_y = np.dot(X, coefficients)
+        y = np.dot(X, coefficients) + label_noise
```
jwmueller, line 995:
```suggestion
        true_y = np.dot(X, coefficients)
        y = np.dot(X, coefficients)
```

jwmueller: why make our test more complicated by adding extra noise here?

tests/datalab/test_datalab.py
```diff
@@ -981,53 +981,148 @@ def test_pred_probs_precedence(self, pred_probs, random_embeddings):
...
+
+        # add extra noisy examples
+        num_errors = int(num_examples * error_frac)
+        extra_noise = np.random.normal(scale=error_noise, size=num_errors)
```
jwmueller, line 999:
```suggestion
        extra_noise = np.random.normal(scale=error_noise, size=num_errors) * max(1, error_noise)
```

jwmueller: idea here is the test will be more stable if the extra_noise is more obvious. I don't think there's any reason to make this test challenging, it should be obvious errors IMO

jwmueller (commented):
still had a couple of questions about the tests

cleanlab/datalab/internal/issue_manager/regression/label.py
elisno: Added a note

image

tests/datalab/test_datalab.py
elisno: I've simplified the noise for this test, it seems to work now after some trial and error.

tests/datalab/test_datalab.py
elisno: The tests checking the high Jaccard similarity were placed behind those lines.

But now the summary assertions have been removed.
I've added some Jaccard Similarity assertions, as well as low FPR assertions as well.

tests/datalab/test_datalab.py
elisno: The tests checking the high Jaccard similarity were placed behind those lines.

But now the summary assertions have been removed.
I've added some Jaccard Similarity assertions, as well as low FPR assertions as well.

tests/datalab/test_datalab.py
jwmueller: in future tests like this, can also shrink size of `num_errors` so that it's not affecting the ML model too much during training

cleanlab/datalab/internal/issue_finder.py
```diff
@@ -210,7 +211,8 @@ def find_issues(
             Out-of-sample predicted class probabilities made by the model for every example in the dataset.
             To best detect label issues, provide this input obtained from the most accurate model you can produce.
 
-            If provided, this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
+            If provided for classification, this must be a 2D array with shape (num_examples, K) where K is the number of classes in the dataset.
+            If provided for regression, this must be a 1D array with shape (num_examples,).
```
jwmueller, line 215:
```suggestion
            If provided for classification, this must be a 2D array with shape ``(num_examples, K)`` where K is the number of classes in the dataset.
            If provided for regression, this must be a 1D array with shape ``(num_examples,)``.
```

cleanlab/datalab/internal/issue_manager/regression/label.py
jwmueller: @elisno did you try 0.13 instead of 0.1?

docs/source/cleanlab/datalab/internal/issue_manager/index.rst
```diff
@@ -25,4 +25,11 @@ These are the issue managers that Datalab has not registered (yet).
 
 .. toctree::
     null
+
+Task-specific issue managers
+----------------------------
```
jwmueller, line 30:
```suggestion
ML task-specific issue managers
---------------------------------
```

tests/datalab/test_datalab.py
```diff
@@ -981,53 +981,144 @@ def test_pred_probs_precedence(self, pred_probs, random_embeddings):
...
+        # jaccard similarity
+        intersection = len(list(set(issue_ids).intersection(set(expected_issue_ids))))
+        union = len(set(issue_ids)) + len(set(expected_issue_ids)) - intersection
+        assert float(intersection) / union > 0.3
 
-    assert "label" in summary["issue_type"].values
-    assert (summary[summary["issue_type"] == "label"]["num_issues"] == 40).all()
+        # FPR
+        fpr = len(list(set(issue_ids).difference(set(expected_issue_ids)))) / len(issue_ids)
+        assert fpr < 0.3
```
jwmueller, line 1121: open GH issue to improve this `fpr` in the future

cleanlab/datalab/internal/issue_manager/regression/label.py
jwmueller: nevermind I see the false-positive rate is too high in the unit-test already, so let's not increase threshold

jwmueller (approved):
LGTM. Go ahead and merge after addressing my remaining minor comments

tests/datalab/test_datalab.py
elisno: - [ ] Improve the stability of results from end-to-end tests of Datalab with label error-detection for regression tasks

Issue #903, opened 2023-11-26
make null issue type more user-facing


jwmueller:
Should be mentioned in datalab issue guide and relevant docstrings.
https://docs.cleanlab.ai/master/cleanlab/datalab/guide/issue_type_description.html

This issue type can also be added to the default Datalab audit at this point, I think it'll be useful to check for by default

https://github.com/cleanlab/cleanlab/pull/856

Make sure to also update text in the tutorials that show this issue type once it's in the defaults

PR #917, opened 2023-12-13
Null Issue Manager: Vectorize loops, Fix row index bug

tataganesh:
## Summary

🎯 **Purpose**: This PR replaces expensive for-loops in the Null Issue Manager implementation with their vectorized equivalents. A bug related to row indices is also fixed and tests are improved to check these indices. 

**Row Indices bug**:
In the `_most_common_issue` function, the row ids with the most frequent pattern is determined. The [computation](https://github.com/cleanlab/cleanlab/blob/f3a65b8c18643e3fb9626988d606f839f8daea9b/cleanlab/datalab/internal/issue_manager/null.py#L108-L110) of these row ids is based on a list representing only null rows as a string. It is not with respect to the original feature matrix.  
The test for the `rows_affected` key was passing even with this bug, so it was updated. 

## Impact

🌐 Areas Affected: Null Issue Manager
 👥 Who’s Affected: NA

## Testing

🔍 Testing Done: Benchmarking was performed with the old and new versions of two functions. Here is the [notebook](https://gist.github.com/tataganesh/3fbf206bc809f519ed839146e68a36ea) to reproduce these benchmarks. 

Function `_calculate_null_issues`

vectorized_func1

Function `_most_common_issue`

vectorized_func2

Part of #903

### DIFF
--- a/tests/datalab/issue_manager/test_null.py
+++ b/tests/datalab/issue_manager/test_null.py
@@ -20,7 +20,7 @@ class TestNullIssueManager:
     def embeddings_with_null(self):
         np.random.seed(SEED)
         embeddings_array = np.random.random((4, 3))
-        embeddings_array[0][0] = np.NaN
+        embeddings_array[[0, 3], 0] = np.NaN
         embeddings_array[1] = np.NaN
         return embeddings_array
 
@@ -65,7 +65,7 @@ class TestNullIssueManager:
             issues_sort["is_null_issue"] == expected_sorted_issue_mask
         ), "Issue mask should be correct"
         assert summary_sort["issue_type"][0] == "null"
-        assert summary_sort["score"][0] == pytest.approx(expected=8 / 12, abs=1e-7)
+        assert summary_sort["score"][0] == pytest.approx(expected=7 / 12, abs=1e-7)
         assert (
             info_sort.get("average_null_score", None) is not None
         ), "Should have average null score"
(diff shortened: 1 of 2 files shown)


tests/datalab/issue_manager/test_null.py
```diff
@@ -20,7 +20,7 @@ def embeddings(self):
     def embeddings_with_null(self):
         np.random.seed(SEED)
         embeddings_array = np.random.random((4, 3))
-        embeddings_array[0][0] = np.NaN
+        embeddings_array[[0, 3], 0] = np.NaN
```
elisno, line 23: Why is this array being changed?

tests/datalab/issue_manager/test_null.py
```diff
@@ -65,7 +65,7 @@ def test_find_issues_with_null(self, issue_manager, embeddings_with_null):
             issues_sort["is_null_issue"] == expected_sorted_issue_mask
         ), "Issue mask should be correct"
         assert summary_sort["issue_type"][0] == "null"
-        assert summary_sort["score"][0] == pytest.approx(expected=8 / 12, abs=1e-7)
+        assert summary_sort["score"][0] == pytest.approx(expected=7 / 12, abs=1e-7)
```
elisno, line 68: Why is this being changed?

elisno (commented):
I left some comments on the changed tests.

Can you double check that the output tuple of the `_calculate_null_issues` are compared, and not just one of the entries.

Similarily, the equality check for the other function has been commented out. Update the notebook to make sure it actually works.

tataganesh:
@elisno I have updated the notebook with the equality check for the new and old function outputs. Note that for `_most_common_issue`, the `rows_affected` key is not checked since the original implementation for it is buggy. 

I have updated the description with an explanation for the  bug in the computation of `rows_affected`. The test input and the outputs were both updated to make sure that the test fails for the buggy code and to test the correct behaviour.

tests/datalab/issue_manager/test_null.py
elisno: I see. The `rows_affected` were not computed correctly before, but now we can assert that they should be 0 and 3.

tests/datalab/issue_manager/test_null.py
elisno: Because we added an example with a nan value, the score will be lower.

elisno (approved):
LGTM

PR #919, opened 2023-12-16
Patch: epsilon for near duplicate

ryansingman:
## Summary

> 🎯 **Purpose**: Update near duplicate check to ensure threshold is non-zero. This allows for datasets with majority of examples containing near duplicate to be more correctly marked as duplicates. Previously, no example in such dataset was correctly marked as near duplicate. With a non-zero threshold, any dataset has a chance of containing a near duplicate now based on the near-duplicate score.

## Impact

> 👥 Who’s Affected: Any near duplicate benchmark test with datasets containing significant amount of near duplicates will now see more examples correctly marked as near duplicates.


elisno (approved):
LGTM

Issue #886, opened 2023-11-09
make data valuation more user-facing


jwmueller:
add documentation, tutorial, etc. All the missing pieces that prevent this from being easily used/discovered.

There is currently nothing about it in the docs:  
https://docs.cleanlab.ai/master/search.html?q=valuation&check_keywords=yes&area=default

jwmueller:
@coding-famer has expressed interest in working on this

PR #925, opened 2023-12-18
make data valuation issue manager  user facing

coding-famer:
## Summary

> 🎯 **Purpose**: Make the data valuation issue manager user facing.

## Impact

> 🌐 Areas Affected: [issue_type_descriptions.rst](https://github.com/cleanlab/cleanlab/blob/master/docs/source/cleanlab/datalab/guide/issue_type_description.rst).

#886

### DIFF
--- a/cleanlab/datalab/internal/issue_manager/data_valuation.py
+++ b/cleanlab/datalab/internal/issue_manager/data_valuation.py
@@ -37,7 +37,22 @@ if TYPE_CHECKING:  # pragma: no cover
 
 
 class DataValuationIssueManager(IssueManager):
-    """Manages data sample with low valuation."""
+    """
+    Manages data sample with low valuation.
+
+    Examples
+    --------
+    >>> from cleanlab import Datalab
+    >>> import numpy as np
+    >>> X = np.random.normal(size=(50, 2))
+    >>> y = np.random.randint(2, size=50)
+    >>> pred_probs = X / X.sum(axis=1, keepdims=True)
+    >>> data = {"X": X, "y": y}
+    >>> lab = Datalab(data, label_name="y")
+    >>> lab.find_issues()
+    >>> issue_types={"data_valuation": {"k": 10, "threshold": 1e-6}}
+    >>> lab.find_issues(issue_types=issue_types)
+    """
 
     description: ClassVar[
         str
--- a/cleanlab/datalab/internal/issue_manager/data_valuation.py
+++ b/cleanlab/datalab/internal/issue_manager/data_valuation.py
@@ -81,7 +81,7 @@ class DataValuationIssueManager(IssueManager):
         assert knn_graph is not None, "knn_graph must be already calculated by other issue managers"
         assert labels is not None, "labels must be provided"
 
-        scores = _knn_shapley_score(knn_graph, labels)
+        scores = _knn_shapley_score(knn_graph, labels, self.k)
 
         self.issues = pd.DataFrame(
             {
@@ -125,12 +125,15 @@ class DataValuationIssueManager(IssueManager):
         return info_dict
 
 
-def _knn_shapley_score(knn_graph: csr_matrix, labels: np.ndarray) -> np.ndarray:
+def _knn_shapley_score(knn_graph: csr_matrix, labels: np.ndarray, k: int) -> np.ndarray:
     """Compute the Shapley values of data points based on a knn graph."""
     N = labels.shape[0]
     scores = np.zeros((N, N))
     dist = knn_graph.indices.reshape(N, -1)
-    k = dist.shape[1]
+    knn_graph_k = dist.shape[1]
+    if k > knn_graph_k:
+        k = knn_graph_k
+        Warning(f"k is larger than the number of neighbors in the knn graph. Using k={k} instead.")
 
     for i, y in enumerate(labels):
         idx = dist[i][::-1]
--- a/docs/source/cleanlab/datalab/guide/issue_type_description.rst
+++ b/docs/source/cleanlab/datalab/guide/issue_type_description.rst
@@ -141,6 +141,14 @@ To find the underperforming group, Cleanlab clusters the data using the provided
 The underperforming group quality score is equal to `q/r` for examples belonging to the underperforming group, and is equal to 1 for all other examples.
 Advanced users:  If you have pre-computed cluster assignments for each example in the dataset, you can pass them explicitly to :py:meth:`Datalab.find_issues <cleanlab.datalab.datalab.Datalab.find_issues>` using the `cluster_ids` key in the `issue_types` dict argument.  This is useful for tabular datasets where you want to group/slice the data based on a categorical column. An integer encoding of the categorical column can be passed as cluster assignments for finding the underperforming group, based on the data slices you define.
 
+Data Valuation Issue
+--------------------
+
+An Example that contribute minimally to a model's training have low data valuation scores.
+
+Data valuation issues can only use a `knn_graph`` passed by user or pre-computed from other issue managers. If you do not provide this argument, or there isn't a `knn_graph` already computed and store in the Datalab object, this type of issue will not be considered.
+
+The data valuation score is a approximate form of shapley value, which is calculated by the label of top k nearest neighbors of an example. The detail of the knn-shapley value could be found in the paper: `Efficient Task-Specific Data Valuation for Nearest Neighbor Algorithms <https://arxiv.org/abs/1908.08619>`_ and `Scalability vs. Utility: Do We Have to Sacrifice One for the Other in Data Importance Quantification? <https://arxiv.org/abs/1911.07128>`_.
 
 Optional Issue Parameters
 =========================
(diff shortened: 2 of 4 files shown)


docs/source/cleanlab/datalab/internal/issue_manager/data_valuation.rst
```diff
@@ -0,0 +1,9 @@
+data_valuation
+=======
```
jwmueller, line 2:
```suggestion
data_valuation
===========
```

cleanlab/datalab/internal/issue_manager/data_valuation.py
```diff
@@ -37,7 +37,22 @@
 
 
 class DataValuationIssueManager(IssueManager):
-    """Manages data sample with low valuation."""
+    """
+    Manages data sample with low valuation.
```
jwmueller, line 41:
```suggestion
    Detect which examples in a dataset are least valuable via an approximate Data Shapely value.
```

cleanlab/datalab/internal/issue_manager/data_valuation.py
```diff
@@ -37,7 +37,22 @@
...
+
```
jwmueller, line 42: Can we change in this file:

```
DEFAULT_THRESHOLDS = 0
```

I think that makes more sense than 1e-6. I.e. datapoints that contribute negative value are considered issues in the default setting.

docs/source/cleanlab/datalab/guide/issue_type_description.rst
```diff
@@ -141,6 +141,14 @@ To find the underperforming group, Cleanlab clusters the data using the provided
...
 
+Data Valuation Issue
+--------------------
+
+An Example that contribute minimally to a model's training have low data valuation scores.
+
+Data valuation issues can only use a `knn_graph`` passed by user or pre-computed from other issue managers. If you do not provide this argument, or there isn't a `knn_graph` already computed and store in the Datalab object, this type of issue will not be considered.
+
+The data valuation score is a approximate form of shapley value, which is calculated by the label of top k nearest neighbors of an example. The detail of the knn-shapley value could be found in the paper: `Efficient Task-Specific Data Valuation for Nearest Neighbor Algorithms <https://arxiv.org/abs/1908.08619>`_ and `Scalability vs. Utility: Do We Have to Sacrifice One for the Other in Data Importance Quantification? <https://arxiv.org/abs/1911.07128>`_.
```
jwmueller, line 151:
```suggestion
Data Valuation Issue
--------------------

The examples in the dataset with lowest data valuation scores contribute least to a trained ML model's performance (those whose value falls below a threshold are flagged with this type of issue).

Data valuation issues can only use a `knn_graph`` passed by user or pre-computed from other issue managers. If you do not provide this argument, or there isn't a `knn_graph` already computed and store in the Datalab object, this type of issue will not be considered.

The data valuation score is a approximate form of shapley value, which is calculated by the label of top k nearest neighbors of an example. The detail of the knn-shapley value could be found in the paper: `Efficient Task-Specific Data Valuation for Nearest Neighbor Algorithms <https://arxiv.org/abs/1908.08619>`_ and `Scalability vs. Utility: Do We Have to Sacrifice One for the Other in Data Importance Quantification? <https://arxiv.org/abs/1911.07128>`_.
```

docs/source/cleanlab/datalab/guide/issue_type_description.rst
```diff
@@ -283,3 +291,16 @@ Underperforming Group Issue Parameters
...
     For more information on generating `cluster_ids` for this issue manager, refer to this `FAQ Section <../../../tutorials/faq.html#How-do-I-specify-pre-computed-data-slices/clusters-when-detecting-the-Underperforming-Group-Issue?>`_.
+
+Data Valuation Issue Parameters
+--------------------------
+
+.. code-block:: python
+
+    data_valuation_kwargs = {
+        "k": # Number of nearest neighbors to calculate the data valuation score,
+        "thresholds": # Thresholds for determing if an example has a data valuation issue,
```
jwmueller, line 302:
```suggestion
        "k": # Number of nearest neighbors used to calculate data valuation scores,
        "threshold": # Examples with scores below this  threshold will be flagged with a data valuation issue
```

jwmueller: @coding-famer  Note I edited to `threshold` which I believe is the argument name

jwmueller (approved):
LGTM, up to my text suggestions. 

Leaving final code review + merge to @elisno

tests/datalab/issue_manager/test_data_valuation.py
```diff
@@ -74,3 +74,11 @@ def test_find_issue_wrong_knn_graph(self, issue_manager, embeddings):
                 features=embeddings["embedding"], issue_types={"outlier": {"k": 3}}
             )
             issue_manager.find_issues(k=4)
+
+    def test_get_larger_k_than_knn_graph(self, issue_manager, embeddings):
+        outlier_issue_manager = OutlierIssueManager(datalab=issue_manager.datalab, k=3)
```
tataganesh, line 79: Nit: Use the `outlier_issue_manager` fixture instead of creating a new `OutlierIssueManager` object.

cleanlab/datalab/internal/issue_manager/data_valuation.py
coding-famer: As discussed before, the data valuation score will be transformed to [0,1] to be the same as other issue scores. So I think 0 doesn't make sense. Maybe I can do more experiments to figure out a better threshold that can work for most datasets.

jwmueller: Ah yes, sorry I forgot about the transformation. Let's keep threshold as is. 

Can you add in the issue type description 1 sentence about how we mathematically transform the raw Data Shapely score to the interval 0-1?

Any additional experiments to help us find a good default threshold would be much appreciated!  Ideally our default threshold should capture most negatively-contributing and useless datapoints

coding-famer: OK, I will do some additional experiments to find a better default threshold later.

cleanlab/datalab/internal/issue_manager/data_valuation.py
```diff
@@ -55,7 +76,7 @@ class DataValuationIssueManager(IssueManager):
         3: ["average_data_valuation"],
     }
 
-    DEFAULT_THRESHOLDS = 1e-6
+    DEFAULT_THRESHOLD = 1e-6
```
jwmueller, line 79: Consider if default of 0.5 makes more sense here. 

If the original values are in [-1, 1] and mapped to [0, 1] via 0.5*x + 1 transform, then threshold of 0.5 would flag datapoints with negative original value (which seems natural to me)

coding-famer: Yes it makes sense, updated.

cleanlab/datalab/internal/issue_manager/data_valuation.py
elisno, line 1: Verified that this change doesn't change output (except that the change doesn't require reshaping the `labels`).
It also removes a Numpy DeprecationWarning.

> Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation.


Here's a code snippet making the comparison:

```python
from scipy.sparse import csr_matrix
import numpy as np

np.random.seed(0)

X = np.random.normal(0, 1, (100, 2))
y = np.random.choice([0, 1], size=100)
X[y == 0] += 2

knn = NearestNeighbors(n_neighbors=4, metric="euclidean").fit(X)
knn_graph = knn.kneighbors_graph(mode='distance')

def _knn_shapley_score(knn_graph: csr_matrix, labels: np.ndarray, k: int) -> np.ndarray:    
    """labels must be reshaped to (-1, 1)"""
    N = labels.shape[0]
    scores = np.zeros((N, N))
    dist = knn_graph.indices.reshape(N, -1)

    for i, y in enumerate(labels):
        idx = dist[i][::-1]
        ans = labels[idx]
        scores[idx[k - 1]][i] = float(ans[k - 1] == y) / k
        cur = k - 2
        for j in range(k - 1):
            scores[idx[cur]][i] = scores[idx[cur + 1]][i] + float(
                int(ans[cur] == y) - int(ans[cur + 1] == y)
            ) / k * (min(cur, k - 1) + 1) / (cur + 1)
            cur -= 1
    return 0.5 * (np.mean(scores, axis=1) + 1)

def _improved_implementation(knn_graph: csr_matrix, labels: np.ndarray, k: int) -> np.ndarray:
    """labels must be a 1D array"""
    N = labels.shape[0]
    scores = np.zeros((N, N))
    dist = knn_graph.indices.reshape(N, -1)

    for (y, s, dist_i) in zip(labels, scores, dist):
        idx = dist_i[::-1]
        ans = labels[idx]
        s[idx[k - 1]] = float(ans[k - 1] == y)
        ans_matches = (ans == y).flatten()
        for j in range(k - 2, -1, -1):
            s[idx[j]] = s[idx[j + 1]] + float(
                int(ans_matches[j]) - int(ans_matches[j + 1])
            )
    return 0.5 * (np.mean(scores / k, axis=0) + 1)

a = _knn_shapley_score(knn_graph, y.reshape(-1, 1), 4)
b = _improved_implementation(knn_graph, y, 4)

np.testing.assert_array_equal(a, b)
```

Benchmarking the runtime of the different functions (the line labeled as 1 is the improved implementation).

image

jwmueller: Awesome! looks good to merge to me (after merge conflict addressed).

elisno (approved):
LGTM. Awesome work @coding-famer!

PR #926, opened 2023-12-19
Minimal MNIST classifier

judeha:
Notebook that trains minimal CNN classifier with 1 resblock for basic MNIST. Attempted < 1000 parameters with >95% accuracy, achieved 1,051 parameters with 94% accuracy.


judeha:
meant to push to my local repo!

PR #934, opened 2023-12-27
class imbalance scores to still show rarest proportion even when is_i…

jwmueller:
Update class imbalance scores to still show rarest proportion even when is_issue = False.

I don't see a reason not to show these scores; the proportion of rarest class seems like useful info regardless of the threshold for determining whether the dataset has this issue.

### DIFF
--- a/tests/datalab/issue_manager/test_imbalance.py
+++ b/tests/datalab/issue_manager/test_imbalance.py
@@ -51,8 +51,10 @@ class TestClassImbalanceIssueManager:
         assert np.all(
             issues["is_class_imbalance_issue"] == np.full(N, False)
         ), "Issue mask should be correct"
-        assert np.min(issues["class_imbalance_score"]) == 0.47
-        assert np.max(issues["class_imbalance_score"]) == 1
+        scores = issues["class_imbalance_score"]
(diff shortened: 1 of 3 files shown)


tests/datalab/issue_manager/test_imbalance.py
```diff
@@ -51,7 +51,8 @@ def test_find_issues_no_imbalance(self, labels, create_issue_manager):
         assert np.all(
             issues["is_class_imbalance_issue"] == np.full(N, False)
         ), "Issue mask should be correct"
-        assert np.all(issues["class_imbalance_score"] == np.ones(N)), "Scores should be correct"
+        assert np.min(issues["class_imbalance_score"]) == 0.47
+        assert np.max(issues["class_imbalance_score"]) == 1
```
tataganesh, line 55: Suggesting a more specific test with error message
```suggestion
        scores = issues["class_imbalance_score"]
        expected_scores = np.ones_like(scores)
        expected_scores[labels == 1] = 0.47 # Rare class proportion
        np.testing.assert_allclose(scores, expected_scores, err_msg="Scores should be correct")
```

elisno (approved):

PR #942, opened 2024-01-04
add support for binary classifiers edge case

aditya1503:
## Summary

> 🎯 **Purpose**: Add support for binary classifiers with near 0.5 probability

### DIFF
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -151,9 +151,7 @@ def num_label_issues(
         label_issues_mask[cl_error_indices] = True
 
         # Remove label issues if given label == model prediction
-        mask = _reduce_issues(
-            pred_probs, labels, get_num_classes(pred_probs=pred_probs, labels=labels)
-        )
+        mask = _reduce_issues(pred_probs=pred_probs, labels=labels)
         label_issues_mask[mask] = False
         num_issues = np.sum(label_issues_mask)
     elif estimation_method == "off_diagonal_calibrated":
@@ -230,10 +228,9 @@ def _reduce_issues(pred_probs, labels, K, mask=None):
     pred = pred_probs.argmax(axis=1)
     if mask is None:
         mask = pred == labels
-    return mask
     if K == 2:
         # Set mask to False wherever pred_probs == 0.5 where num_classes == 2
-        mask = mask & ((pred_probs[:, 0] < 0.5 - EPSILON) & (pred_probs[:, 0] > 0.5 + EPSILON))
+        mask = mask & ((pred_probs[:, 0] < 0.5 - EPSILON) | (pred_probs[:, 0] > 0.5 + EPSILON))
     return mask
 
 
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -151,7 +151,11 @@ def num_label_issues(
         label_issues_mask[cl_error_indices] = True
 
         # Remove label issues if given label == model prediction
-        mask = _reduce_issues(pred_probs=pred_probs, labels=labels)
+        mask = _reduce_issues(
+            pred_probs=pred_probs,
+            labels=labels,
+            K=get_num_classes(pred_probs=pred_probs, labels=labels),
+        )
         label_issues_mask[mask] = False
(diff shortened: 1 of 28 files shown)


cleanlab/filter.py
```diff
@@ -446,7 +446,11 @@ class 0, 1, ..., K-1.
     if filter_by not in ["low_self_confidence", "low_normalized_margin"]:
         # Remove label issues if given label == model prediction if issues haven't been removed yet
         pred = pred_probs.argmax(axis=1)
-        label_issues_mask[pred == labels] = False
+        mask = pred == labels
+        if K == 2:
+            # Set mask to False wherever pred_probs == 0.5 where num_classes == 2
+            mask = mask & (pred_probs[:, 0] != 0.5)
```
aditya1503, line 452: Should this be between +/- (0.5 + TINY_VALUE)?

cleanlab/filter.py
```diff
@@ -446,7 +446,11 @@ class 0, 1, ..., K-1.
...
+        label_issues_mask[mask] = False
```
jwmueller, line 453:
```suggestion
        pred_confidence = pred_probs[:, pred]  # TODO: fix this with the correct syntax
        mask = mask & (pred_confidence < 1/K + EPS)
        label_issues_mask[mask] = False
```

jwmueller: @aditya1503 what about doing it like this ^ instead so it's more general solution?

cleanlab/filter.py
```diff
@@ -446,7 +446,11 @@ class 0, 1, ..., K-1.
...
```
jwmueller, line 454: Don't forget any updates to find_label_issues from now on must also be considered for find_label_issues_batched

cleanlab/filter.py
jwmueller: yes it should be:

pred_probs[:,given_label] > 0.5 - EPS

cleanlab/filter.py
jwmueller: Note this actually also happening in this method: https://github.com/cleanlab/cleanlab/blob/master/cleanlab/count.py#L153

I'm worried it is too unsafe to just do this update here, we are likely to miss other related methods where it needs to happen too

tests/test_multilabel_classification.py
```diff
@@ -17,23 +17,24 @@
...
+from cleanlab.internal import multilabel_scorer as ml_scorer
+from cleanlab.internal.multilabel_utils import get_onehot_num_classes, onehot2int, stack_complement
+from cleanlab.multilabel_classification import filter
 from cleanlab.multilabel_classification.dataset import (
     common_multilabel_issues,
-    rank_classes_by_multilabel_quality,
-    overall_multilabel_health_score,
     multilabel_health_summary,
+    overall_multilabel_health_score,
+    rank_classes_by_multilabel_quality,
```
aditya1503, line 35: These file changes are just iSort, no code changes here

cleanlab/filter.py
aditya1503: I've added support in all places I could find

cleanlab/count.py
```diff
@@ -224,6 +228,15 @@ def _num_label_issues_multilabel(
     return sum(issues_idx)
 
 
+def _reduce_issues(pred_probs, labels, num_classes):
+    pred = pred_probs.argmax(axis=1)
+    mask = pred == labels
+    if num_classes == 2:
+        # Set mask to False wherever pred_probs == 0.5 where num_classes == 2
+        mask = mask | ((pred_probs[:, 0] < 0.5 - EPSILON) | (pred_probs[:, 0] > 0.5 + EPSILON))
```
jwmueller, line 236: I think what I suggested would be cleaner code:

mask = pred_probs[:,given_label] > 0.5 - EPS

that in future could generalize to more classes beyond K = 2

jwmueller: even cleaner could be:

```
pred_probs[:,given_label] += EPS
pred = pred_probs.argmax(axis=1)
mask = pred == labels
```

PR #943, opened 2024-01-05
Near duplicate score transformation

elisno:
## Summary

> 🎯 **Purpose**: Standardize distance scoring function for near-duplicates.

In this PR, when computing near-duplicate scores in Datalab, we switch from tanh-transformed distances ($`x \rightarrow \tanh(x)`$) to an exponential-transformed distance ($`x \rightarrow 1 - \exp(-tx)`$) where $`t`$ is a temperature parameter.

image

image

### DIFF
--- a/cleanlab/datalab/internal/issue_manager/duplicate.py
+++ b/cleanlab/datalab/internal/issue_manager/duplicate.py
@@ -98,8 +98,12 @@ class NearDuplicateIssueManager(IssueManager):
             knn_graph = knn.kneighbors_graph(mode="distance")
         N = knn_graph.shape[0]
         nn_distances = knn_graph.data.reshape(N, -1)[:, 0]
-        median = max(np.median(nn_distances), np.finfo(np.float_).eps)  # avoid threshold = 0
-        self.near_duplicate_sets = self._neighbors_within_radius(knn_graph, self.threshold, median)
+        median_nn_distance = max(
+            np.median(nn_distances), np.finfo(np.float_).eps
+        )  # avoid threshold = 0
+        self.near_duplicate_sets = self._neighbors_within_radius(
+            knn_graph, self.threshold, median_nn_distance
+        )
 
         # Flag every example in a near-duplicate set as a near-duplicate issue
         all_near_duplicates = np.unique(np.concatenate(self.near_duplicate_sets))
@@ -114,7 +118,7 @@ class NearDuplicateIssueManager(IssueManager):
         )
 
         self.summary = self.make_summary(score=scores.mean())
-        self.info = self.collect_info(knn_graph=knn_graph)
+        self.info = self.collect_info(knn_graph=knn_graph, median_nn_distance=median_nn_distance)
 
     @staticmethod
     def _neighbors_within_radius(knn_graph: csr_matrix, threshold: float, median: float):
@@ -169,7 +173,7 @@ class NearDuplicateIssueManager(IssueManager):
             knn_graph = None
         return knn_graph
 
-    def collect_info(self, knn_graph: csr_matrix) -> dict:
+    def collect_info(self, knn_graph: csr_matrix, median_nn_distance: float) -> dict:
         issues_dict = {
             "average_near_duplicate_score": self.issues[self.issue_score_key].mean(),
             "near_duplicate_sets": self.near_duplicate_sets,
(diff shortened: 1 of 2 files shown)


cleanlab/datalab/internal/issue_manager/duplicate.py
```diff
@@ -98,13 +98,19 @@ def find_issues(
...
+        self.near_duplicate_sets = self._neighbors_within_radius(
+            knn_graph, self.threshold, median_nn_distance
+        )
 
         # Flag every example in a near-duplicate set as a near-duplicate issue
         all_near_duplicates = np.unique(np.concatenate(self.near_duplicate_sets))
         is_issue_column = np.zeros(N, dtype=bool)
         is_issue_column[all_near_duplicates] = True
+        temperature = 1.0 / median_nn_distance
+        scores = _compute_scores_with_exp_transform(nn_distances, temperature=temperature)
```
elisno, line 113: Scoring is now handled by an internal function.

cleanlab/datalab/internal/issue_manager/duplicate.py
```diff
@@ -233,3 +239,40 @@ def _set_threshold(
...
+        The nearest neighbor distances for each example.
+
+    Returns
+    -------
+    scores :
+        The near-duplicate scores for each example. The scores are in the range [0, 1].
+        A lower score indicates that an example is more likely to be a near-duplicate than
+        an example with a higher score.
+        A score of 0 indicates that an example has an exact duplicate.
+    """
```
elisno, line 273: Most of this docstring was written for the docs page, but it won't show up as it's an internal/private function.

It's fine to remove this docstring to reduce clutter.

tests/datalab/issue_manager/test_duplicate.py
```diff
@@ -72,7 +72,7 @@ def test_find_issues(self, issue_manager, embeddings):
             issues["is_near_duplicate_issue"] == expected_issue_mask
         ), "Issue mask should be correct"
         assert summary["issue_type"][0] == "near_duplicate"
-        assert summary["score"][0] == pytest.approx(expected=0.03122489, abs=1e-7)
+        assert summary["score"][0] == pytest.approx(expected=0.4734458, abs=1e-7)
```
elisno, line 75: The value changing here does not affect the detected outliers, just the interpretability of the score.

Changing the test value here to pass CI.

tests/datalab/issue_manager/test_duplicate.py
```diff
@@ -86,6 +86,17 @@ def test_find_issues(self, issue_manager, embeddings):
         )
         new_issue_manager.find_issues(features=embeddings["embedding"])
 
+    def test_scores_of_examples_with_issues_are_smaller_than_those_without(
+        self, issue_manager, embeddings
+    ):
+        # TODO: Turn this into a property-based test
```
elisno, line 92: This is more of a reminder to make this a more robust test.
It's not strictly related to the PR.
But since we're changing how the score looks, it may be a good idea to have this sanity-check.

I may open an issue to track this todo. It should be addressed in a follow-up PR.

cleanlab/datalab/internal/issue_manager/duplicate.py
```diff
@@ -98,13 +98,19 @@ def find_issues(
             knn_graph = knn.kneighbors_graph(mode="distance")
         N = knn_graph.shape[0]
         nn_distances = knn_graph.data.reshape(N, -1)[:, 0]
-        scores = np.tanh(nn_distances)
-        self.near_duplicate_sets = self._neighbors_within_radius(knn_graph, self.threshold)
+        median_nn_distance = max(
+            np.median(nn_distances), np.finfo(np.float_).eps
+        )  # avoid threshold = 0
```
ulya-tkch, line 103:
```suggestion
        )  # avoid zero median in denominator
```

no discussion of threshold here yet

elisno: That's from the previous PR. 

It's unrelated to the temperature being set.

elisno: It's actually still relevant to the `self.threshold`. Which is already set at this point in the method.

elisno: I'd keep it as is.

ulya-tkch (approved):
Matches what we discussed, just confirm if we should scale the temperature or the distances by the median (functionally identical but we might want a consistent definition of temperature)

tests/datalab/issue_manager/test_duplicate.py
jwmueller: opening GH issue sounds good!

PR #945, opened 2024-01-05
add property based test for multilabel classification 

aditya1503:
## Summary

> 🎯 **Purpose**: Add property based tests for Multi-label
he goal is to incorporate property-based tests for multi-label functionality. Specifically, this test case focuses on replicating a scenario where label == pred_label and is_issue is true during the data generation step. (where probabilities are close to 0.5) 

current code will fail until #950  is merged in
Screenshot 2024-01-09 at 12 53 04 PM

## To-Do

- [x] These tests will pass once https://github.com/cleanlab/cleanlab/pull/950 is merged into Master


tests/test_multilabel_classification.py
```diff
@@ -17,23 +17,27 @@
 
 import itertools
 import typing
+
```
jwmueller, line 20: [nit] group all hypothesis imports together

tests/test_multilabel_classification.py
```diff
@@ -718,3 +722,66 @@ def test_invalid_alpha(self):
...
+    def test_find_label_issues(self, data):
+        true_labels, noisy_labels, pred_probs = data
+        noisy_labels_list = onehot2int(noisy_labels)
+        is_issue = filter.find_label_issues(
+            labels=noisy_labels_list, pred_probs=np.array(pred_probs)
+        )
+        threshold = 0.5
+        pred_labels = (pred_probs > threshold).astype(int)
+        equal_pred = np.where(np.all(pred_labels == noisy_labels, axis=1), True, False)
+        assert sum(equal_pred & is_issue) == 0
```
jwmueller, line 787: can you also repeat this test (to ensure suggested != given label whenever is_issue=True) where you make some of the one-vs-rest pred_probs exactly equal to 0.5?

Say one binary classification problem, where you treat it as a multi-label problem and you set all pred_probs = 0.5

Say one 3-class multi-label problem where you set a few  pred_probs = [0.5, 0.25, 0.25]

In these cases, the labels can come from the same property-based test; doesn't matter too much where they come from.

tests/test_multilabel_classification.py
```diff
@@ -718,3 +722,66 @@ def test_invalid_alpha(self):
...
+                max_size=num_classes,
+            ),
+            min_size=num_samples,
+            max_size=num_samples,
+        )
+    )
+    return true_labels, noisy_labels, np.array(pred_probs)
+
+
+class TestCleanlab:
```
jwmueller, line 776: any reason for this particular naming?

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 # Python dependencies for development
 coverage != 6.3, != 6.3.*
 datasets
-hypothesis[numpy]
```
elisno, line 4: Why is the `numpy` extra being removed?

aditya1503: I thought hypotheses will cover everything including numpy, will add this as as separate comment

tests/test_multilabel_classification.py
aditya1503: I'll change

tests/test_multilabel_classification.py
aditya1503: ok, hypotheses covers then exact 0.5 case, but it's good to have one outside

jwmueller: I don't understand what you are saying here. I was requesting to add 2 additional tests. 

The current test looks to me like it's just using randomly generated pred_probs. If I'm misunderstanding that, please help me better understand it!

tests/test_multilabel_classification.py
```diff
@@ -718,3 +722,66 @@ def test_invalid_alpha(self):
...
+    @given(cleanlab_data_strategy())
+    def test_find_label_issues(self, data):
+        true_labels, noisy_labels, pred_probs = data
```
jwmueller, line 779:
```suggestion
        true_labels, noisy_labels, pred_probs = data
        pred_probs[0,0] = 0.5
```

jwmueller: how about just adding this to ensure at least one pred_prob is exactly 0.5

.github/workflows/ci.yml
```diff
@@ -85,7 +85,7 @@ jobs:
           pip install .
       - name: Install test dependencies
         run: |
-          pip install pytest pytest-lazy-fixture pipdeptree
+          pip install pytest pytest-lazy-fixture pipdeptree hypothesis
```
aditya1503, line 88: hypothesis is being added as a specific test dependency in the test-without-extras-min-versions

requirements-dev.txt
elisno: Oh, right, my mistake. That extra is already covered by`pip install hypothesis`.

tests/test_multilabel_classification.py
```diff
@@ -718,3 +722,71 @@ def test_invalid_alpha(self):
...
+                st.integers(min_value=0, max_value=1),
+                min_size=num_classes,
+                max_size=num_classes,
+            ),
+            min_size=num_samples,
+            max_size=num_samples,
+        )
+    )
+
+    # Generate noise matrix for multi-label
```
elisno, line 749: IIUC, we're flipping an entire one-hot encoded/binarized row this way?

```python
>>> import numpy as np
>>> def flip_labels(label, flip_prob):
...     return 1 - np.array(label) if np.random.rand() < flip_prob else label
... 
>>> true_labels = np.random.randint(0, 2, (6, 4))
>>> true_labels
# array([[0, 0, 0, 1],
#       [1, 1, 0, 1],
#       [0, 0, 0, 0],
#       [1, 1, 0, 0],
#       [1, 0, 0, 1],
#       [1, 0, 0, 1]])
>>> flip_prob = 0.2
>>> noisy_labels = np.array([flip_labels(label, flip_prob) for label in true_labels])
>>> noisy_labels
# array([[0, 0, 0, 1],
#       [1, 1, 0, 1],
#       [0, 0, 0, 0],
#       [1, 1, 0, 0],
#       [0, 1, 1, 0],   # This one is completely flipped
#       [0, 1, 1, 0]])  # This one also.
```

tests/test_multilabel_classification.py
```diff
@@ -718,3 +722,71 @@ def test_invalid_alpha(self):
...
+    flip_prob = 0.2
+    noisy_labels = np.array([flip_labels(label, flip_prob) for label in true_labels])
+
+    # Modify the noisy_labels using your logic
+    for i in range(num_samples):
+        for j in range(num_classes):
+            if draw(st.floats(min_value=0, max_value=1)) < 0.2:
+                noisy_labels[i][j] = 0
```
elisno, line 757: Could you tell me what this is supposed to do? Is the same `flip_prob` being applied here element-wise?

tests/test_multilabel_classification.py
```diff
@@ -718,3 +722,71 @@ def test_invalid_alpha(self):
...
+        st.lists(
+            st.lists(
+                st.integers(min_value=0, max_value=1),
+                min_size=num_classes,
+                max_size=num_classes,
+            ),
+            min_size=num_samples,
+            max_size=num_samples,
+        )
+    )
```
elisno, line 747: Prefer the [(numpy) `arrays` strategy](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.arrays) for generating the `true_labels`.

tests/test_multilabel_classification.py
```diff
@@ -718,3 +722,71 @@ def test_invalid_alpha(self):
...
+        st.lists(
+            st.lists(
+                st.floats(min_value=0, max_value=1),
+                min_size=num_classes,
+                max_size=num_classes,
+            ),
+            min_size=num_samples,
+            max_size=num_samples,
+        )
+    )
```
elisno, line 770: Also switch to the arrays strategy mentioned in https://github.com/cleanlab/cleanlab/pull/945/files#r1446033392

tests/test_multilabel_classification.py
```diff
@@ -718,3 +722,71 @@ def test_invalid_alpha(self):
...
+    @settings(deadline=10000)
+    def test_find_label_issues(self, data):
+        true_labels, noisy_labels, pred_probs = data
+        noisy_labels_list = onehot2int(noisy_labels)
+        is_issue = filter.find_label_issues(
+            labels=noisy_labels_list, pred_probs=np.array(pred_probs)
+        )
+        threshold = 0.5
+        pred_labels = (pred_probs >= threshold).astype(int)
+        equal_pred = np.where(np.all(pred_labels == noisy_labels, axis=1), True, False)
```
elisno, line 791: The `np.where(..., True, False)` is redundant.

```suggestion
        equal_pred = np.all(pred_labels == noisy_labels, axis=1)
```

elisno (commented):
Left some comments.

tests/test_multilabel_classification.py
aditya1503: I've changed the code

tests/test_multilabel_classification.py
aditya1503: It's random now (not across entire row), with the Docstring,

tests/test_multilabel_classification.py
```diff
@@ -718,3 +723,61 @@ def test_invalid_alpha(self):
...
+            labels=noisy_labels_list, pred_probs=np.array(pred_probs)
+        )
+        threshold = 0.5
+        predicted_labels = (pred_probs >= threshold).astype(int)
+
+        # Check if predicted labels are the same as noisy labels for each example
+        labels_match = np.all(predicted_labels == noisy_labels, axis=1)
+        
+        # For any example flagged as having an issue, there should be at least one label mismatch
+        assert not np.any(labels_match & is_issue), "Examples with issues must have at least one label mismatch."
```
elisno, line 783: Switch operand order for readability and consistency
```suggestion
        assert not np.any(is_issue & labels_match), "Examples with issues must have at least one label mismatch."
```

jwmueller (approved):
LGTM. Just make the code pass the Format check

elisno (approved):

PR #947, opened 2024-01-07
Add identifier issue_manager

01PrathamS:
## Summary

> 🎯 **Purpose**:  Add new issue manager like ```identifier_column``` that checks whether features contains a column that is entirely sequential integers.

**[ ✏️this issue manager checks if data has identifier column if yes than it returns 0  else return None ]**

```python
X = np.array([[0, 1], [1, 1], [2, 2], [3, 0]])
y = np.array([0, 1, 1, 0])
lab = Datalab(data={"X": X, "y": y}, label_name="y")
lab.find_issues(features=X, issue_types={"identifier":{}})
```

### DIFF
--- a/cleanlab/datalab/internal/issue_manager/identifier.py
+++ b/cleanlab/datalab/internal/issue_manager/identifier.py
@@ -1,0 +1,39 @@
+import numpy as np
+from typing import Tuple, Optional
+
+from cleanlab.datalab.internal.issue_manager import IssueManager
+
+
+class IdentifierIssueManager(IssueManager):
+    """An issue manager that keeps track of issues related to identifier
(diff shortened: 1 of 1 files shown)


MaxJoas:
I was wondering if we should have a minimum requirement for sample_size for the `_is_sequential` check. For a low number of samples it could have a sequential column by change. Should we think about this edge case?

jwmueller:
> I was wondering if we should have a minimum requirement for sample_size for the `_is_sequential` check. For a low number of samples it could have a sequential column by change. Should we think about this edge case?

Good question! But I think it's ok to still raise an alert about that even when the number of samples is low. Somebody can disregard the alert

PR #950, opened 2024-01-08
ensure near-predicted labels are not flagged as label issue

aditya1503:
## Summary

> 🎯 **Purpose**: Add support for binary classifier where pred probs are close to 0.5

Ignore cases where for a binary classifier, the probability is very close to 0.5 for both classes, and avoid that being flagged as a label issue.

### DIFF
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -229,6 +229,7 @@ def _num_label_issues_multilabel(
 
 
 def _reduce_issues(pred_probs, labels, num_classes):
+    """Returns a boolean mask denoting correct predictions or predictions within a margin around 0.5 for binary classification, suitable for filtering out indices in 'is_label_issue'."""
     pred = pred_probs.argmax(axis=1)
     mask = pred == labels
     if num_classes == 2:
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -155,8 +155,8 @@ def num_label_issues(
         label_issues_mask[cl_error_indices] = True
 
         # Remove label issues if model prediction is close to given label
-        mask = _reduce_issues(pred_probs=pred_probs, labels=labels)
-        label_issues_mask[mask] = False
+        mask = _reduce_issues(pred_probs=pred_probs[cl_error_indices], labels=labels[cl_error_indices])
+        label_issues_mask[cl_error_indices[mask]] = False
         num_issues = np.sum(label_issues_mask)
     elif estimation_method == "off_diagonal_calibrated":
         calculated_confident_joint = compute_confident_joint(
--- a/cleanlab/count.py
+++ b/cleanlab/count.py
@@ -38,7 +38,7 @@ from typing import Tuple, Union, Optional
 
 from cleanlab.typing import LabelLike
 from cleanlab.internal.multilabel_utils import stack_complement, get_onehot_num_classes
-from cleanlab.internal.constants import TINY_VALUE, CONFIDENT_THRESHOLDS_LOWER_BOUND
+from cleanlab.internal.constants import TINY_VALUE, CONFIDENT_THRESHOLDS_LOWER_BOUND, EPSILON
 
 from cleanlab.internal.util import (
     value_counts_fill_missing_classes,
@@ -151,8 +151,12 @@ def num_label_issues(
         label_issues_mask[cl_error_indices] = True
 
         # Remove label issues if given label == model prediction
-        pred = pred_probs.argmax(axis=1)
-        label_issues_mask[pred == labels] = False
+        mask = _reduce_issues(
+            pred_probs=pred_probs,
+            labels=labels,
+            num_classes=get_num_classes(pred_probs=pred_probs, labels=labels),
+        )
+        label_issues_mask[mask] = False
         num_issues = np.sum(label_issues_mask)
     elif estimation_method == "off_diagonal_calibrated":
         calculated_confident_joint = compute_confident_joint(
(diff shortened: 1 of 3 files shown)


cleanlab/count.py
```diff
@@ -224,6 +228,15 @@ def _num_label_issues_multilabel(
     return sum(issues_idx)
 
 
+def _reduce_issues(pred_probs, labels, num_classes):
+    """Returns a boolean mask denoting correct predictions or predictions within a margin around 0.5 for binary classification, suitable for filtering out indices in 'is_label_issue'."""
+    pred = pred_probs.argmax(axis=1)
+    mask = pred == labels
+    if num_classes == 2:
+        mask = mask | ((pred_probs[:, 0] >= 0.5 - EPSILON) & (pred_probs[:, 0] <= 0.5 + EPSILON))
```
jwmueller, line 236: After thinking more, I think this would be cleaner:

```
pred_probs[:,given_label] += EPS
pred = pred_probs.argmax(axis=1)
mask = pred == labels
```

with this code here, we can get rid of the `if num_classes == 2` statement IMO.

Note that my above code is just pseudocode but you get the idea

jwmueller: More formally I believe the line above should be:

```
pred_probs[np.arange(len(labels)), labels] += FLOATING_POINT_COMPARISON
```

aditya1503: Added here: https://github.com/cleanlab/cleanlab/blob/309ff7de153573ca030e3c257ded916386e709a6/cleanlab/count.py#L234

cleanlab/filter.py
```diff
@@ -445,8 +445,8 @@ class 0, 1, ..., K-1.
 
     if filter_by not in ["low_self_confidence", "low_normalized_margin"]:
         # Remove label issues if given label == model prediction if issues haven't been removed yet
```
jwmueller, line 447:
```suggestion
        # Remove label issues if model prediction is close to given label
```

jwmueller (approved):
LGTM. Just had a minor suggestions to make the comments more precise. 

Passing to @elisno for final review+merge. In particular, see if you have an idea to make the `_reduce_issues()` more efficient (the copy operation is maybe unnecessary in a different implementation)

cleanlab/count.py
```diff
@@ -150,9 +154,9 @@ def num_label_issues(
         label_issues_mask = np.zeros(len(labels), dtype=bool)
         label_issues_mask[cl_error_indices] = True
 
-        # Remove label issues if given label == model prediction
-        pred = pred_probs.argmax(axis=1)
-        label_issues_mask[pred == labels] = False
+        # Remove label issues if model prediction is close to given label
+        mask = _reduce_issues(pred_probs=pred_probs, labels=labels)
+        label_issues_mask[mask] = False
```
elisno, line 159: Since the `label_issues_mask` is initialized with `False`s, you only need to reset to `False` at the relevant `cl_error_indices`.

```suggestion
        mask = _reduce_issues(pred_probs=pred_probs[cl_error_indices], labels=labels[cl_error_indices])
        label_issues_mask[cl_error_indices[mask]] = False
```

cleanlab/filter.py
```diff
@@ -444,9 +444,9 @@ class 0, 1, ..., K-1.
         label_issues_mask = find_predicted_neq_given(labels, pred_probs, multi_label=multi_label)
 
     if filter_by not in ["low_self_confidence", "low_normalized_margin"]:
-        # Remove label issues if given label == model prediction if issues haven't been removed yet
-        pred = pred_probs.argmax(axis=1)
-        label_issues_mask[pred == labels] = False
+        # Remove label issues if model prediction is close to given label
+        mask = _reduce_issues(pred_probs=pred_probs, labels=labels)
+        label_issues_mask[mask] = False
```
elisno, line 449:
```suggestion
        mask = _reduce_issues(pred_probs=pred_probs[cl_error_indices], labels=labels[cl_error_indices])
        label_issues_mask[cl_error_indices[mask]] = False
```

elisno:
> Passing to @elisno for final review+merge. In particular, see if you have an idea to make the `_reduce_issues()` more efficient (the copy operation is maybe unnecessary in a different implementation)

The function will have to do a copy for now. However, it only needs to be applied to a subset of the dataset.
The relevant examples affected by _reduce_issues are those at `cl_issue_indices`. I assume that's usually a small fraction of the dataset.

cleanlab/experimental/label_issues_batched.py
```diff
@@ -234,11 +234,13 @@ def find_label_issues_batched(
         pbar.close()
 
     label_issues_indices = lab.get_label_issues()
+    label_issues_mask = np.zeros(len(labels), dtype=bool)
+    label_issues_mask[label_issues_indices] = True
+    mask = _reduce_issues(pred_probs=pred_probs, labels=labels)
+    label_issues_mask[mask] = False
```
elisno, line 240:
```suggestion
    mask = _reduce_issues(pred_probs=pred_probs[label_issues_indices], labels=labels[label_issues_indices])
    label_issues_mask[label_issues_indices[mask]] = False
```

cleanlab/filter.py
elisno: On second thought. The variable `cl_error_indices` is not always defined at this point, so the mask needs to be applied to the full set of `pred_probs` and `labels`. 🤔

jwmueller:
> > Passing to @elisno for final review+merge. In particular, see if you have an idea to make the `_reduce_issues()` more efficient (the copy operation is maybe unnecessary in a different implementation)
> 
> The function will have to do a copy for now. However, it only needs to be applied to a subset of the dataset. The relevant examples affected by _reduce_issues are those at `cl_issue_indices`. I assume that's usually a small fraction of the dataset.
> 
> EDIT: Unfortunately, `cl_issue_indices` is not accessible in all the places suggested.

Thanks for your clever changes! Seems good to merge to me

PR #953, opened 2024-01-11
Rescale ood scores in outlier package

elisno:


elisno:
The failing link-check is unrelated and has already been fixed on the master branch.

jwmueller:
ask Ulyna for review when ready, thanks!

jwmueller:
Thanks! @elisno Can you confirm you ran the tutorial notebook:
https://docs.cleanlab.ai/master/tutorials/outliers.html

And the new scores look good for both train and for test data?  (ie. specifically the test data scores look like they're in similar ranges as the train data scores as opposed to being in a completely different range)

elisno:
> Thanks! @elisno Can you confirm you ran the tutorial notebook: https://docs.cleanlab.ai/master/tutorials/outliers.html
> 
> And the new scores look good for both train and for test data? (ie. specifically the test data scores look like they're in similar ranges as the train data scores as opposed to being in a completely different range)

I can confirm.

The distribution has shifted, but looks similar.

image
image

Anyone interested can take a look at the different scores in [this gist](https://gist.github.com/elisno/b3aa29ca423d452c3b61a94067d46da5) (two files: train_scores.csv and test_scores.csv).
The PCC between the different scores is very high.

```
PearsonRResult(statistic=0.9963142580354383, pvalue=0.0) # train
PearsonRResult(statistic=0.9962786900544783, pvalue=0.0) # test
```

cleanlab/datalab/internal/issue_manager/outlier.py
```diff
@@ -103,7 +103,11 @@ def find_issues(
             t = cast(int, self.ood.params["t"])
             distances = knn_graph.data.reshape(-1, k)
             assert isinstance(distances, np.ndarray)
-            scores = transform_distances_to_scores(distances, k=k, t=t)
+            avg_distances = distances.mean(axis=1)
+            median_distance = np.median(avg_distances)
```
ulya-tkch, line 107: [nit] This would be `median_avg_distance`

cleanlab/datalab/internal/issue_manager/regression/label.py
```diff
@@ -64,7 +64,7 @@ def __init__(
         self,
         datalab: Datalab,
         clean_learning_kwargs: Optional[Dict[str, Any]] = None,
-        threshold: float = 0.1,
+        threshold: float = 0.05,
```
ulya-tkch, line 67: Why is this change here? If default is now 0.05, update the docstring for `RegressionLabelIssueManager` above with new value^

cleanlab/internal/outlier.py
```diff
@@ -33,18 +35,17 @@ def transform_distances_to_scores(distances: np.ndarray, k: int, t: int) -> np.n
...
-        Each row contains the distances to each example's `num_neighbors` nearest neighbors.
-        It is assumed that each row is sorted in ascending order.
-
-    k : int
-        Number of neighbors used to compute the average distance to each example.
-        This assumes that the second dimension of distances is k or greater, but it
-        uses slicing to avoid indexing errors.
+    avg_distances : np.ndarray
+        An array of distances of shape ``(N)``, where N is the number of examples.
+        Each row represents an example's average distance to its k nearest neighbors.
```
ulya-tkch, line 40: Shouldn't it be "each example represents .." since array is only a single dimension (N,)

cleanlab/internal/outlier.py
```diff
@@ -33,18 +35,17 @@ def transform_distances_to_scores(distances: np.ndarray, k: int, t: int) -> np.n
...
 
     t : int
-        Controls transformation of distances between examples into similarity scores that lie in [0,1].
+        A sensitivity parameter that modulates the strength of the transformation from distances to scores.
+        Higher values of `t` result in more pronounced differentiation between the scores of examples
+        lying in the range [0,1].
+
+    scaling_factor : float
```
ulya-tkch, line 47: What is the range for this/what should it be ideally? What is the repercussion of making this value larger or smaller?

ulya-tkch (approved):
Changes lgtm outside of few comments.

cleanlab/internal/outlier.py
jwmueller: 'entry' not 'example'

cleanlab/datalab/internal/issue_manager/outlier.py
jwmueller: I also think the current variable name is confusing and prefer `median_avg_distance`

cleanlab/datalab/internal/issue_manager/regression/label.py
elisno: The label quality score for regression is based on the [outer score](https://docs.cleanlab.ai/stable/cleanlab/regression/rank.html?highlight=outre#cleanlab.regression.rank.get_label_quality_scores) (basically the feature-based OOD score from `OutOfDistrubtion`).

The way label issues are flagged in this case is:

```python
label_quality_score: np.ndarray
is_label_issue = label_quality_score < threshold * np.median(label_quality_score)
```

The distribution of the `label_quality_score`  is affected by this PR. I've noticed that the normalization step causes the label quality scores to spread out more than the previous score.
A lower threshold is necessary for a similar "rate" of expected issues.

cleanlab/internal/outlier.py
elisno: A valid scaling factor lies in the range $(0,  +\infty)$, but ideally, it should be selected based on some reference distance measured in the dataset. By default, its value is set to the median of the average distances to the k-nearest neighbors. The choice of scaling factor is flexible.

- The value can be increased (_decreased_) to dilute (_amplify_) the relative differences in distances between examples (the distribution of scores will be shifted towards 1 (_0_)).

PR #958, opened 2024-01-17
Add missing ref to url links in OpenGraph metadata

elisno:
The opengraph extension for sphinx only links to `https://docs.cleanlab.ai/[path/to/page].html`, but should be `https://docs.cleanlab.ai/[ref_name]/[path/to/page].html`.

where `ref_name` should be either `stable`, master or a released version (e.g. v.2.6.0 in the future).

Without this fix, all the links will fail a link-check.

This PR aims to fix that issue.


elisno:
Merging to test if deployed docs for the master branch work.

Issue #909, opened 2023-12-07
update regression tutorial to use datalab regression


jwmueller:
https://docs.cleanlab.ai/master/tutorials/regression.html

Just add a prior section doing the issue detection via Datalab first, before the current cleanlab code in this tutorial. Keep rest of tutorial mostly as is, so at the end of tutorial, should still showcase the alternative use of CleanLearning and regression.get_label_quality_scores. Make it clear these final sections are specifically focused on label issues only.

PR #966, opened 2024-01-24
Update regression tutorial to use Datalab

elisno:
The markdown cells have not been update at this point.

I've double-checked that the printed issue indices in the previous version are reproduced with this version.

The only difference I see is that the label quality scores look different, but are just rescaled due to the outlier-score rescaling PR (examples are ranked the same, but the scores should be more "human-readable").


jwmueller:
Closes https://github.com/cleanlab/cleanlab/issues/909

Addresses question asked here:  https://github.com/cleanlab/cleanlab/issues/967

elisno:
Work moved to #992

PR #968, opened 2024-01-25
add bbox area computation

aditya1503:
Adding area computation function for bounding boxes, Numpy implementation


jwmueller (approved):
LGTM!

Issue #967, opened 2024-01-25
Datalab equivalent for Regression?


tqdo:
I wonder whether Cleanlab has something for regression problems that is equivalent to Datalab? Datalab's find_issues function requires `pred_probs`, which is not applicable for regression problems. I really like Datalab because it shows all key issues conveniently. Thanks

jwmueller:
Hi @tqdo , yes doing regression tasks with Datalab already exists in the latest developer version.

```
pip install "git+https://github.com/cleanlab/cleanlab.git#egg=cleanlab[all]"
```

https://docs.cleanlab.ai/master/cleanlab/datalab/datalab.html
Note the `task` argument in Datalab (you can set it to 'regression')

We have a tutorial for using this in the works here:
https://github.com/cleanlab/cleanlab/pull/966

Would appreciate your feedback!

jwmueller:
the tutorial is currently in the works not yet published. You can follow along with our progress in the linked PR. 

More precisely, the current working version of it is here (which you should be able to run today as long as you install the developer version of cleanlab):

https://github.com/cleanlab/cleanlab/blob/6811d14acb849a99699e3ff5e7f90967d8747bc2/docs/source/tutorials/regression.ipynb

jwmueller:
actually all of those capabilities are already available! That's why we extended Datalab to work for multi-label and regression tasks. 

This is not demonstrated in the current tutorials yet, but after you run `datalab.find_issues(...)` for a regression or [multi-label](https://docs.cleanlab.ai/master/tutorials/multilabel_classification.html) task, it will detect all of the types of data issues featured in our datalab quickstart tutorial:

https://docs.cleanlab.ai/master/tutorials/datalab/datalab_quickstart.html

elisno:
Hi @tqdo!
Thank you for pointing this out!

At the moment, on the latest developer version, `Datalab(..., task="regression")` is only configured to find label issues (specifically for this task).

I've got a PR in the making that should open this weekend. That includes the issues checks for a regression task:
- Outlier detection
- (near) duplicate detection
- Non-IID check
- Null-value check.

I'll ping you here again when you can try it out again in Colab.

jwmueller:
Hi @tqdo, sorry I gave you the pointer a bit too early, but now in the latest release: cleanlab v2.6.0, the original tip will now do what you wanted:

[...]

PR #979, opened 2024-02-06
Add existing issue managers to more tasks

elisno:
## Summary

Allow Datalab for other tasks than classification to look for feature/knn_graph-based issue types.

  - The existing issue managers are included in the registry for other tasks.
  - The required arguments for these issue managers are the same as for classification, but omit `pred_probs` from that list.
    - The exception to that rule is that the label issue managers all have task-specific implementations supporting task-specific inputs (`pred_probs` for classification, `predictions` for regression, etc.).

Also, end-to-end tests for task-specific Datalab have been moved to their own modules instead of being kept all in test_datalab.py.

  - For now, these check that, by default, the expected issue managers are run.
  - The test fixtures for find_issues have also been parametrized in the tests, so we know what happens when only `pred_probs` is passed to Datlab, vs only `features, etc.

## Testing
Most of the tests `tests/datalab/test_datalab.py` were moved to a new subdirectory.

They are now in `tests/datalab/datalab/test_datalab.py`, similar to the `cleanlab/datalab` directory structure.

**Unaddressed Cases**

Some TODO items were added in  `tests/datalab/test_datalab.py/test_multilabel.py` for adding support and tests for running label issue checks with features or knn_graphs for multilabel tasks.

- These issue managers weren't executable via tasks other than classifications, as mentioned in #967.


cleanlab/datalab/internal/issue_finder.py
```diff
@@ -435,7 +454,11 @@ def get_available_issue_types(self, **kwargs):
             warnings.warn("No labels were provided. " "The 'label' issue type will not be run.")
             issue_types_copy.pop("label")
 
-        outlier_check_needs_features = "outlier" in issue_types_copy and not self.datalab.has_labels
+        outlier_check_needs_features = (
+            "outlier" in issue_types_copy
+            and not self.datalab.has_labels
+            and self.task != "regression"
```
sanjanag, line 460:
```suggestion
        outlier_check_needs_features = (
            self.task == "classification"
            and "outlier" in issue_types_copy
            and not self.datalab.has_labels
```

sanjanag (approved):

PR #1000, opened 2024-02-13
Skip class imbalance issue check if labels are unavailable

### LINKED ISSUE #988: Class Imbalance issue checker should not run if labels are not provided in Datalab
```
from cleanlab import Datalab

lab = Datalab(data=df_without_labels)
lab.find_issues()
```

It should not run the ClassImbalanceIssueManager, but it tries to anyway.

Just add a check that the Datlab had labels specified, then it can run the ClassImbalanceIssueManager in find_issues.

elisno:
Fixes #988


elisno:
This is a trivial fix. CI failing is unrelated to this PR.

PR #1023, opened 2024-02-21
Support features as input for data valuation check in Datalab

### LINKED ISSUE #928: allow Data Valuation Issue Manager to operate on provided features
It currently only operates on `knn_graph`. Should simply construct the KNN graph based on the features if one does not already exist in the Datalab object. This is what the Outlier Issue Manager does.

elisno:
## Summary

Let `DataValuationIssueManager` accept `features` as input to construct a `knn_graph` for estimating Data Shapely 
 values of every data point in the dataset.

```python
# Example code snippet
import numpy as np
from cleanlab import Datalab

X, y = np.random.rand(20, 2), np.random.randint(0,2,20)
lab = Datalab(data={"X": X, "y": y}, label_name="y")

# Passing in `features` should work with this change
lab.find_issues(features=X, issue_types={"data_valuation": {}})

# data-valuation results should appear in the report like so
lab.report(show_all_issues = True)
```

Closes #928


sanjanag (approved):

jwmueller (approved):
Thanks for the quick PR!

PR #1024, opened 2024-02-22
List out issue names for each issue type in Datalab Issue Type Guide

elisno:
## Summary

Introduce a new `tip` admonition template to be utilized for each section within the guide. The template's appearance within the documentation resembles the following:

image

## High-level technical changes
Within conf.py, extract and organize the names of all registered issue managers into individual `<ISSUE_NAME>_context` variables for Jinja templating purposes. Currently, each context solely retains the `issue name`:

For example:

- `label_context` contains `{"issue_name": "label"}` within its context.
- `outlier_context` contains `{"issue_name": "outlier"}` within its context.
- and so forth.

The intention is to enrich these contexts with additional valuable information (such as the required_args in `Datalab.find_issues()`). While this enhancement may seem minimal at present, it lays the groundwork for future work, including the potential execution of automated tests that cater to default issue types.

## Reviewer notes

Take a closer look at the new template in `docs/source/cleanlab/datalab/guide/_templates/issue_types_tip.rst`.
- Is a `tip` admonition the best way to go here? 
- Can the text be shorter? The template size should be minimal, as it is used in about 8 sections on the same page.

### DIFF
--- a/docs/source/cleanlab/datalab/guide/issue_type_description.rst
+++ b/docs/source/cleanlab/datalab/guide/issue_type_description.rst
@@ -45,8 +45,8 @@ Label Issue
 Examples whose given label is estimated to be potentially incorrect (e.g. due to annotation error) are flagged as having label issues.
 Datalab estimates which examples appear mislabeled as well as a numeric label quality score for each, which quantifies the likelihood that an example is correctly labeled.
 
-For now, Datalab can only detect label issues in a multi-class classification dataset.
-The cleanlab library has alternative methods you can us to detect label issues in other types of datasets (multi-label, multi-annotator, token classification, etc.).
+For now, Datalab can only detect label issues in multi-class classification datasets, regression datasets, and multi-label classification datasets.
+The cleanlab library has alternative methods you can us to detect label issues in other types of datasets (multi-annotator, token classification, etc.).
 
 Label issues are calculated based on provided `pred_probs` from a trained model. If you do not provide this argument, but you do provide `features`, then a K Nearest Neighbor model will be fit to produce `pred_probs` based on your `features`. Otherwise if neither `pred_probs` nor `features` is provided, then this type of issue will not be considered.
 For the most accurate results, provide out-of-sample `pred_probs` which can be obtained for a dataset via `cross-validation <https://docs.cleanlab.ai/stable/tutorials/pred_probs_cross_val.html>`_.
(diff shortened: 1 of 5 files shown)


docs/source/cleanlab/datalab/guide/issue_type_description.rst
```diff
@@ -38,11 +38,15 @@ Each input is optional, if you do not provide it, Datalab will skip checks for t
 Label Issue
 -----------
 
+.. jinja:: label_context
+    :file:  cleanlab/datalab/guide/_templates/issue_types_tip.rst
```
jwmueller, line 42: can we move this to the end of this Label Issue section (right before Outlier Issue section begins below). That way these tips appear after the description of the issue which makes more sense to me. 

(Same idea would apply for all issue types)

jwmueller:
how will somebody contributing a new issue type remember to add this tip? 
Maybe briefly mention it (without specifics since those might change) here:  https://github.com/cleanlab/cleanlab/blob/master/DEVELOPMENT.md#contributing-new-issue-types-to-datalab

jwmueller:
my only other concern is that this is advanced jinja functionality that may only work with limited versions of jinja (but I have no clue), and potentially make docs builds more fragile. 

But I assume you've tested that and everything seems stable, in which case I'm fine with this update.

docs/source/cleanlab/datalab/guide/issue_type_description.rst
elisno: Moved all tips to the end of their respective sections.

elisno:
> my only other concern is that this is advanced jinja functionality that may only work with limited versions of jinja (but I have no clue), and potentially make docs builds more fragile.
> 
> But I assume you've tested that and everything seems stable, in which case I'm fine with this update.

I've made the usage more straightforward by just setting the context variable in the rst file (not in conf.py, which will be hard to keep track of). But I still want to use sphinx_jinja to be able to apply jinja directly to the rst file.
We may wish to remove that dependency and write the extension ourselves (like is done [here](https://ericholscher.com/blog/2016/jul/25/integrating-jinja-rst-sphinx/))


### From the linked post

> Here is the full extension: 
>```python
> def rstjinja(app, docname, source):
>     """
>     Render our pages as a jinja template for fancy templating goodness.
>     """
>     # Make sure we're outputting HTML
>     if app.builder.format != 'html':
>         return
>     src = source[0]
>     rendered = app.builder.templates.render_string(
>         src, app.config.html_context
>     )
>     source[0] = rendered
> 
> def setup(app):
>     app.connect("source-read", rstjinja)
> ```


Then we can use the `html_context` variable in `conf.py` to also provide the context to our rst files.

jwmueller (approved):

PR #1025, opened 2024-02-24
Fix error handling in get_issues method with empty issue dataframes

### LINKED ISSUE #986: Calling Datalab.get_issues() when there are no issues throws an error
When finding issues using code below (on a dataset where there are no issues), the following attempts (1,2) throw error:

```python
lab = cleanlab.Datalab(data=train[numerical], label_name="label", task="regression")
lab.find_issues(pred_probs=y_train_pred)
```
1.`label_quality_scores = lab.get_issues('label')['label_issues']`
2.`label_quality_scores = lab.get_issues('label')`

Instead should print something like "There are no isses"

elisno:
## Summary

The objective of this PR is to improve user guidance and error message clarity when issue retrieval fails when calling `Datalab.get_issues()`

It addresses the need for users to understand why retrieving issues might fail, specifically focusing on scenarios where `Datalab.find_issues()` has not been executed or has not included the requested issue type in its checks.

**Changes**
- Updated the `DataIssue.get_issues()` method to provide detailed error messages for an empty issue set or when no columns match a specified issue type, guiding users toward resolution.
  - It's not easy at call time to find out what step the user should pick.
- Added test cases in TestDatalabGetIssuesMethod to validate these error messages under various scenarios. Tagged with issue number that this PR is intended to resolve.

Fixes #986, enhancing error communication in the get_issues method.

### DIFF
--- a/cleanlab/datalab/datalab.py
+++ b/cleanlab/datalab/datalab.py
@@ -513,6 +513,15 @@ class Datalab:
 
             Additional columns may be present in the DataFrame depending on the type of issue specified.
         """
+
+        # Validate issue_name
+        if issue_name is not None and issue_name not in self.list_possible_issue_types():
+            raise ValueError(
+                f"""Invalid issue_name: {issue_name}. Please specify a valid issue_name from the list of possible issue types.
+            Either, specify one of the following: {self.list_possible_issue_types()}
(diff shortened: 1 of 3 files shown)


elisno:
Failing CI check is unrelated.

PR #1026, opened 2024-02-24
Verify get_issues("label") method works when no label issues are found

### LINKED ISSUE #986: Calling Datalab.get_issues() when there are no issues throws an error

elisno:
This pull request fixes an issue with the get_issues method in the Datalab class. Previously, if no label issues were found, the method would raise an error. This PR adds a check to handle this scenario and ensures that the method returns an empty DataFrame when no issues are found.

It is a quick follow-up with #1025, confirming it closes #986.

### DIFF
--- a/tests/datalab/datalab/test_datalab.py
+++ b/tests/datalab/datalab/test_datalab.py
@@ -1756,3 +1756,28 @@ class TestDatalabGetIssuesMethod:
         invalid_issue_name = "nul"
         with pytest.raises(ValueError, match=f"Invalid issue_name: {invalid_issue_name}."):
             lab.get_issues(issue_name=invalid_issue_name)
+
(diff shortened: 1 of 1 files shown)


jwmueller (approved):

elisno:
CI failing is unrelated to this PR (jax isn't used in this package).

I will follow up in a separate PR if this persists (I may look at pinning the tensorflow version differently if that is possible).

PR #1029, opened 2024-02-26
Fixed CleanLearning instantiation to use correct model

mturk24:
## Summary
Fixed a line in in-depth CLOS tutorial to use the model instantiated in previous line


elisno (approved):
Nice catch! Thank you @mturk24!

mturk24:
> Nice catch! Thank you @mturk24!

No prob :)

PR #1032, opened 2024-02-27
Transform regression label quality scores 

huiwengoh:
## Summary

Transform the `label_quality_scores` in `cleanlab.regression.CleanLearning` by its median so that it is more human readable.


jwmueller (approved):
lgtm

jwmueller:
@elisno make sure to specifically highlight the changing range of these scores in the next release, just like we did for the outlier scores

elisno (approved):
LGTM!

Issue #973, opened 2024-01-30
create user-facing cleanlab.data_valuation module


jwmueller:
Move the following method to the cleanlab.data_valuation module, such that it can be a standalone user-facing method:

[cleanlab.datalab.internal.issue_maanger.data_valuation._knn_shapley_score
](https://docs.cleanlab.ai/master/_modules/cleanlab/datalab/internal/issue_manager/data_valuation.html#DataValuationIssueManager.find_issues)

01PrathamS:
To clarify, are we currently in the process of moving the private method ```_knn_shapely_score``` to the ```DataValuationIssueManager``` class to make it a standalone user-facing method under the ```data_valuation.py``` module?

jwmueller:
@01PrathamS The goal of this GH issue is to make this move. 

I'd name the public method `cleanlab.data_valuation.data_shapely(labels, features: None, knn_graph: None, k=10)`

and internally the method should look something like:

```
if knn_graph is not None and features is not None:
    raise ValueError(...)

if knn_graph is None:
    knn_graph = reuse existing helper methods in datalab to build KNN graph from the features vector

return _knn_shapley_score(knn_graph, labels, k)
```

coding-famer:
I think #928 is quite related to this issue, maybe it's better to start from that?

jwmueller:
@coding-famer yep! you probably easily address both at once given you created all of this code :)

coding-famer:
> @coding-famer yep! you probably easily address both at once given you created all of this code :)

Sure, if you @01PrathamS are interested in these two issues, I'm glad to work with you.

01PrathamS:
Thank you @coding-famer for taking me into consideration, it's been great to work on this..

PR #1050, opened 2024-03-11
user-facing cleanlab.datavaluation module

coding-famer:
## Summary

Solve #973 

> 🎯 **Purpose**: Move the data valuation calculation function to `cleanlab.datavaluation` module and support `features` as input

# Necessary Imports:
`from cleanlab.datavaluation import data_shapley_knn`

## Impact

> 🌐 Areas Affected: `datalab.issue_manager.DataValuationIssueManager`.

**Unaddressed Cases**

One thing may need discussion: when both `knn_graph` and `features` are provided and the given `k` is larger than the `k` in `knn_graph`, the default behavior now is recomputing `knn_graph` using `features.

### DIFF
--- a/cleanlab/datalab/internal/issue_manager/data_valuation.py
+++ b/cleanlab/datalab/internal/issue_manager/data_valuation.py
@@ -34,6 +34,7 @@ from sklearn.exceptions import NotFittedError
 from sklearn.neighbors import NearestNeighbors
 from sklearn.utils.validation import check_is_fitted
 
+from cleanlab.datavaluation import data_shapley_knn
 from cleanlab.datalab.internal.issue_manager import IssueManager
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -161,7 +162,7 @@ class DataValuationIssueManager(IssueManager):
         if labels is None:
             raise ValueError("labels must be provided to run data valuation")
 
-        scores = _knn_shapley_score(knn_graph, labels, self.k)
+        scores = data_shapley_knn(knn_graph, labels, self.k)
 
         self.issues = pd.DataFrame(
             {
@@ -231,19 +232,3 @@ class DataValuationIssueManager(IssueManager):
                 statistics_dict["statistics"]["knn_metric"] = self.metric
 
         return statistics_dict
-
-
-def _knn_shapley_score(knn_graph: csr_matrix, labels: np.ndarray, k: int) -> np.ndarray:
-    """Compute the Shapley values of data points based on a knn graph."""
-    N = labels.shape[0]
-    scores = np.zeros((N, N))
-    dist = knn_graph.indices.reshape(N, -1)
-
-    for y, s, dist_i in zip(labels, scores, dist):
-        idx = dist_i[::-1]
-        ans = labels[idx]
-        s[idx[k - 1]] = float(ans[k - 1] == y)
-        ans_matches = (ans == y).flatten()
-        for j in range(k - 2, -1, -1):
-            s[idx[j]] = s[idx[j + 1]] + float(int(ans_matches[j]) - int(ans_matches[j + 1]))
-    return 0.5 * (np.mean(scores / k, axis=0) + 1)
--- a/cleanlab/datalab/internal/issue_manager/data_valuation.py
+++ b/cleanlab/datalab/internal/issue_manager/data_valuation.py
@@ -143,26 +143,22 @@ class DataValuationIssueManager(IssueManager):
                 )
             if self.metric is None:
                 self.metric = "cosine" if features.shape[1] > 3 else "euclidean"
-            knn = NearestNeighbors(n_neighbors=self.k, metric=self.metric).fit(features)
+            knn_graph = None
 
-            if self.metric and self.metric != knn.metric:
-                warnings.warn(
-                    f"Metric {self.metric} does not match metric {knn.metric} used to fit knn. "
-                    "Most likely an existing NearestNeighbors object was passed in, but a different "
-                    "metric was specified."
-                )
-            self.metric = knn.metric
-
-            try:
-                check_is_fitted(knn)
-            except NotFittedError:
-                knn.fit(features)
-
-            knn_graph = knn.kneighbors_graph(mode="distance")
         if labels is None:
             raise ValueError("labels must be provided to run data valuation")
 
-        scores = data_shapley_knn(knn_graph, labels, self.k)
+        scores, self.k, new_metric = data_shapley_knn(
+            labels, knn_graph=knn_graph, features=features, metric=self.metric, k=self.k
+        )
+
+        if self.metric and self.metric != new_metric:
+            warnings.warn(
+                f"Metric {self.metric} does not match metric {new_metric} used to fit knn. "
+                "Most likely an existing NearestNeighbors object was passed in, but a different "
+                "metric was specified."
+            )
+        self.metric = new_metric
 
         self.issues = pd.DataFrame(
             {
--- a/cleanlab/datavaluation.py
+++ b/cleanlab/datavaluation.py
@@ -1,0 +1,50 @@
+# Copyright (C) 2017-2024  Cleanlab Inc.
+# This file is part of cleanlab.
+#
+# cleanlab is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# cleanlab is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
+
+import numpy as np
+from scipy.sparse import csr_matrix
+
+
+def data_shapley_knn(
+    knn_graph: csr_matrix,
+    labels: np.ndarray,
+    k: int = 10,
+) -> np.ndarray:
+    """Compute the Shapley values of data points based on a knn graph.
+    Based on KNN-Shapley value described in https://arxiv.org/abs/1911.07128
+    The larger the score, the more valuable the data point is, the more contribution it will make to the model's training.
+
+    Parameters
+    ----------
+    knn_graph : csr_matrix
+        A sparse matrix representing the knn graph.
+    labels: np.ndarray
+        The labels of the data points.
+    k: int
+        The number of nearest neighbors to consider.
+    """
+    N = labels.shape[0]
+    scores = np.zeros((N, N))
+    dist = knn_graph.indices.reshape(N, -1)
+
+    for y, s, dist_i in zip(labels, scores, dist):
+        idx = dist_i[::-1]
+        ans = labels[idx]
+        s[idx[k - 1]] = float(ans[k - 1] == y)
+        ans_matches = (ans == y).flatten()
+        for j in range(k - 2, -1, -1):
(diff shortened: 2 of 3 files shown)


cleanlab/datalab/internal/issue_manager/data_valuation.py
```diff
@@ -142,26 +140,22 @@ def find_issues(
...
-                    "metric was specified."
-                )
-            self.metric = knn.metric
-
-            try:
-                check_is_fitted(knn)
-            except NotFittedError:
-                knn.fit(features)
-
-            knn_graph = knn.kneighbors_graph(mode="distance")
```
elisno, line 160: Please revert to this logic on deciding whether to compute a `knn_graph` within Datalab (and this issue manager).
The cleanlab.data_valuation module should not be responsible for computing a knn_graph **and passing it** on to Datalab.
It's fine to copy the helper function to the data_valuation module in the case when features are passed to the function in the `cleanlab.data_valuation`module.

cleanlab/datalab/internal/issue_manager/data_valuation.py
```diff
@@ -142,26 +140,22 @@ def find_issues(
...
+            labels, knn_graph=knn_graph, features=features, metric=self.metric, k=self.k
+        )
+
+        if self.metric and self.metric != new_metric:
+            warnings.warn(
+                f"Metric {self.metric} does not match metric {new_metric} used to fit knn. "
+                "Most likely an existing NearestNeighbors object was passed in, but a different "
+                "metric was specified."
+            )
+        self.metric = new_metric
```
elisno, line 158: See the previous comment about reverting the logic of computing within Datalab.

cleanlab/datavaluation.py
```diff
@@ -0,0 +1,98 @@
...
+    if metric == None:
+        metric = "cosine" if features.shape[1] > 3 else "euclidean"
+
+    knn = NearestNeighbors(n_neighbors=k, metric=metric).fit(features)
+    knn_graph = knn.kneighbors_graph(mode="distance")
+    try:
+        check_is_fitted(knn)
+    except NotFittedError:
+        knn.fit(features)
+    return knn_graph, knn.metric
```
elisno, line 59: This helper function only needs to compute the `knn_graph` within this module. This module is not responsible for tracking the metric used by the simple function. We're not re-implementing the logic of storing the metric outside of Datalab.

```suggestion
def _process_knn_graph_from_features(features: np.ndarray, metric: Optional[str], k: int = 10) -> csr_matrix:
    """Calculate the knn graph from the features if it is not provided in the kwargs."""
    if k > len(features):  # Ensure number of neighbors less than number of examples
        raise ValueError(
            f"Number of nearest neighbors k={k} cannot exceed the number of examples N={len(features)} passed into the estimator (knn)."
        )

    if metric == None:
        metric = "cosine" if features.shape[1] > 3 else "euclidean"

    knn = NearestNeighbors(n_neighbors=k, metric=metric).fit(features)
    knn_graph = knn.kneighbors_graph(mode="distance")
    try:
        check_is_fitted(knn)
    except NotFittedError:
        knn.fit(features)
    return knn_graph
```

cleanlab/datavaluation.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def data_shapley_knn(
+    labels: np.ndarray,
+    metric: str,
+    knn_graph: Optional[csr_matrix] = None,
+    features: Optional[np.ndarray] = None,
+    k: int = 10,
+) -> Tuple[np.ndarray, int, str]:
```
elisno:
```suggestion
) -> np.ndarray:
```

The original method was only supposed to return the scores ([see comment in issue](https://github.com/cleanlab/cleanlab/issues/973#issuecomment-1922822344)). This public function is not responsible for tracking the other information.

cleanlab/datavaluation.py
```diff
@@ -0,0 +1,98 @@
...
+                k = knn_graph.nnz // knn_graph.shape[0]
+                Warning(
+                    f"k is larger than the number of neighbors in the knn graph. Using k={k} instead."
+                )
+        new_metric = metric
+    else:
+        if features is None:
+            raise ValueError("features must be provided if knn_graph is not provided.")
+        knn_graph, new_metric = _process_knn_graph_from_features(features, metric, k)
+    return _knn_shapley_score(knn_graph, labels, k), k, new_metric
```
elisno, line 100: The method needs either features or a knn_graph. It's simplest in this function to prefer a knn_graph. Make this explicit in the docstring for this function.

```suggestion
    if knn_graph is None and features is None:
        raise ValueError("Either knn_graph or features must be provided.")
    
    if knn_graph is None:
        knn_graph = _process_knn_graph_from_features(cast(np.ndarray, features), metric, k)
    return _knn_shapley_score(knn_graph, labels, k)
```

elisno (changes requested):
Thanks for the PR.

Remember that the main idea of the PR is to move the `_knn_shapley_score` out of the DataValuationIssueManager and into a `cleanlab.data_valuation` module (which you need to rename correctly, by the way).

To keep the module's scope simple, it can borrow some logic from the Datalab necessary to compute the scores (e.g., how to compute the knn_graph), but it's not supposed to keep track of all this information.

elisno (approved):
Code LGTM.
Made some docs-related changes.

docs/source/index.rst
```diff
@@ -181,6 +181,7 @@ Link to Cleanlab Studio docs: `help.cleanlab.ai <https://help.cleanlab.ai/>`_
    cleanlab/dataset
    cleanlab/outlier
    cleanlab/multiannotator
+   cleanlab/data_valuation
```
jwmueller, line 184: put at bottom of the list, not here

cleanlab/data_valuation.py
```diff
@@ -0,0 +1,134 @@
...
+
+def data_shapley_knn(
+    labels: np.ndarray,
+    knn_graph: Optional[csr_matrix] = None,
+    features: Optional[np.ndarray] = None,
+    metric: Optional[str] = None,
+    k: int = 10,
+) -> np.ndarray:
+    """
+    Compute the Shapley values of data points using a K-Nearest Neighbors (KNN) graph.
```
jwmueller, line 80: Would be good to copy some of the text I'd already written describing this method in the issue types guide

Screen Shot 2024-03-25 at 9 34 21 PM

jwmueller: here is where to describe the approximate KNN algorithm, not in the top-level description of this module

cleanlab/data_valuation.py
```diff
@@ -0,0 +1,134 @@
...
+#
+# cleanlab is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
+"""
+Provides methods for computing the Shapley values of data points using a K-Nearest Neighbors (KNN) graph.
```
jwmueller, line 17: this top-level description of the module should be more vague. Don't mention one specific method or Data Shapely.

Just talk about Data Valuation in general and quantifying the benefits of individual data points to a trained ML model's performance.

That way we can add more methods to the module in the future and don't need to change the top-level description. Also do not need to list out individual methods in the top-level description of a module ever

cleanlab/data_valuation.py
```diff
@@ -0,0 +1,134 @@
...
+    Compute the Shapley values of data points using a K-Nearest Neighbors (KNN) graph.
+
+    This function calculates the contribution (Shapley value) of each data point in a dataset
+    for model training, based on the principle that data points contributing more to the
+    model's accuracy are of higher value.
+
+    Parameters
+    ----------
+    labels :
+        An array of labels for the data points.
```
jwmueller, line 89: need to mention this is only for multi-class classification datasets

cleanlab/data_valuation.py
```diff
@@ -0,0 +1,134 @@
...
+    knn_graph :
+        A precomputed sparse KNN graph. If not provided, it will be computed from the `features` using the specified `metric`.
+    features :
+        The feature matrix of data points. Necessary if `knn_graph` is not supplied.
```
jwmueller, line 93: feature matrix of data points is unclear language. Please copy language from elsewhere where we've described features args for eg. Datalab

cleanlab/data_valuation.py
```diff
@@ -0,0 +1,134 @@
...
+    knn_graph = knn.kneighbors_graph(mode="distance")
+    try:
+        check_is_fitted(knn)
+    except NotFittedError:
+        knn.fit(features)
+    return knn_graph
+
+
+def data_shapley_knn(
+    labels: np.ndarray,
```
jwmueller, line 73:
```suggestion
    labels: np.ndarray, *,
```

jwmueller: we always want to be able to re-order the keyword only args for every method in cleanlab package

cleanlab/data_valuation.py
```diff
@@ -0,0 +1,134 @@
...
+    knn_graph: Optional[csr_matrix] = None,
+    features: Optional[np.ndarray] = None,
```
jwmueller, line 75: put features arg ahead of knn_graph. Note how all the other args are related to knn_graph. 

Also recommended approach we expect most users to take is to supply features

cleanlab/data_valuation.py
jwmueller: swap ordering of features and knn graph in documentation too

cleanlab/data_valuation.py
```diff
@@ -0,0 +1,134 @@
...
+    knn_graph :
+        A precomputed sparse KNN graph. If not provided, it will be computed from the `features` using the specified `metric`.
+    features :
+        The feature matrix of data points. Necessary if `knn_graph` is not supplied.
+    metric : Optional[str], default=None
+        The distance metric for KNN graph construction.
+        Supports metrics available in ``sklearn.neighbors.NearestNeighbors``
+        Default metric is ``"cosine"`` for ``dim(features) > 3``, otherwise ``"euclidean"`` for lower-dimensional data.
+    k :
+        The number of neighbors to consider for the KNN graph and Shapley value computation.
```
jwmueller, line 99:
```suggestion
        The number of neighbors to consider for the KNN graph and Data Shapley value computation.
```

jwmueller: never say "Shapley value", it's always "Data Shapley value".
The former means something entirely different in ML (Feature attribution)

cleanlab/data_valuation.py
```diff
@@ -0,0 +1,134 @@
...
+        Must be less than the total number of data points.
+        The value may not exceed the number of neighbors of each data point stored in the KNN graph.
+
+    Returns
+    -------
+    scores :
+        An array containing transformed Shapley values for each data point, calibrated to indicate their relative importance.
+        These scores have been adjusted to fall within the range of 0 to 1.
+        Values closer to 1 suggest data points are highly influential and positively contribute to the model's performance.
+        Conversely, scores below 0.5 are interpreted as indicating a negative impact on model performance.
```
jwmueller, line 109:
```suggestion
    scores :
        An array of transformed Data Shapley values for each data point, calibrated to indicate their relative importance.
        These scores have been adjusted to fall within 0 to 1.
        Values closer to 1 indicate data points that are highly influential and positively contribute to a trained ML model's performance.
        Conversely, scores below 0.5 indicate data points estimated to  negatively impact model performance.
```

cleanlab/data_valuation.py
coding-famer: BTW, the document in this figure seems to need updating too (data valuation issue now can accept features as input), can I update it in this PR?

jwmueller: yes please! Thanks!

jwmueller (approved):
lgtm AFAICT, leaving merge to @elisno

elisno (approved):

PR #1087, opened 2024-04-07
Fix internal multilabel label

gogetron:
## Summary

> 🎯 **Purpose**: Fix bugs in cleanlab/datalab/internal/issue_manager/multilabel/label.py
>

The doctests were producing errors when I ran them. After fixing the `kwargs` arg without the `**` and fixing the expected results in the doctests I realized the expected outputs were a list inside a tuple thus the code was not correct. I added a new test to ensure that the kwargs added to this function are correctly forwarded to the other functions.

The bug can be reproduced using the following code. Note: I mostly copied the existing unit test data and code but modifying the `frac_noise` parameter. The result is that this parameter is ignored with current version. This PR fixes this issue.

```python
import numpy as np

from cleanlab.datalab.internal.issue_manager.multilabel.label import MultilabelIssueManager
from cleanlab.internal.multilabel_utils import onehot2int
from cleanlab import Datalab


np.random.seed(10)
true_y = np.array([[0, 1, 0], [0, 1, 1], [1, 1, 1], [1, 0, 1], [0, 1, 1], [1, 1, 0], [1, 1, 0]])
pred_probs = np.full_like(true_y, fill_value=0.1, dtype=float)
pred_probs[true_y == 1] = 0.9
# Flip labels of some rows to add noise
candidate_rows = np.where(true_y.sum(axis=1) < 3)[0]
noisy_rows = np.random.choice(candidate_rows, 2, replace=False)
noisy_y = true_y.copy()
noisy_y[noisy_rows] = 1 - true_y[noisy_rows]
labels = onehot2int(noisy_y)


lab = Datalab({"labels": labels}, task="multilabel", label_name="labels")
issue_manager = MultilabelIssueManager(datalab=lab)

issue_manager.find_issues(pred_probs=pred_probs, frac_noise=1)
issues = issue_manager.issues
issues.index[issues["is_label_issue"]].tolist()
# [3, 6]

issue_manager.find_issues(pred_probs=pred_probs, frac_noise=0.5)
issues = issue_manager.issues
issues.index[issues["is_label_issue"]].tolist()
# [3, 6]
# Here we would expect the output to be [3]
```

## Impact

> 🌐 Areas Affected: Any call to `cleanlab/datalab/internal/issue_manager/multilabel/label.py` with arguements that were not the default values.
>

## Testing

> 🔍 Testing Done: Existing test suite, doctests for the modified file and a new added test.
>

## References

> 📚 Include any extra information that may be helpful to reviewers of your
> Pull-Request here (e.g. relevant URLs or Wiki pages that could help give
> background information). 
>
> Share links, docs, or sources that facilitated your changes.
>
> If relevant, please include additional code snippets
> (and their outputs/return values) showing alternative ways to use your newly added methods.

### DIFF
--- a/cleanlab/datalab/internal/issue_manager/multilabel/label.py
+++ b/cleanlab/datalab/internal/issue_manager/multilabel/label.py
@@ -19,14 +19,15 @@ from typing import TYPE_CHECKING, Any, ClassVar, Dict, List
 
 import pandas as pd
 
-from cleanlab.multilabel_classification.filter import find_label_issues
-from cleanlab.multilabel_classification.rank import get_label_quality_scores
 from cleanlab.datalab.internal.issue_manager import IssueManager
 from cleanlab.internal.multilabel_utils import onehot2int
+from cleanlab.multilabel_classification.filter import find_label_issues
+from cleanlab.multilabel_classification.rank import get_label_quality_scores
 
 if TYPE_CHECKING:  # pragma: no cover
-    import pandas as pd
     import numpy.typing as npt
+    import pandas as pd
+
     from cleanlab.datalab.datalab import Datalab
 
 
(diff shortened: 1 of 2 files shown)


elisno (approved):
LGTM!

Thank you for spotting the accidental tuple and adding that test!

Just changing one type-hint.

PR #1091, opened 2024-04-08
remove sklearn install from tutorial

jwmueller:
it's already a cleanlab dependency, and now `pip install sklearn` leads to unhappy messages:

Screen Shot 2024-04-08 at 5 41 05 PM


elisno (approved):
LGTM! I've double-checked that there's no other mention of installing `scikit-learn` (or `sklearn`) anywhere else in this repo.

We just set [`scikit-learn` as a dependency in our pyproject.toml](https://github.com/cleanlab/cleanlab/blob/7d1f3350174b280aa5752059e39174098ed33bec/pyproject.toml#L13).

Issue #1092, opened 2024-04-10
Doctests are failing for some functions


gogetron:
I have realized after working on #1087 that there are other functions in some files with failing doctests / code examples in the functions documentation. In addition, the idea of running doctests in CI was mentioned in #366. If this is something that might be of interest I would be willing to work on fixing the doctests and then adding this verification as part of the CI. However it will require some work to fix all of the doctests and I wanted to ask first if this is something you are interested in. If so how would you prefer this issue to be handled (1 single PR with multiple commits fixing all doctests, multiple PRs fixing the doctests for each file and then another PR adding doctests to CI, ...)?

One example of this failing test can be found in "cleanlab/datalab/internal/issue_manager/label.py"
```python
...
@staticmethod
    def _process_find_label_issues_kwargs(kwargs: Dict[str, Any]) -> Dict[str, Any]:
        """Searches for keyword arguments that are meant for the
        CleanLearning.find_label_issues method call

        Examples
        --------
        >>> from cleanlab.datalab.internal.issue_manager.label import LabelIssueManager
        >>> LabelIssueManager._process_clean_learning_kwargs(thresholds=[0.1, 0.9])
        {'thresholds': [0.1, 0.9]}
        """
        accepted_kwargs = [
            "thresholds",
            "noise_matrix",
            "inverse_noise_matrix",
            "save_space",
            "clf_kwargs",
            "validation_func",
        ]
        return {k: v for k, v in kwargs.items() if k in accepted_kwargs and v is not None}
```
There a a few issues with the example code in the docs. First of all `_process_clean_learning_kwargs` does not exist for LabelIssueManager. In addition, thresholds is not a parameter for this function, we need to add `**` before the `kwargs` parameter in the function definition to make that behavior possible.

desboisGIT:
ok il do. i would like to update the docs like that:
```
>>> from cleanlab.datalab.internal.issue_manager.label import LabelIssueManager
>>> LabelIssueManager._process_find_label_issues_kwargs(thresholds=[0.1, 0.9])
{'thresholds': [0.1, 0.9]}
```

PR #1094, opened 2024-04-11
Improve docstrings in token_classification/summary.py

gogetron:
## Summary

> 🎯 **Purpose**: Fix doctest issues in token_classification/summary.py
>

To ensure that the doctests are working correctly, I ran:
```python
python3 -m doctest cleanlab/token_classification/summary.py
```
It produces an error on the current version that has been addressed with this PR.

I have removed the extra space in the print messages because `print` produces a new line by default, thus this space is not useful and conflicts with doctests.

## Testing

> 🔍 Testing Done: Existing test suite and `python3 -m doctest cleanlab/token_classification/summary.py`.
>

#1092

### DIFF
--- a/cleanlab/token_classification/summary.py
+++ b/cleanlab/token_classification/summary.py
@@ -90,9 +90,6 @@ def display_issues(
     ...     ["An", "sentence", "with", "a", "typo"],
     ... ]
     >>> display_issues(issues, tokens)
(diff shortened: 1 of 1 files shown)


elisno (approved):
Fixed a print condition for the display_issues (so the docstring requires less modification).

Ortherwise, LGTM!

Issue #862, opened 2023-10-05
Improve efficiency and runtimes or memory usage (performance optimization)


jwmueller:
Try to speed up the runtime of certain methods, or reduce how much memory they use.

This is a great issue to get started contributing to this repo! There are many ways to achieve speedup. It is easy to verify speedup via basic benchmarking that also ensures the results remain the same after you have edited a method.

Start by running one specific method on a big dataset and profiling it thoroughly to see where the code bottlenecks are.

Known methods where big speedups can likely be obtained include all methods in:

1. [`cleanlab.multiannotator`](https://docs.cleanlab.ai/master/cleanlab/multiannotator.html)
2. [`cleanlab.multilabel_classification`](https://docs.cleanlab.ai/master/cleanlab/multilabel_classification/)
3. [`cleanlab.segmentation`](https://docs.cleanlab.ai/master/cleanlab/segmentation/index.html)
4. [`cleanlab.object_detection`](https://docs.cleanlab.ai/master/cleanlab/object_detection/index.html)
5. [Datalab NonIIDIssueManager](https://docs.cleanlab.ai/master/cleanlab/datalab/internal/issue_manager/noniid.html)

arvind-swami:
DO you have any handy large real-world multiannotator dataset I can use?

jwmueller:
> DO you have any handy large real-world multiannotator dataset I can use?

There are some multiannotator datasets here:

https://github.com/cleanlab/multiannotator-benchmarks

https://github.com/cleanlab/examples/blob/master/multiannotator_cifar10/multiannotator_cifar10.ipynb

PR #1098, opened 2024-04-14
Perf object detection

gogetron:
## Summary
This PR partially addresses #862 

> 🎯 **Purpose**: Improve performance of find_label_issues in object_detection/filter.py file.
>

A significant improvement comes from vectorizing `_get_iou`, `_euc_dis` and `_calculate_average_precision`. I searched for all references to the deleted `_mod_coordinates` and `_get_overlap` but they are not used anywhere else in the code. I assumed it is ok to remove them because they are not part of the public api however let me know if you prefer to deprecate them or keep both versions.

For memory I used the memory-profiler library. The code I used for benchmarking is copied below. In addition I sorted the imports in the modified files.

**Code Setup**
```python
import random

import numpy as np

from cleanlab.object_detection.filter import find_label_issues

np.random.seed(0)
random.seed(0)

N = 5_000
SIZE = 256
MAX_L = 10
MAX_M = 100
K = 20

# Create input data
labels = []
predictions = []
for _ in range(N):
    x = random.randint(1, MAX_L)
    bboxes = np.empty((x, 4))
    bboxes[:, :2] = np.random.randint(0, SIZE // 2, size=(x, 2))
    bboxes[:, 2:] = np.random.randint(SIZE // 2, SIZE, size=(x, 2))
    img_labels = np.random.randint(K, size=x)
    label = {"bboxes": bboxes, "labels": img_labels}
    labels.append(label)

    y = random.randint(x, MAX_M)
    prediction = np.empty((K, y, 5))
    prediction[:, :, :2] = np.random.randint(0, SIZE // 2, size=(K, y, 2))
    prediction[:, :, 2:4] = np.random.randint(SIZE // 2, SIZE, size=(K, y, 2))
    prediction[:, :, 4] = np.random.random((K, y))
    predictions.append(prediction)
```

**Current version**
```python
%%timeit
%memit find_label_issues(labels, predictions)
# peak memory: 1317.51 MiB, increment: 820.18 MiB
# peak memory: 1317.58 MiB, increment: 0.07 MiB
# peak memory: 1317.60 MiB, increment: 0.02 MiB
# peak memory: 1317.86 MiB, increment: 0.28 MiB
# peak memory: 1319.04 MiB, increment: 1.18 MiB
# peak memory: 1319.09 MiB, increment: 0.07 MiB
# peak memory: 1319.09 MiB, increment: 0.02 MiB
# peak memory: 1319.11 MiB, increment: 0.02 MiB
# 8min 7s ± 30.2 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

**This PR**
```python
%%timeit
%memit find_label_issues(labels, predictions)
# peak memory: 1291.41 MiB, increment: 792.09 MiB
# peak memory: 1291.82 MiB, increment: 1.19 MiB
# peak memory: 1291.87 MiB, increment: 0.05 MiB
# peak memory: 1292.05 MiB, increment: 0.18 MiB
# peak memory: 1292.07 MiB, increment: 0.00 MiB
# peak memory: 1292.07 MiB, increment: 0.00 MiB
# peak memory: 1292.07 MiB, increment: 0.00 MiB
# peak memory: 1292.07 MiB, increment: 0.00 MiB
# 29.2 s ± 574 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

## Testing

> 🔍 Testing Done: Existing test suite and output verification (explained below).

## References

I was not sure about which input would be more realistic, I have tried different options but the improvements were similar, however let me know if you want me to test a specific combination of inputs, I tried to use the largest input dataset I could while still being able to run the master %%timeit in about an hour.

I have noticed that the `num_procs` argument in `_calculate_ap_per_class` is always set to the default value. Thus, i made a few changes there to ensure that the pool is only created when `num_procs` is higher than 1.

After each change I ran the tests and they were catching any minor mistake I made. However once they were passing I ensured with some random input that both versions (master and this PR) still produced the same exact output.

### DIFF
--- a/cleanlab/object_detection/rank.py
+++ b/cleanlab/object_detection/rank.py
@@ -280,12 +280,13 @@ def _separate_prediction_single_box(
 ) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
     """Separates predictions into class labels, bounding boxes and pred_prob lists"""
     labels = []
-    boxes = []
+    bboxes = []
+    pred_probs = []
     for idx, prediction_class in enumerate(prediction):
         labels.extend([idx] * len(prediction_class))
-        boxes.extend(prediction_class.tolist())
-    bboxes = [box[:4] for box in boxes]
-    pred_probs = [box[-1] for box in boxes]
+        for inner_pred_class in prediction_class:
+            bboxes.append(inner_pred_class[:4])
+            pred_probs.append(inner_pred_class[-1])
     return np.array(bboxes), np.array(labels), np.array(pred_probs)
 
 
@@ -315,12 +316,9 @@ def _separate_prediction(
 def _get_overlap_matrix(bb1_list: np.ndarray, bb2_list: np.ndarray) -> np.ndarray:
     """Takes in two lists of bounding boxes and returns an IoU matrix where IoU[i][j] is the overlap between
     the i-th box in `bb1_list` and the j-th box in `bb2_list`."""
-    wd = np.zeros(shape=(len(bb1_list), len(bb2_list)))
-    if not len(bb2_list):
-        return wd
-    for i, bb1 in enumerate(bb1_list):
-        wd[i] = _get_iou(bb1, bb2_list)
-    return wd
+    if not len(bb1_list) or not len(bb2_list):
+        return np.zeros(shape=(len(bb1_list), len(bb2_list)))
+    return _get_iou(bb1_list, bb2_list)
 
 
 def _get_iou(bb1: np.ndarray, bb2: np.ndarray) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


cleanlab/object_detection/rank.py
```diff
@@ -314,72 +313,55 @@ def _separate_prediction(
...
+    bb1 : np.ndarray
+        Shape [N1, 4] where bb2[i, :] must be an array of shape [x1, y1, x2, y2] corresponding to the bouding box i.
+    bb2 : np.ndarray
+        Shape [N2, 4] where bb2[j, :] must be an array of shape [x1, y1, x2, y2] corresponding to the bouding box j.
     Returns
     -------
-    float
-        in [0, 1]
+    np.ndarray
+        values in [0, 1]
```
elisno, line 339:
```suggestion
    iou: np.ndarray
        An array of shape [N1, N2] with values in [0, 1], where iou[i, j] shows the Intersection over Union between bounding box i in bb1 and bounding box j in bb2.
```

elisno (changes requested):
Nice work @gogetron!
I have a few suggestions for rank.py.

gogetron:
Thank you for your review, your suggestions are very clear and useful. I have just committed your suggestions.

PR #1108, opened 2024-04-22
Remove unsqueeze call in image tutorial

elisno:
With the latest release of the datasets library (2.19.0), the torch formatter now handles PIL objects of 2d arrays (e.g. grayscale images) differently.

Now when calling `Dataset.with_format("torch")`, every image tensor now has shape (1, H, W) for grayscale images. Previously, this was done by the `unsqueeze(0)` call in the subsequent step in the preprocessing pipeline, but calling `unsqueeze` is redundant for 2d image datasets (stored as PIL objects).

You can verify this by running:

```python
from datasets import load_dataset

dataset = load_dataset("fashion_mnist", split="train")
transformed_dataset = dataset.with_format("torch")
transformed_dataset[:2]["image"].shape
#torch.Size([2, 1, 28, 28])
```

Previously, this would have returned

```
#torch.Size([2, 28, 28])
```

which required the call to `unsqueeze`.

### DIFF
--- a/docs/source/tutorials/datalab/image.ipynb
+++ b/docs/source/tutorials/datalab/image.ipynb
@@ -57,7 +57,7 @@
     "You can use `pip` to install all packages required for this tutorial as follows:\n",
     "\n",
     "```ipython3\n",
-    "!pip install matplotlib torch torchvision datasets\n",
+    "!pip install matplotlib torch torchvision datasets>=2.19.0\n",
     "!pip install \"cleanlab[image]\"\n",
     "# We install cleanlab with extra dependencies for image data\n",
     "# Make sure to install the version corresponding to this tutorial\n",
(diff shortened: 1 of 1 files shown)


sanjanag (approved):

docs/source/tutorials/datalab/image.ipynb
```diff
@@ -173,7 +173,7 @@
     "\n",
     "# Apply transformations\n",
     "def normalize(example):\n",
-    "    example[\"image\"] = (example[\"image\"] / 255.0).unsqueeze(0)\n",
+    "    example[\"image\"] = (example[\"image\"] / 255.0)\n",
```
jwmueller, line 176: does this change mean the code no longer works on older versions of the `datasets` package? If so would be good to specify which is the lower-bound version at the top of tutorial

elisno: That's right.
It's an backwards incompatible change in the datasets package that affects the dataset.with_format("torch") call

jwmueller, line 176:
```suggestion
    "    example[\"image\"] = (example[\"image\"] / 255.0)  # each pixel value was originally between 0 and 255 \n",
```

Issue #889, opened 2023-11-09
[code refactor] features -> KNN graph code in Datalab Issue Managers


jwmueller:
The KNN graph is currently being constructed in each IssueManager, via duplicated code from the other IssueManagers. 

Goal: make this a helper method that is reused across the issue-managers

01PrathamS:
here we are are trying to move ```_process_knn_graph_from_inputs``` this method in another file say utils.py with class name ConstructedKnnGraph and an instance method ``process_knn_graph_from_inputs```

PR #1113, opened 2024-04-30
Fix numerical instability with Euclidean distance metric

### LINKED ISSUE #1055: add unit test with all identical dataset
Run datalab on this dataset and verify especially near duplicates and for outliers have expected outputs.
But we also want to validate that label issues and other things behave somewhat reasonably for future versions of the codebase.

Let's have all of these datapoints have the same label, except one datapoint which has different label. 
Verify label issue is properly detected in this one datapoint

elisno:
## Summary

The currently selected implementation for computing the pair-wise Euclidean distances between data points is computationally efficient but sacrifices numerical stability when dealing with nearly-identical data points.

```python
import numpy as np
from sklearn.metrics.pairwise import euclidean_distances

# Comparing 2d points
a = np.array([[100000001, 100000000]])
b = np.array([[100000000, 100000001]])

print("sklearn:", euclidean_distances(a, b)[0,0])  # Used under-the-hood by NearestNeighbors
print("correct:", np.sqrt(np.sum((a-b)**2)))
```

The same happens for small values. While the relative error is small, the absolute error can be arbitrarily large.

SciPy has an accurate implementation and is already supported by scikit-learn (often as a string argument, with the exception of the `"euclidean"`). Due to a name clash with the [scikit-learn implementation](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.euclidean_distances.html), passing the [callable](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.euclidean.html#scipy.spatial.distance.euclidean) directly to NearestNeighbors is done in this PR.

Some benchmarks I ran show that the SciPy metric is several orders of magnitude slower for larger inputs (y-axis is runtime, x-axis is the number of points to compare): 
image.
For smaller datasets, it may be preferable to run the precise algorithm. Using the current scikit-learn implementation should generally be fine, as long as the norm of the data points isn't significantly greater than their mutual distances.

Code for reproducing this figure can be found in [this gist](https://gist.github.com/elisno/5b92fe0414f113089047b20bb447d65f).

## Impact

This PR affects all parts of the code-base that instantiate a NearestNeighbors object and select a default metric based on the dimensionality of the features. The results for the euclidean metrics will be more precise for smaller datasets, as it uses the numerically stable algorithm to find nearest neighbors.
For larger datasets, the computationally efficient algorithm implemented in scikit-learn is still used.

## Testing

> 🔍 Testing Done: Tests in tests/datalab/datalab/end_to_end_tests/test_all_identical_examples.py have been updated to use a larger dataset of 100 points (the cutoff between the accurate and the efficient implementations for computing the euclidean pairwise distances). The same is done in tests/test_outlier.py.

**Unaddressed Cases**

Testing the "all-identical" datasets with [scikit-learn's `"euclidean"` metric](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.euclidean_distances.html) has not been addressed.
It's not possible to guarantee an upper limit for the numerical errors it gets.

-

- This issue resolves #1055 

## References

You may wish to run this test code to get a better sense of how the different metrics work with scikit-learn and cleanlab's `OutOfDistribution` class.

```python
# test_distance_computations.py

"""This test module provides various tests to validate the behavior of distance calculation functionalities in a numerical computing environment.
It utilizes a combination of libraries including numpy, scipy, scikit-learn, hypothesis for property-based testing, and cleanlab for handling outlier detection.

The module includes the following main components:

1. `calculate_distances`: A function to calculate the nearest neighbor distances using different metrics provided by the scikit-learn library.

2. `test_calculate_distances_for_2d_data`: A test to ensure that the distance calculations are accurate and stable for two-dimensional data using multiple distance metrics, emphasizing the handling of numerical instabilities in certain metrics.

3. `test_calculate_distances_for_high_dimensional_data`: Similar to the 2D test but adapted for high-dimensional data, testing the robustness and correctness of distance calculations across various metrics, particularly focusing on the potential for numerical instabilities.

4. `test_outlier_scores_based_on_features_for_high_dimensional_data`: Tests the functionality of outlier detection based on distances in a high-dimensional space, ensuring that the OutOfDistribution class from cleanlab operates as expected with various distance metrics.

Each test is constructed using the hypothesis library to generate a wide range of test cases, ensuring comprehensive coverage of potential issues with distance calculations and outlier detection functionalities.
"""


import numpy as np

from scipy.spatial.distance import euclidean

from hypothesis import given, settings
from hypothesis.extra.numpy import arrays
from hypothesis.strategies import floats, integers

from cleanlab.outlier import OutOfDistribution


def calculate_distances(X: np.ndarray, nn: NearestNeighbors) -> np.ndarray:
    # Query the nearest neighbors
    distances, _ = nn.kneighbors(X)
    return distances


@given(X=arrays(float, shape=(1, 2), elements=floats(0, 1_000_000_000, allow_nan=False, allow_infinity=False, allow_subnormal=False, exclude_min=True)))
@settings(max_examples=1000, deadline=None)
def test_calculate_distances_for_2d_data(X: np.ndarray):

    # Store distances for each metric
    distances_dict: dict[str, np.ndarray] = {}

    metrics = ["euclidean", "cosine", euclidean, "seuclidean"]
    metric_ids = ["euclidean", "cosine", "euclidean from scipy"]
    for (metric, id)  in zip(metrics, metric_ids):
        if id == "euclidean":
            # The "euclidean" metric from scikit-learn suffers from numerical instability when the distance
            # between points in X are small compared to their norms. This is because the distance is calculated
            # as dot(x,x) - 2 * dot(x, y) + dot(y, y). These individual terms can blow up to huge values that are
            # not numerically stable. The upside of this implementation is the computational efficiency and allows for precomputing either
            # the dot(x,x) or dot(y,y) terms.
            continue
        nn = NearestNeighbors(n_neighbors=1, metric=metric).fit(X)
        distances_dict[id] = calculate_distances(X, nn)

    # Collect {metric_id: distances} that are not close to zero
    distances_not_close_to_zero = {
        metric_id: distances[0, 0]
        for metric_id, distances in distances_dict.items() 
        if not np.allclose(distances, 0, atol=1e-17)
    }
    
    assert not distances_not_close_to_zero, f"Distances between identical data points are not close to zero for the following metrics: {distances_not_close_to_zero}"

@given(X=arrays(float, shape=integers(2, 1000), elements=floats(0, 1_000_000_000_000, allow_nan=False, allow_infinity=False, allow_subnormal=False, exclude_min=True)))
@settings(max_examples=1000, deadline=None)
def test_calculate_distances_for_high_dimensional_data(X: np.ndarray):

    X = X.reshape(1, -1)
    # Store distances for each metric
    distances_dict: dict[str, np.ndarray] = {}

    metrics = ["euclidean", "cosine", euclidean]
    metric_ids = ["euclidean", "cosine", "euclidean from scipy"]
    for (metric, id)  in zip(metrics, metric_ids):
        if id == "euclidean":
            # The "euclidean" metric from scikit-learn suffers from numerical instability when the distance
            # between points in X are small compared to their norms. This is because the distance is calculated
            # as dot(x,x) - 2 * dot(x, y) + dot(y, y). These individual terms can blow up to huge values that are
            # not numerically stable. The upside of this implementation is the computational efficiency and allows for precomputing either
            # the dot(x,x) or dot(y,y) terms.
            continue
        nn = NearestNeighbors(n_neighbors=1, metric=metric).fit(X)
        distances_dict[id] = calculate_distances(X, nn)

    # Collect {metric_id: distances} that are not close to zero
    distances_not_close_to_zero = {
        metric_id: distances[0, 0]
        for metric_id, distances in distances_dict.items() 
        if not np.allclose(distances, 0, atol=1e-15)
    }
    
    assert not distances_not_close_to_zero, f"Distances between identical data points are not close to zero for the following metrics: {distances_not_close_to_zero}"


# Now run the same checks through the OutOfDistribution class

@given(X=arrays(float, shape=integers(2, 1000), elements=floats(0, 1_000_000_000_000, allow_nan=False, allow_infinity=False, allow_subnormal=False, exclude_min=True)))
def test_outlier_scores_based_on_features_for_high_dimensional_data(X: np.ndarray):

    X = X.reshape(1, -1)
    X = np.repeat(X, 11, axis=0)
    # Store distances for each metric
    distances_dict: dict[str, np.ndarray] = {}

    metrics = ["euclidean", "cosine", euclidean]
    metric_ids = ["euclidean", "cosine", "euclidean from scipy"]
    for (metric, id)  in zip(metrics, metric_ids):
        if id == "euclidean":
            # The "euclidean" metric from scikit-learn suffers from numerical instability when the distance
            # between points in X are small compared to their norms. This is because the distance is calculated
            # as dot(x,x) - 2 * dot(x, y) + dot(y, y). These individual terms can blow up to huge values that are
            # not numerically stable. The upside of this implementation is the computational efficiency and allows for precomputing either
            # the dot(x,x) or dot(y,y) terms.
            continue
        nn = OutOfDistribution(params={"metric": metric, 
        distances_dict[id] = calculate_distances(X, nn)

    # Collect {metric_id: distances} that are not close to zero
    distances_not_close_to_zero = {
        metric_id: distances[0, 0]
        for metric_id, distances in distances_dict.items() 
        if not np.allclose(distances, 0, atol=1e-15)
    }
    
    assert not distances_not_close_to_zero, f"Distances between identical data points are not close to zero for the following metrics: {distances_not_close_to_zero}"
```
Try running it with something like:

```pytest test_distance_computations.py```

### DIFF
--- a/cleanlab/data_valuation.py
+++ b/cleanlab/data_valuation.py
@@ -19,10 +19,11 @@ Data Valuation helps us assess individual training data points' contributions to
 """
 
 
-from typing import Optional, cast
+from typing import Callable, Optional, Union, cast
 
 import numpy as np
 from scipy.sparse import csr_matrix
+from scipy.spatial.distance import euclidean
 from sklearn.neighbors import NearestNeighbors
 from sklearn.exceptions import NotFittedError
 from sklearn.utils.validation import check_is_fitted
@@ -45,7 +46,7 @@ def _knn_shapley_score(knn_graph: csr_matrix, labels: np.ndarray, k: int) -> np.
 
 
 def _process_knn_graph_from_features(
-    features: np.ndarray, metric: Optional[str], k: int = 10
+    features: np.ndarray, metric: Optional[Union[str, Callable]], k: int = 10
 ) -> csr_matrix:
     """Calculate the knn graph from the features if it is not provided in the kwargs."""
     if k > len(features):  # Ensure number of neighbors less than number of examples
@@ -53,7 +54,11 @@ def _process_knn_graph_from_features(
             f"Number of nearest neighbors k={k} cannot exceed the number of examples N={len(features)} passed into the estimator (knn)."
         )
     if metric == None:
-        metric = "cosine" if features.shape[1] > 3 else "euclidean"
+        metric = (
+            "cosine"
+            if features.shape[1] > 3
+            else "euclidean" if features.shape[0] > 100 else euclidean
+        )
     knn = NearestNeighbors(n_neighbors=k, metric=metric).fit(features)
     knn_graph = knn.kneighbors_graph(mode="distance")
     try:
@@ -68,7 +73,7 @@ def data_shapley_knn(
     *,
     features: Optional[np.ndarray] = None,
     knn_graph: Optional[csr_matrix] = None,
-    metric: Optional[str] = None,
+    metric: Optional[Union[str, Callable]] = None,
     k: int = 10,
 ) -> np.ndarray:
     """
@@ -94,10 +99,12 @@ def data_shapley_knn(
             If provided, this must be a 2D array with shape (num_examples, num_features).
     knn_graph :
         A precomputed sparse KNN graph. If not provided, it will be computed from the `features` using the specified `metric`.
-    metric : Optional[str], default=None
+    metric : Optional[str or Callable], default=None
         The distance metric for KNN graph construction.
         Supports metrics available in ``sklearn.neighbors.NearestNeighbors``
         Default metric is ``"cosine"`` for ``dim(features) > 3``, otherwise ``"euclidean"`` for lower-dimensional data.
+        The euclidean is computed with an efficient implementation from scikit-learn when the number of examples is greater than 100.
+        When the number of examples is 100 or fewer, a more numerically stable version of the euclidean distance from scipy is used.
     k :
         The number of neighbors to consider for the KNN graph and Data Shapley value computation.
         Must be less than the total number of data points.
--- a/cleanlab/datalab/internal/issue_manager/duplicate.py
+++ b/cleanlab/datalab/internal/issue_manager/duplicate.py
@@ -15,12 +15,13 @@
 # along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Dict, List, Optional, Union
 import warnings
 
 import numpy as np
 import pandas as pd
 from scipy.sparse import csr_matrix
+from scipy.spatial.distance import euclidean
 from sklearn.neighbors import NearestNeighbors
 from sklearn.exceptions import NotFittedError
 from sklearn.utils.validation import check_is_fitted
@@ -53,7 +54,7 @@ class NearDuplicateIssueManager(IssueManager):
     def __init__(
         self,
         datalab: Datalab,
-        metric: Optional[str] = None,
+        metric: Optional[Union[str, Callable]] = None,
         threshold: float = 0.13,
         k: int = 10,
         **_,
@@ -79,7 +80,11 @@ class NearDuplicateIssueManager(IssueManager):
                     "If a knn_graph is not provided, features must be provided to fit a new knn."
                 )
             if self.metric is None:
-                self.metric = "cosine" if features.shape[1] > 3 else "euclidean"
+                self.metric = (
+                    "cosine"
+                    if features.shape[1] > 3
+                    else "euclidean" if features.shape[0] > 100 else euclidean
+                )
             knn = NearestNeighbors(n_neighbors=self.k, metric=self.metric)
 
             if self.metric and self.metric != knn.metric:
--- a/cleanlab/datalab/internal/issue_manager/data_valuation.py
+++ b/cleanlab/datalab/internal/issue_manager/data_valuation.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     ClassVar,
     Dict,
     List,
@@ -30,6 +31,7 @@ import warnings
 import numpy as np
 import pandas as pd
 from scipy.sparse import csr_matrix
+from scipy.spatial.distance import euclidean
 from sklearn.exceptions import NotFittedError
 from sklearn.neighbors import NearestNeighbors
 from sklearn.utils.validation import check_is_fitted
@@ -101,7 +103,7 @@ class DataValuationIssueManager(IssueManager):
     def __init__(
         self,
         datalab: Datalab,
-        metric: Optional[str] = None,
+        metric: Optional[Union[str, Callable]] = None,
         threshold: Optional[float] = None,
         k: int = 10,
         **kwargs,
@@ -142,7 +144,11 @@ class DataValuationIssueManager(IssueManager):
                     "If a knn_graph is not provided, features must be provided to fit a new knn."
                 )
             if self.metric is None:
-                self.metric = "cosine" if features.shape[1] > 3 else "euclidean"
+                self.metric = (
+                    "cosine"
+                    if features.shape[1] > 3
+                    else "euclidean" if features.shape[0] > 100 else euclidean
+                )
             knn = NearestNeighbors(n_neighbors=self.k, metric=self.metric).fit(features)
 
             if self.metric and self.metric != knn.metric:
--- a/cleanlab/datalab/internal/issue_manager/noniid.py
+++ b/cleanlab/datalab/internal/issue_manager/noniid.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, ClassVar, Dict, Optional, Union, cast
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Dict, Optional, Union, cast
 import warnings
 import itertools
 
@@ -8,6 +8,7 @@ from scipy.stats import gaussian_kde
 import numpy as np
 import pandas as pd
 from scipy.sparse import csr_matrix
+from scipy.spatial.distance import euclidean
 from sklearn.neighbors import NearestNeighbors
 from sklearn.exceptions import NotFittedError
 from sklearn.utils.validation import check_is_fitted
@@ -106,7 +107,7 @@ class NonIIDIssueManager(IssueManager):
     def __init__(
         self,
         datalab: Datalab,
-        metric: Optional[str] = None,
+        metric: Optional[Union[str, Callable]] = None,
         k: int = 10,
         num_permutations: int = 25,
         seed: Optional[int] = 0,
@@ -203,7 +204,11 @@ class NonIIDIssueManager(IssueManager):
         features_to_use = self._determine_features(features, pred_probs)
 
         if self.metric is None:
-            self.metric = "cosine" if features_to_use.shape[1] > 3 else "euclidean"
+            self.metric = (
+                "cosine"
+                if features_to_use.shape[1] > 3
+                else "euclidean" if features_to_use.shape[0] > 100 else euclidean
+            )
 
         knn = NearestNeighbors(n_neighbors=self.k, metric=self.metric)
 
--- a/cleanlab/datalab/internal/issue_manager/underperforming_group.py
+++ b/cleanlab/datalab/internal/issue_manager/underperforming_group.py
@@ -15,13 +15,14 @@
 # along with cleanlab.  If not, see <https://www.gnu.org/licenses/>.
 from __future__ import annotations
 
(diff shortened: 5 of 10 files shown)

### CODE REVIEW

File: cleanlab/datalab/internal/issue_manager/duplicate.py [hunk 1]
```diff
@@ -79,7 +80,11 @@ def find_issues(
                     "If a knn_graph is not provided, features must be provided to fit a new knn."
                 )
             if self.metric is None:
-                self.metric = "cosine" if features.shape[1] > 3 else "euclidean"
+                self.metric = (
+                    "cosine"
+                    if features.shape[1] > 3
+                    else "euclidean" if features.shape[0] > 100 else euclidean
+                )
```
jwmueller, line 87: this code is duplicated in multiple methods, should probably all be an imported helper method?
eg. `self.metric = get_knn_metric(features)`

elisno: Thanks for pointing this out! This duplication comes from outlier.py ([see diff](https://github.com/cleanlab/cleanlab/pull/1113/files#diff-2831e7e2853b669892b9ed7c43245319d8b116093d2813d1a2c0b3b121131edfL429-L431)). The broader refactor you're suggesting would be valuable and aligns with what we want in #889. However, to avoid creeping elegance, let's handle that separately and keep this PR focused on the immediate bug fix (a numerically unstable algorithm for computing the Euclidean distance is affecting tests for all-identical examples). The docstrings ([primarily this one](https://github.com/cleanlab/cleanlab/pull/1113/files#diff-2831e7e2853b669892b9ed7c43245319d8b116093d2813d1a2c0b3b121131edfR64-R73)) address the current parameter choices; I'll ensure they're clear. We can revisit the refactor in a future PR.

elisno: This PR has turned into a bit of [shotgun surgery](https://refactoring.guru/smells/shotgun-surgery). I suggest we handle the refactor separately to keep things manageable.

ulya-tkch:
Why is the cutoff 100 datapoints?

I see the graph and understand the timing I just feel like setting the n to such a small value will make this PR invisible to most users who will keep getting differing results. What was your logic for deciding that value?

elisno:
> Why is the cutoff 100 datapoints?

Thank you for your feedback on the cutoff value. I understand your concerns regarding the choice of 100 data points and how it might appear to cater to very specific cases.

This cutoff addresses an edge case that, while not common, is crucial for ensuring the integrity of our distance calculations when dealing with uniform datasets where all N data points are identical. Such datasets can significantly skew the quality scores.


The `self.metric` configuration logic you're alluding to:

```python
self.metric = (
    "cosine"
    if features.shape[1] > 3
    else "euclidean" if features.shape[0] > 100 else "euclidean"
)
```

is becoming a point of concern. This sort of condition stacking is a symptom of our code adapting to contrived scenarios. It risks complicating the maintenance and could potentially introduce bugs or performance issues not immediately apparent.

> I see the graph and understand the timing I just feel like setting the n to such a small value will make this PR invisible to most users (or is the value more stable if there are more than 100 datapoints?)

Indeed, the original `"euclidean"` distance metric works fine in most cases, so we should use it! Also, the Euclidean distance metric isn't invariant to scaling. Surely users remember to scale their features to reasonable values?

The choice of 100 as a cutoff is primarily driven by considerations of computational efficiency balanced against algorithmic accuracy. This value was determined based on tests run on a single desktop, only for 1NN-search. If we were to increase the cutoff to 200 data points, the algorithm would run slower, especially on not-so-powerful computers, making the experience worse for many users.

<details><summary>Some quick benchmark printouts</summary>
<p>

Scipy is about 76x slower at N = 100, while ~127x slower at N=200. The slowdown gets much worse for larger values.

```python
from scipy.spatial.distance import euclidean
import numpy as np

def run_scipy(X, k):
    knn = NearestNeighbors(n_neighbors=k, metric=euclidean)
    knn.fit(X)
    knn.kneighbors()

def run_sklearn(X, k):
    knn = NearestNeighbors(n_neighbors=k, metric="euclidean")
    knn.fit(X)
    knn.kneighbors()


for N in [10, 50, 100, 200, 500, 1000, 2000, 5000]:
    X = np.random.rand(N, 2)
    k = min(N-1, 10)

    print(f"N={N}")
    print("Sklearn")
    %timeit run_sklearn(X, k)
    print("Scipy")
    %timeit run_scipy(X, k)

    print("\n---\n")
```

```
N=10
Sklearn
832 µs ± 5.85 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Scipy
720 µs ± 3.93 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

---

N=50
Sklearn
458 µs ± 19.9 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Scipy
11.1 ms ± 888 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

---

N=100
Sklearn
517 µs ± 19.6 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Scipy
39.3 ms ± 844 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

---

N=200
Sklearn
662 µs ± 27 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Scipy
84.3 ms ± 2.21 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

---

N=500
Sklearn
1.05 ms ± 12.1 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Scipy
212 ms ± 4.03 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

---

N=1000
Sklearn
1.84 ms ± 11.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Scipy
533 ms ± 7.71 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

---

N=2000
Sklearn
3.44 ms ± 21.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
Scipy
1.17 s ± 100 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

---

N=5000
Sklearn
9.17 ms ± 36.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
Scipy
4.07 s ± 107 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

</p>
</details> 


A slowdown of over two orders of magnitude at N=200, especially on a high-end system, is not an attractive option considering general hardware limitations.

> who will keep getting differing results.

This PR focuses on ensuring that the knn search algorithm functions effectively without requiring users to make significant adjustments or face unpredictable performance. It’s designed to correct numerical inconsistencies specific to cases like the following, ensuring that our approach is both precise and pragmatic:

```python
>>> X
array([[0.06, 2.03],
       [0.06, 2.03],
       [0.06, 2.03],
       [0.06, 2.03],
       [0.06, 2.03],
       [0.06, 2.03],
       ...
       [0.06, 2.03],
       [0.06, 2.03],
       [0.06, 2.03],
       [0.06, 2.03]])
>>> knn = NearestNeighbors(..., metric="euclidean").fit(X)
>>> distances, _ = knn.kneighbors(return_distances=True)
>>> # Use distances for scores
```

The `"euclidean"` metric is generally effective, but may not perform poorly in scenarios where there is an extreme skew in the scale of features or many points are close together. Such conditions can greatly affect the accuracy of distance measurements. To mitigate this, it is the responsibility of users to ensure their datasets are appropriately scaled. 

Addressing these edge cases shouldn't compromise the clarity and efficiency of our code. It may be worthwhile to reconsider if a `knn` graph-based approach is always appropriate, or if we should explore more robust methodologies for these specific situations. This would help us avoid overfitting our solution to scenarios that don't align well with the practical use cases most of our users face.

I'm open to revisiting our current approach and discussing alternatives that can simplify our implementation while still providing the robustness users expect. Let's find a balance between practicality and performance, ensuring our solutions remain accessible to all users without unnecessary complexity.

cleanlab/datalab/internal/issue_manager/duplicate.py
jwmueller: at least put a TODO or GH issue? How is anybody going to remember to factor out the helper method later on?

cleanlab/data_valuation.py
```diff
@@ -45,15 +46,19 @@ def _knn_shapley_score(knn_graph: csr_matrix, labels: np.ndarray, k: int) -> np.
...
     if k > len(features):  # Ensure number of neighbors less than number of examples
         raise ValueError(
             f"Number of nearest neighbors k={k} cannot exceed the number of examples N={len(features)} passed into the estimator (knn)."
         )
     if metric == None:
-        metric = "cosine" if features.shape[1] > 3 else "euclidean"
+        metric = (
+            "cosine"
+            if features.shape[1] > 3
+            else "euclidean" if features.shape[0] > 100 else euclidean
```
jwmueller, line 60: upon further reflection, the limitations of this are unclear to me.

What happens when:

1. the dataset has size 101 and is all identical?

2. the rest of the unit tests that were supposed to be added by now from this GH issue: https://github.com/cleanlab/cleanlab/issues/1055
in scenarios where the dataset is less than or greater than n=100

jwmueller (approved):
let's merge this and get the bigger np.unique improvements in

PR #1114, opened 2024-05-02
avoid sensitive divisions without ensuring denominator is not small

jwmueller:
want to avoid possible numeric instabilities, especially over edge-case datasets


jwmueller:
@elisno I haven't deeply probed the failing unit test, appreciate if you're able to take over:

Screen Shot 2024-05-02 at 2 43 17 PM

It was failing since the very first commit here, which merely did the following change:

Screen Shot 2024-05-02 at 2 44 29 PM

elisno (approved):

PR #1115, opened 2024-05-03
All identical datasets tests

### LINKED ISSUE #1055: add unit test with all identical dataset

elisno:
Add further test cases for an all-identical dataset.

This PR covers:

- Ensure that label quality scores for regression datasets work reasonably for all identical labels/targets.
- Verify that label issue detection behaves reasonably in Datalab for regression and classification.

Resolves #1055 

---

What issue types are tested against these kinds of datasets:

- [x] Label issues
- [x] Outlier issues
- [x] Near duplicate issues
- [x] Underperforming group issues (never any examples flagged, since there's usually just one example with a label error)
- [x] Class Imbalance issues (Only works for >20 example datasets, as the imbalance cutoff is 0.1*(1/K)=0.5 for a 2-class dataset.
- [ ] Non-IID issues (hard to reproduce, not a priority)
- [ ] Null issues (not a priority, all examples are generated)

### DIFF
--- a/tests/datalab/datalab/end_to_end_tests/test_all_identical_examples.py
+++ b/tests/datalab/datalab/end_to_end_tests/test_all_identical_examples.py
@@ -10,7 +10,13 @@ class TestAllIdenticalExamplesDataset:
         N, K = request.param
         # Create the dataset with all identical examples
         X = np.full((N, K), fill_value=np.random.rand(K))
-        return {"X": X}
+
+        # All labels for identical points should be the same
+        y = ["a"] * N
+        # One of the labels is different, so it should be flagged with a label issue
+        y[-1] = "b"
+
+        return {"X": X, "y": y}
 
     @pytest.fixture
     def dataset_with_one_unique_example(self, request):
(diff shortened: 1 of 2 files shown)


tests/datalab/datalab/end_to_end_tests/test_all_identical_examples.py
```diff
@@ -64,25 +123,216 @@ def test_issue_detection(self, dataset):
             atol=5e-16,
         )
 
+        label_issues = lab.get_issues("label")[["is_label_issue", "label_score"]]
+        expected_label_issues = pd.DataFrame(
+            [{"is_label_issue": False, "label_score": 1.0}] * (N - 1)
+            + [
+                {"is_label_issue": False, "label_score": 0.0}
```
elisno, line 130: Note that this point is not flagged as having a label issue, but the LQS is at a minimum.
It's the only point that disagrees with other points.

@jwmueller I suppose that is the expected behavior for a dataset:

```python
labels = ["Class A"] * 99 + ["Class B"]
pred_probs = np.array([[1.0, 0.0] * 100])  # All points predicted to be "Class A"
```

aditya1503 (approved):

PR #1117, opened 2024-05-04
Add a knn module

elisno:
# Summary

This PR prioritizes a refactoring of the two modules
- cleanlab/outlier.py
- cleanlab/datalab/internal/issue_manager/duplicate.py

that rely heavily on a knn object (`NearestNeighbors`) for scoring and flagging different kinds of issues.

MERGE THIS AFTER #1115 !

### Included docs

image

## More information

The logic for constructing a `NearestNeighbor` object has been duplicated in several places like so:

```python

# Make sure features exist to fit a knn object
if features is None:
    raise ValueError(
        "Both knn and features arguments cannot be None at the same time. Not enough information to compute outlier scores."
    )

# Select a good value for the number of neighbors
if k is None:
    k = DEFAULT_K  # use default when knn and k are both None
N, M = features.shape
if k > N:  # Ensure number of neighbors less than number of examples
    raise ValueError(
    )

# Select a decent distance metric
if M > 3:  # use euclidean distance for lower dimensional spaces
    metric = "cosine"
elif N > 100:  # Use efficient implementation (numerically unstable in edge cases)
    metric = "euclidean"
else:  # Use scipy implementation for precise results
    metric = euclidean

# Construct the knn object
knn = NearestNeighbors(n_neighbors=k, metric=metric).fit(features)
```

This PR reduces this headache to a single function call, as this is usually some boilerplate code for creating a knn-object from scratch.

```python
knn: NearestNeighbors = features_to_knn(features, n_neighbors=k)  # Optionally, specify a metric
```

--- 

Additionally, there are some dev docs for module added in the new directory `cleanlab/internal/neighbor`. Those are not a priority and can be omitted in this PR.


cleanlab/internal/neighbor/neighbor.py
```diff
@@ -0,0 +1,91 @@
...
+from sklearn.neighbors import NearestNeighbors
+
+if TYPE_CHECKING:
+    from cleanlab.internal.neighbor.types import FeatureArray, Metric
+
+from cleanlab.internal.neighbor.metric import decide_metric
+from cleanlab.internal.neighbor.search import construct_knn
+
+
+DEFAULT_K = 10
```
elisno, line 15: As the PR is only aimed at refactoring outlier-detection and near-duplicate detection. Currently, the outlier check uses DEFAULT_K as a default, while near-duplicate checks just need one neighbor (other than the query point).

cleanlab/internal/neighbor/neighbor.py
```diff
@@ -0,0 +1,91 @@
...
+    return csr_matrix((distances.reshape(-1), indices.reshape(-1), indptr), shape=(N, N))
+
+
+def _configure_num_neighbors(features: FeatureArray, k: Optional[int]):
+    if k is not None and k >= features.shape[0]:
+        raise ValueError(
+            f"Number of nearest neighbors k={k} cannot exceed the number of examples N={len(features)} passed into the estimator (knn)."
+        )
+
+    k = min(k or DEFAULT_K, features.shape[0] - 1)
```
elisno, line 90: Explaining this code a bit:

At this point in the codem, `k` can still be unspecified (`None`), so the default value is actually set in this line.

But the check above has only verified that if `k` wasn't None, then it must be smaller than the length of the dataset.
We still haven't done that check here.

So the right-hand side is ONLY chosen when the default value of `k` is equal to or larger than the length of the dataset.

```
k or DEFAULT_K

File: cleanlab/internal/neighbor/types.py
Comment (elisno) [inline cleanlab/internal/neighbor/types.py:1]: Just some helpful type aliases while developing the neighbor modules.

File: cleanlab/outlier.py [hunk 3]
```diff
@@ -424,30 +424,12 @@ def _get_ood_features_scores(
...
-            else:  # Use scipy implementation for precise results
-                metric = euclidean
-
-            knn = NearestNeighbors(n_neighbors=k, metric=metric).fit(features)
+            knn = features_to_knn(features, n_neighbors=k)
             features = None  # features should be None in knn.kneighbors(features) to avoid counting duplicate data points
-            distance_metric = metric if isinstance(metric, str) else str(metric.__name__)
+            distance_metric = (
+                metric if isinstance((metric := knn.metric), str) else str(metric.__name__)
+            )
```
elisno, line 431: Still need to "track" the distance metric to take care of some numerical precision corrections when we finally compute the scores.
The [Walrus-operator](https://realpython.com/python-walrus-operator/) (`:=`) comes somewhat in handy here to handle the weird distinction between a `metric: str` (sklearn implementations) and a `metric: Callable` (e.g. scipy- or custom implementations).

cleanlab/outlier.py
```diff
@@ -424,30 +424,12 @@ def _get_ood_features_scores(
...
+            k = knn.n_neighbors
```
elisno, line 432: This variable still needs to be set at the end for consistency in this function.

cleanlab/regression/rank.py
```diff
@@ -142,7 +141,7 @@ def _get_outre_score_for_each_label(
     *,
     residual_scale: float = 5,
     frac_neighbors: float = 0.5,
-    neighbor_metric: Optional[Union[str, Callable]] = "euclidean",
+    neighbor_metric: Optional[Union[str, Callable]] = None,
```
elisno, line 144: Looks like previous PRs updating the choice of metric implementations forgot to switch to the correct behavior.

cleanlab/regression/rank.py
```diff
@@ -181,12 +180,7 @@ def _get_outre_score_for_each_label(
     features = np.array([labels, residual]).T
 
     neighbors = int(np.ceil(frac_neighbors * labels.shape[0]))
-    if neighbor_metric is None:
-        if features.shape[0] > 100:
-            neighbor_metric = "euclidean"
-        else:
-            neighbor_metric = euclidean
-    knn = NearestNeighbors(n_neighbors=neighbors, metric=neighbor_metric).fit(features)
+    knn = features_to_knn(features, n_neighbors=neighbors, metric=neighbor_metric)
```
elisno, line 183: Because the `features` are 2d, then we can safely assume that `features_to_knn` will pick a suitable metric, even if `neighbor_metric = None`.

docs/source/cleanlab/internal/neighbor/index.rst
```diff
@@ -0,0 +1,21 @@
...
+.. automodule:: cleanlab.internal.neighbor
+   :autosummary:
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+.. toctree::
+    neighbor
+    metric
+    search
```
elisno, line 21: You can ignore most of these docs for now.

tests/datalab/datalab/end_to_end_tests/test_all_identical_examples.py
```diff
@@ -5,44 +5,103 @@
 
 
 class TestAllIdenticalExamplesDataset:
+    """There are 4 types of datasets this class tests:
```
elisno, line 8: These changes are introduced in #1115, not really part of this PR.

tests/internal/neighbor/test_metric.py
```diff
@@ -0,0 +1,31 @@
...
+def test_decide_metric_for_2d_and_3d_features(N):
+    # 2D and 3D features should always use the euclidean metric, disregarding the different implementations.
+    for M in [2, 3]:
+        X = np.random.rand(N, M)
+        metric = decide_metric(X)
+        if hasattr(metric, "__name__"):
+            error_msg = "The metric should be the string 'euclidean' for N > 100."
+            assert N <= 100, error_msg
+            metric = getattr(metric, "__name__")
+        assert metric == "euclidean"
```
elisno, line 20: Here's how we can be confident that 2d and 3d data use a Euclidean metric (although the implementation details can differ based on the dataset length).

tests/internal/neighbor/test_metric.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@pytest.mark.parametrize(
```
elisno, line 23: Here we test that by default, we prefer the cosine metric for "high" dimensional datasets.

tests/internal/neighbor/test_neighbor.py
```diff
@@ -0,0 +1,75 @@
...
+    else:
+        metric = knn.metric
+        if N <= 100:
+            assert hasattr(metric, "__name__")
+            metric = metric.__name__
+        assert metric == "euclidean"
+
+    # The knn object should be fitted to the features.
+    # TODO: This is not a good test, but it's the best we can do without exposing the internal state of the knn object.
+    # Assert is_
```
elisno, line 40:
```suggestion
```

elisno: Consider that this comment does not affect the CI results.

tests/internal/neighbor/test_neighbor.py
```diff
@@ -0,0 +1,75 @@
...
+def test_knn_kwargs():
+    """Check that features_to_knn passes additional keyword arguments to the NearestNeighbors constructor correctly."""
+    features = np.random.rand(100, 10)
+    V = features.var(axis=0)
+    knn = features_to_knn(features, n_neighbors=6, metric="seuclidean", metric_params={"V": V})
+
+    assert knn.n_neighbors == 6
+    assert knn.metric == "seuclidean"
+    assert knn._fit_X is features
+    assert knn.metric_params == {"V": V}
```
elisno, line 53: Basically, the `features_to_knn` function allows you to configure the NearestNeigbhors object as you please.

tests/internal/neighbor/test_neighbor.py
```diff
@@ -0,0 +1,75 @@
...
+    assert knn_graph.dtype == np.float64
+    assert np.all(knn_graph.data >= 0)
+    assert np.all(knn_graph.indices >= 0)
+    assert np.all(knn_graph.indices < 100)
+
+    distances = knn_graph.data.reshape(N, k)
+    indices = knn_graph.indices.reshape(N, k)
+
+    # Assert all rows in distances are sorted
+    assert np.all(np.diff(distances, axis=1) >= 0)
```
elisno, line 75: It's not easy in this refactor to move the construction of a `knn_graph` sparse matrix into something like `features_to_knn`. Keeping it in one function may make it easier for us in the future to "correct" the knn_graph that we build by default.

tests/test_regression.py
```diff
@@ -243,3 +243,25 @@ def test_save_space():
...
+    predictions[-cutoff_index:] += 1
+
+    scores = get_label_quality_scores(labels=labels, predictions=predictions, method=method)
+    np.testing.assert_allclose(scores[:-cutoff_index], 1, atol=1e-04)
+    if method == "outre":
+        # Assert that the scores for the last (bad) ~quarter of examples are close to 0
+        np.testing.assert_allclose(scores[-cutoff_index:], 0, atol=1e-04)
+    else:
+        # Residual method should give "imperfect" scores for the last ~quarter of examples, but not necessarily near 0
+        assert np.all(scores[-cutoff_index:] < 1)
```
elisno, line 267: This change is a part of #1115, ignore in this PR.

jwmueller:
Closes https://github.com/cleanlab/cleanlab/issues/889

cleanlab/internal/neighbor/neighbor.py
```diff
@@ -0,0 +1,91 @@
+from __future__ import annotations
```
jwmueller, line 1: [nit] name of this file is a bit vague, given it's in the neighbor/ subfolder already. Could consider knn_graph.py or knearestneighbors.py something more explicit like that

cleanlab/datalab/internal/issue_manager/duplicate.py
```diff
@@ -21,12 +21,11 @@
 import numpy as np
```
jwmueller, line 21: There's no code changes needed in other issue managers like [Underperforming Group](https://github.com/cleanlab/cleanlab/blob/master/cleanlab/datalab/internal/issue_manager/underperforming_group.py) or [Data Valuation](https://github.com/cleanlab/cleanlab/blob/master/cleanlab/datalab/internal/issue_manager/data_valuation.py) ?

It seems like those files heavily rely on KNN graph, so that's a bit surprising they don't need to be updated

jwmueller: If they do need to be updated, please do so in this PR, in order to comprehensively test this refactored code works properly across the use-cases it needs to support.

cleanlab/internal/neighbor/neighbor.py
```diff
@@ -0,0 +1,91 @@
...
+    # Decide the number of neighbors to use in the KNN search.
+    n_neighbors = _configure_num_neighbors(features, n_neighbors)
+
+    knn = construct_knn(n_neighbors, metric, **sklearn_knn_kwargs)
+    return knn.fit(features)
+
+
+def knn_to_knn_graph(knn: NearestNeighbors) -> csr_matrix:
+    distances, indices = knn.kneighbors(return_distance=True)
+
```
jwmueller, line 74: please add a one-liner docstring explaining what is the purpose of this method. Also consider renaming the method such that it's purpose is more clear

cleanlab/internal/neighbor/neighbor.py
jwmueller: can you make a one-liner docstring with this explanation in it?

generally if code warrants an explanation, there should be a docstring and code comments, so the explanation remains preserved years from now

cleanlab/internal/neighbor/neighbor.py
jwmueller: please add a comment next to this constant (same for all constants we ever introduce) explaining it.

jwmueller: I thought Datalab more generally uses `DEFAULT_K` to construct its KNN graph. We are only ever constructing a single KNN graph, that will be reused across all issue-managers that are based on KNN: Outliers, Duplicates, Data Valuation, Underperforming Group, etc.

cleanlab/internal/neighbor/metric.py
```diff
@@ -0,0 +1,68 @@
...
+"""
+ROW_COUNT_CUTOFF: int = 100
+"""
+The cutoff value for the number of rows in the feature array. 
+
+If the number of rows (N) in the feature array is greater than this cutoff value,
+the "euclidean" metric is used (implemented in scikit-learn for efficiency reasons).
+Otherwise, the implementation of the euclidean metric from scipy is used, which is much slower but
+more accurate in general.
+"""
```
jwmueller, line 22:
```suggestion
HIGH_DIMENSION_CUTOFF: int = 3
"""
If the number of columns (M) in the `features` array is greater than this cutoff value,
then by default K-nearest-neighbor operations will use the "cosine" metric. The cosine metric is more suitable for high-dimensional data.
"""
ROW_COUNT_CUTOFF: int = 100
"""
The cutoff value for the number of rows in the feature array. 

If the number of rows (N) in the feature array is greater than this cutoff value,
the "euclidean" metric is used (implemented in scikit-learn for efficiency reasons).
Otherwise, the implementation of the euclidean metric from scipy is used, which is much slower but
more accurate in general.
"""
```

cleanlab/internal/neighbor/metric.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def decide_metric(features: np.ndarray) -> Metric:
+    """
+    Decide the metric to be used based on the shape of the feature array.
```
jwmueller, line 27:
```suggestion
    Decide the KNN metric to be used based on the shape of the feature array.
```

cleanlab/internal/neighbor/search.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+if TYPE_CHECKING:
+
+    from cleanlab.internal.neighbor.types import Metric
+
+
+def construct_knn(n_neighbors: int, metric: Metric, **knn_kwargs) -> NearestNeighbors:
+    """
+    Constructs a k-nearest neighbors search object.
```
jwmueller, line 14:
```suggestion
    Constructs a k-nearest neighbors search object. You can implement a similar method to run cleanlab with your own approximate-KNN library.
```

cleanlab/internal/neighbor/types.py
```diff
@@ -0,0 +1,17 @@
+from typing import Callable, Union
```
jwmueller, line 1: Can you move these definitions to the existing file here?

https://github.com/cleanlab/cleanlab/blob/master/cleanlab/typing.py

cleanlab/regression/rank.py
jwmueller: what do you mean?  This change here is INCORRECT.

The metric in this get_outre_score() method should always be the Euclidean distance. Whether to use your "euclidean"-sklearn version or euclidean-scipy version I'll leave up to you. But this method should never be using cosine ever

jwmueller: you could be relying on the fact that internally the features corresponding to this method will never exceed 3-D and thus it will automatically use Euclidean distance, but that is a pretty dangerous implicit assumption. If you want to rely on this assumption, it needs to be explicitly commented in the code/documentation, so that it's not forgotten in the future

cleanlab/outlier.py
jwmueller: can you add (shorter version of) this explanation as a code comment here?

cleanlab/datalab/internal/issue_manager/duplicate.py
jwmueller: also none of the code in the outlier issue manager needs to change at all?  I didn't look at it, just surprised no change is needed.

cleanlab/internal/neighbor/search.py
```diff
@@ -0,0 +1,37 @@
...
+        The number of nearest neighbors to consider.
+    metric :
+        The distance metric to use for computing distances between points.
+        See :py:mod:`~cleanlab.internal.neighbor.metric` for more information.
+    **knn_kwargs:
+        Additional keyword arguments to be passed to the search index constructor.
+        See https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.NearestNeighbors.html for more details on the available options.
+
+    Returns:
+        NearestNeighbors: A k-nearest neighbors search object.
```
jwmueller, line 28: where is the single source of truth across the entire library that dictates what attributes and what properties this NearestNeighbors object fundamentally must possess?

If there isn't a place yet, let's enumerate this here, and now before we forget to.  That knowledge would be crucial for changing the default KNN library in the future, for others to use their own KNN library, and for future devs adding issue-types that rely on KNN to understand what properties they are guaranteed to be able to rely on.

I assume the properties are simply a subset of the NearestNeighbors object that sklearn returns, much like CleanLearning can be run on any classifier that implements a subset of the properties of sklearn's base classifier object.

tests/internal/neighbor/test_neighbor.py
```diff
@@ -0,0 +1,75 @@
...
+        metric = knn.metric
+        if N <= 100:
+            assert hasattr(metric, "__name__")
+            metric = metric.__name__
+        assert metric == "euclidean"
+
+    # The knn object should be fitted to the features.
+    # TODO: This is not a good test, but it's the best we can do without exposing the internal state of the knn object.
+    # Assert is_
+    assert knn._fit_X is features
```
jwmueller, line 41: could we have more tests here that check:

each attributes and each properties the NearestNeighbors object fundamentally must possess.

Even if we know these tests will trivially pass for sklearn today, it could be very useful to have them if sklearn updates, or we eventually use a different KNN library than sklearn.

tests/internal/neighbor/test_neighbor.py
```diff
@@ -0,0 +1,75 @@
...
+    ids=lambda x: f"N={x}",
+)
+@pytest.mark.parametrize(
+    "M",
+    [2, 3, 4, 5, 10, 50, 100],
+    ids=lambda x: f"M={x}",
+)
+def test_features_to_knn(N, M):
+
+    features = np.random.rand(N, M)
```
jwmueller, line 22: can we make the last 10 entries of features all identical, just to include an edge-case in this test already?

jwmueller (changes requested):
left various comments

cleanlab/regression/rank.py
elisno: I've now ensured that this function picks a euclidean metric.

cleanlab/internal/neighbor/search.py
elisno: I've now listed the necessary methods, the standard attributes as well as some some additional information about fitted attributes (those that are only set after calling `NearestNeighbors.fit()`)

tests/internal/neighbor/test_neighbor.py
elisno: I've added as much as I can at this stage.

cleanlab/datalab/internal/issue_manager/data_valuation.py
```diff
@@ -139,32 +142,11 @@ def find_issues(
...
-                )
-            if self.metric is None:
-                self.metric = (
-                    "cosine"
-                    if features.shape[1] > 3
-                    else "euclidean" if features.shape[0] > 100 else euclidean
-                )
-            knn = NearestNeighbors(n_neighbors=self.k, metric=self.metric).fit(features)
-
-            if self.metric and self.metric != knn.metric:
```
elisno, line 154: This condition won't evaluate to `True`, because even if the metric changed, then the `knn` object must have been given the same metric in this part of the code.

Removing it is safe.

cleanlab/internal/neighbor/neighbor.py
elisno: Most types use the default value of 10, but each issue type happens to set the same default value.

In any case, I've added a comment.

cleanlab/internal/neighbor/knn_graph.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+DEFAULT_K = 10  # Value is set for issue type that requires the largest number of neighbors. Most of the issue types require 10 neighbors by default.
+"""Default number of neighbors to consider in the k-nearest neighbors search,
+unless the size of the feature array is too small or the user specifies a different value.
+
+This should be the largest desired value of k for all desired issue types that require a KNN graph.
+
+E.g. if near duplicates wants k=1 but outliers wants 10, then DEFAULT_K should be 10.
+"""
```
jwmueller, line 22:
```suggestion
DEFAULT_K = 10
"""Default number of neighbors to consider in the k-nearest neighbors search,
unless the size of the feature array is too small or the user specifies a different value.

This should be the largest desired value of k for all desired issue types that require a KNN graph.

E.g. if near duplicates wants k=1 but outliers wants 10, then DEFAULT_K should be 10. This way, all issue types can rely on the same KNN graph.
"""
```
```suggestion
DEFAULT_K = 10  # Value is set for issue type that requires the largest number of neighbors. Most of the issue types require 10 neighbors by default.
"""Default number of neighbors to consider in the k-nearest neighbors search,
unless the size of the feature array is too small or the user specifies a different value.

This should be the largest desired value of k for all desired issue types that require a KNN graph.

E.g. if near duplicates wants k=1 but outliers wants 10, then DEFAULT_K should be 10.
"""
```

cleanlab/internal/neighbor/knn_graph.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # Decide the number of neighbors to use in the KNN search.
+    n_neighbors = _configure_num_neighbors(features, n_neighbors)
+
+    knn = construct_knn(n_neighbors, metric, **sklearn_knn_kwargs)
+    return knn.fit(features)
+
+
+def construct_knn_graph_from_index(knn: NearestNeighbors) -> csr_matrix:
+    """Construct a KNN graph from a fitted NearestNeighbors search object.
```
jwmueller, line 73:
```suggestion
    """Construct a sparse distance matrix representation of KNN graph out of a fitted NearestNeighbors search object.
```

cleanlab/datalab/internal/issue_manager/duplicate.py
```diff
@@ -76,32 +75,11 @@ def find_issues(
...
-            self.metric = knn.metric
-
-            try:
-                check_is_fitted(knn)
-            except NotFittedError:
-                knn.fit(features)
-
-            knn_graph = knn.kneighbors_graph(mode="distance")
+            _features = cast(np.ndarray, features)
+            self.metric = self.metric or decide_default_metric(_features)
```
jwmueller, line 79: it seems like this:
`self.metric = self.metric or decide_default_metric(_features)`
is always being used before `construct_knn_graph_from_features()` 

in every IssueManager?  Why not move this line of code inside of the `construct_knn_graph_from_features()` function?

jwmueller: same with the `cast(np.ndarray, features)` call, which appears to be happening in many of the IssueManagers, and could perhaps be inside of the `construct_knn_graph_from_features()` function. I'm not sure if the cast is supposed to have longer-term effects on `features` though that persist outside of this function...

elisno: The `self.metric = ...` turned out to be necessary when `construct_knn_graph_from_features()` only returned a `csr_matrix` (a sparse matrix), while most of the issue managers wanted to track the metric that was set for the `knn` object.

The line:

```python
self.metric = self.metric or decide_default_metric(_features)
```
was already inside of the `construct_knn_graph_from_features()` (specifically the `features_to_knn()` function that constructed the `knn` object).

I've now updated the function to return both the spare matrix as well as the `knn` metric, which we can assume to have a `metric` attribute.

---

Regarding the `cast()`, it has no runtime impact, just a way to ensure the type-checker that at this point in the code, that features is considered a `np.ndarray`. This is now redundant after I removed `self.metric` lines, because `construct_knn_graph_from_features()` properly raises a runtime error when `features = None`.

cleanlab/internal/neighbor/metric.py
```diff
@@ -0,0 +1,107 @@
...
+def _euclidean_large_dataset(features: FeatureArray):
+    return "euclidean"
+
+
+def _euclidean_small_dataset(features: FeatureArray):
+    return euclidean
+
+
+def _cosine_metric(features: FeatureArray):
+    return "cosine"
```
elisno, line 34:
```suggestion
# Metric decision functions
def _euclidean_large_dataset() -> str:
    return "euclidean"

def _euclidean_small_dataset() -> Metric:
    return euclidean

def _cosine_metric() -> str:
    return "cosine"
```

cleanlab/internal/neighbor/metric.py
```diff
@@ -0,0 +1,107 @@
...
+    --------
+    ROW_COUNT_CUTOFF: The cutoff value for the number of rows in the feature array.
+    sklearn.metrics.pairwise.euclidean_distances: The euclidean metric function from scikit-learn.
+    scipy.spatial.distance.euclidean: The euclidean metric function from scipy.
+    """
+    num_rows = features.shape[0]
+    if num_rows > ROW_COUNT_CUTOFF:
+        return _euclidean_large_dataset(features)
+    else:
+        return _euclidean_small_dataset(features)
```
elisno, line 70:
```suggestion
    if num_rows > ROW_COUNT_CUTOFF:
        return _euclidean_large_dataset()
    else:
        return _euclidean_small_dataset()
```

cleanlab/internal/neighbor/metric.py
```diff
@@ -0,0 +1,107 @@
...
+If the number of rows (N) in the `features` array is greater than this cutoff value,
+then by default, Euclidean distances are computed via the "euclidean" metric
+(implemented in sklearn for efficiency reasons).
+Otherwise, Euclidean distances are by default computed via
+the ``euclidean`` metric from scipy (slower but numerically more precise/accurate).
+"""
+
+
+# Metric decision functions
+def _euclidean_large_dataset(features: FeatureArray):
```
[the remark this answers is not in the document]
elisno: Thanks for the feedback! I've removed the unused `features` parameter from the internal metric decision functions to improve clarity. The parameter is now only used in the main decision functions where it's necessary for consistency.

elisno:
I've addressed the comments made by @jwmueller. @HuiWen could you review + merge this?

cleanlab/internal/neighbor/metric.py
```diff
@@ -0,0 +1,105 @@
...
+    Otherwise, a euclidean metric is used.
+    That is handled by the :py:meth:`~cleanlab.internal.neighbor.metric.decide_euclidean_metric` function.
+
+    See Also
+    --------
+    HIGH_DIMENSION_CUTOFF: The cutoff value for the number of columns in the feature array.
+    sklearn.metrics.pairwise.cosine_distances: The cosine metric function from scikit-learn
+    """
+    if features.shape[1] > HIGH_DIMENSION_CUTOFF:
+        return _cosine_metric(features)
```
huiwengoh, line 104: `_cosine_metric` above has no expected arg

elisno, line 104:
```suggestion
        return _cosine_metric()
```

huiwengoh (approved):
lgtm! nice work

PR #1119, opened 2024-05-06
Improve KNN Graph Construction for Handling Exact Duplicates and Numerical Precision

elisno:
### Summary

This PR introduces a new argument, `correct_exact_duplicates: bool = True`, to the function `construct_knn_graph_from_features` and others. This update ensures that exact duplicates in the feature array are handled correctly during the k-nearest neighbors (KNN) graph construction.

#### New Functions

- **`correct_knn_graph`**:
  Wrapper around `correct_knn_distances_and_indices` to correct KNN graph based on the feature array.
  ```python
  def correct_knn_graph(features: FeatureArray, knn_graph: csr_matrix) -> csr_matrix:
      ...
  ```
  
- **`correct_knn_distances_and_indices_with_exact_duplicate_sets_inplace`**:
  Main logic for in-place correction of distances and indices arrays.
  ```python
  def correct_knn_distances_and_indices_with_exact_duplicate_sets_inplace(
      distances: np.ndarray,
      indices: np.ndarray,
      exact_duplicate_sets: List[np.ndarray],
  ) -> None:
      ...
  ```

- **`correct_knn_distances_and_indices`**:
  Corrects KNN distances and indices with optional exact duplicate sets and warning.
  ```python
  def correct_knn_distances_and_indices(
      features: FeatureArray,
      distances: np.ndarray,
      indices: np.ndarray,
      exact_duplicate_sets: Optional[List[np.ndarray]] = None,
      enable_warning: bool = False,
  ) -> tuple[np.ndarray, np.ndarray]:
      ...
  ```

### Impact

- The default behavior of most functions now includes correction for exact duplicates unless a `knn: NearestNeighbors` object is passed explicitly.
- This change affects the outlier detection in `cleanlab/outliers.py`, where correction is applied manually if no `knn` object is provided.
- The `noniid` check in Datalab is updated to construct a KNN graph without relying on `NearestNeighbors` from sklearn.

### What this PR does *not* address:

- The correction logic addresses exact duplicates but does not explicitly cover scenarios where there are near-duplicates or small variations in features that might need similar handling.
- Performance optimization for large duplicated datasets. Users with large datasets and numerous sets of exact duplicates might experience slower performance due to the iteration across all the different duplicate sets.
- The PR adds corrections for exact duplicates during KNN graph construction on training data, but it does not provide flexibility for other k nearest neighbor search libraries. Such graphs should be constructed by the user, and subsequently corrected with the same features.
  - The same applies to correcting knn graphs on test data.

### Benchmark Results

Two benchmark scenarios were tested:

1. **All-Identical Dataset**:
    - One unique point duplicated N times.

2. **Copied Dataset**:
    - Several points duplicated a few times.

The graphs below compare runtime and memory usage for different functions:
- **Top Graph**: Runtime vs. Number of Points
- **Bottom Graph**: Memory Usage vs. Number of Points

For the  **All-Identical Dataset**, the correction function spends its time constructing a small circulant matrix to find the nearest neighbors of the first k+1 elements. All other points just refer to the first k points. The purple line shows how the correction algorithm performs if all the duplicate information is pre-computed and the output can be modified in-place. No knn-graph construction occurs in that function.

image

For the **Copied Dataset**, there are far more sets to iterate over, which impacts the performance of the correction function.
This shouldn't really exceed the runtime of the exhaustive search algorithm by too much.

image

The benchmark code and additional results are provided in the expandable sections.

### Benchmark Code for All-Identical Dataset

<details><summary>Code for All-Identical Dataset</summary>
<p>

```python
from __future__ import annotations
import time
import tracemalloc
import numpy as np
import pandas as pd
from tqdm.auto import tqdm

from cleanlab.internal.neighbor.knn_graph import (
    _compute_exact_duplicate_sets,
    correct_knn_distances_and_indices,
    correct_knn_distances_and_indices_with_exact_duplicate_sets_inplace,
    features_to_knn,
    construct_knn_graph_from_index,
    create_knn_graph_and_index,
)

# Define the sizes of feature arrays for the benchmark
feature_sizes = np.logspace(1, 7.0, num=25, base=10, dtype=int)

# Define a function to benchmark the memory and runtime of a given function
def benchmark_function(func, *args, **kwargs):
    # Record the start time and memory usage
    start_time = time.time()
    tracemalloc.start()
    
    # Run the function
    result = func(*args, **kwargs)
    
    # Record the end time and memory usage
    end_time = time.time()
    current, peak = tracemalloc.get_traced_memory()
    tracemalloc.stop()
    
    # Calculate runtime and peak memory usage
    runtime = end_time - start_time
    peak_memory = peak / 10**6  # Convert to MB
    
    return runtime, peak_memory, result

# Initialize a DataFrame to store the benchmark results
columns = [f_col:='Function', N_col:='Num points', runtime_col:='Runtime (s)', memory_col:='Memory (MB)']
results = pd.DataFrame(columns=columns)

# Define functions to benchmark
functions = {
    # 'features_to_knn': features_to_knn,
    'construct_knn_graph_from_index_without_correction': construct_knn_graph_from_index,
    'construct_knn_graph_from_index_with_correction': construct_knn_graph_from_index,
    # 'create_knn_graph_and_index': create_knn_graph_and_index,
    'correct_knn_distances_and_indices': correct_knn_distances_and_indices,
    'correct_with_precomputed_exact_duplicate_sets': correct_knn_distances_and_indices,
    'correct_with_precomputed_exact_duplicate_sets_inplace': correct_knn_distances_and_indices_with_exact_duplicate_sets_inplace,
}
results_list_of_dicts = []
# Run the benchmark for each function and feature size
N_max_slow = 20000
for N in tqdm(feature_sizes):
    # features = np.random.rand(N, 10)  # Generate random feature array
    features = np.ones((N, 10))
    if N > N_max_slow:
        k = min(10, N-1)
        distances = np.tile(np.ones(k), (N, 1))
        indices = np.tile(np.arange(k), (N, 1))
        exact_duplicate_sets = [np.arange(N)]
    else:

        knn_graph, knn = create_knn_graph_and_index(features, correct_exact_duplicates=False)
        distances = knn_graph.data.reshape(knn_graph.shape[0], -1)
    
        indices = knn_graph.indices.reshape(knn_graph.shape[0], -1)

        exact_duplicate_sets = _compute_exact_duplicate_sets(features)
    
    for func_name, func in functions.items():
        if func_name == 'construct_knn_graph_from_index_without_correction':
            if N > N_max_slow:
                continue
            runtime, peak_memory, _ = benchmark_function(func, knn)
        elif func_name == 'construct_knn_graph_from_index_with_correction':
            if N > N_max_slow:
                continue
            runtime, peak_memory, _ = benchmark_function(func, knn, correct_exact_duplicates=True)
        elif func_name == 'create_knn_graph_and_index':
            if N > N_max_slow:
                continue
            runtime, peak_memory, _ = benchmark_function(func, features)
        elif func_name == 'correct_knn_distances_and_indices':
            if N > N_max_slow:
                continue
            runtime, peak_memory, _ = benchmark_function(func, features=features, distances=distances, indices=indices)
        elif func_name == 'correct_with_precomputed_exact_duplicate_sets':
            runtime, peak_memory, _ = benchmark_function(func, features=features, distances=distances, indices=indices, exact_duplicate_sets=exact_duplicate_sets)
        elif func_name == 'correct_with_precomputed_exact_duplicate_sets_inplace':
            runtime, peak_memory, _ = benchmark_function(func, distances=distances, indices=indices, exact_duplicate_sets=exact_duplicate_sets)
        else:
            if N > 1000:
                continue
            runtime, peak_memory, _ = benchmark_function(func, features)
        
        
        # Store the results in the DataFrame
        results_list_of_dicts = results_list_of_dicts + [dict({
            f_col: func_name,
            N_col: N,
            runtime_col: runtime,
            memory_col: peak_memory,
        })]

results = pd.DataFrame(results_list_of_dicts)
# Save the results to a CSV file
results.to_csv('benchmark_results.csv', index=False)

# Print the results
print(results)

# Plot the results
import matplotlib.pyplot as plt

fig, axes = plt.subplots(2, 1, figsize=(10, 10))
for func_name in functions.keys():
    subset = results[results[f_col] == func_name]
    axes[0].plot(subset[N_col], subset[runtime_col], label=func_name)
    axes[1].plot(subset[N_col], subset[memory_col], label=func_name)

axes[0].set_xlabel(N_col)
axes[0].set_ylabel(runtime_col)
axes[0].set_title('Runtime vs. Num Points')
axes[0].legend()
axes[0].set_xscale('log')
axes[0].set_yscale('log')

axes[1].set_xlabel(N_col)
axes[1].set_ylabel(memory_col)
axes[1].set_title('Memory Usage vs. Num Points')
axes[1].legend()
axes[1].set_xscale('log')
axes[1].set_yscale('log')

plt.tight_layout()
plt.show()

``` 

</p>
</details>

### Benchmark Code for Copied Dataset

<details><summary>Code for Copied Dataset</summary>
<p>

```python
from __future__ import annotations
import time
import tracemalloc
import numpy as np
import pandas as pd
from memory_profiler import memory_usage
from tqdm.auto import tqdm

from cleanlab.internal.neighbor.knn_graph import (
    _compute_exact_duplicate_sets,
    correct_knn_distances_and_indices,
    correct_knn_distances_and_indices_with_exact_duplicate_sets_inplace,
    features_to_knn,
    construct_knn_graph_from_index,
    create_knn_graph_and_index,
)

# Define the sizes of feature arrays for the benchmark
feature_sizes = np.logspace(1, 4.6, num=25, base=10, dtype=int)

# Define a function to benchmark the memory and runtime of a given function
def benchmark_function(func, *args, **kwargs):
    # Record the start time and memory usage
    start_time = time.time()
    tracemalloc.start()
    
    # Run the function
    result = func(*args, **kwargs)
    
    # Record the end time and memory usage
    end_time = time.time()
    current, peak = tracemalloc.get_traced_memory()
    tracemalloc.stop()
    
    # Calculate runtime and peak memory usage
    runtime = end_time - start_time
    peak_memory = peak / 10**6  # Convert to MB
    
    return runtime, peak_memory, result

# Initialize a DataFrame to store the benchmark results
columns = [f_col:='Function', N_col:='Num points', runtime_col:='Runtime (s)', memory_col:='Memory (MB)']
results = pd.DataFrame(columns=columns)

# Define functions to benchmark
functions = {
    # 'features_to_knn': features_to_knn,
    'construct_knn_graph_from_index_without_correction': construct_knn_graph_from_index,
    'construct_knn_graph_from_index_with_correction': construct_knn_graph_from_index,
    # 'create_knn_graph_and_index': create_knn_graph_and_index,
    'correct_knn_distances_and_indices': correct_knn_distances_and_indices,
    'correct_with_precomputed_exact_duplicate_sets': correct_knn_distances_and_indices,
    'correct_with_precomputed_exact_duplicate_sets_inplace': correct_knn_distances_and_indices_with_exact_duplicate_sets_inplace,
}
results_list_of_dicts = []
# Run the benchmark for each function and feature size
N_max_slow = 20000
num_copies = 5
for N in tqdm(feature_sizes):

    features = np.random.rand(N // num_copies, 10)
    features = np.vstack([features] * num_copies)
    


    N = features.shape[0]

    knn_graph, knn = create_knn_graph_and_index(features, correct_exact_duplicates=False)
    distances = knn_graph.data.reshape(knn_graph.shape[0], -1)

    indices = knn_graph.indices.reshape(knn_graph.shape[0], -1)

    exact_duplicate_sets = _compute_exact_duplicate_sets(features)
    
    for func_name, func in functions.items():
        if func_name == 'construct_knn_graph_from_index_without_correction':
            if N > N_max_slow:
                continue
            runtime, peak_memory, _ = benchmark_function(func, knn)
        elif func_name == 'construct_knn_graph_from_index_with_correction':
            if N > N_max_slow:
                continue
            runtime, peak_memory, _ = benchmark_function(func, knn, correct_exact_duplicates=True)
        elif func_name == 'create_knn_graph_and_index':
            if N > N_max_slow:
                continue
            runtime, peak_memory, _ = benchmark_function(func, features)
        elif func_name == 'correct_knn_distances_and_indices':
            if N > N_max_slow:
                continue
            runtime, peak_memory, _ = benchmark_function(func, features=features, distances=distances, indices=indices)
        elif func_name == 'correct_with_precomputed_exact_duplicate_sets':
            runtime, peak_memory, _ = benchmark_function(func, features=features, distances=distances, indices=indices, exact_duplicate_sets=exact_duplicate_sets)
        elif func_name == 'correct_with_precomputed_exact_duplicate_sets_inplace':
            runtime, peak_memory, _ = benchmark_function(func, distances=distances, indices=indices, exact_duplicate_sets=exact_duplicate_sets)
        else:
            if N > 1000:
                continue
            runtime, peak_memory, _ = benchmark_function(func, features)
        
        
        # Store the results in the DataFrame
        results_list_of_dicts = results_list_of_dicts + [dict({
            f_col: func_name,
            N_col: N,
            runtime_col: runtime,
            memory_col: peak_memory,
        })]

results = pd.DataFrame(results_list_of_dicts)
# Save the results to a CSV file
results.to_csv('benchmark_results_with_dataset_copy.csv', index=False)

# Print the results
print(results)

# Plot the results
import matplotlib.pyplot as plt

fig, axes = plt.subplots(2, 1, figsize=(10, 10))
for func_name in functions.keys():
    subset = results[results[f_col] == func_name]
    axes[0].plot(subset[N_col], subset[runtime_col], label=func_name)
    axes[1].plot(subset[N_col], subset[memory_col], label=func_name)

axes[0].set_xlabel(N_col)
axes[0].set_ylabel(runtime_col)
axes[0].set_title('Runtime vs. Num Points')
axes[0].legend()
axes[0].set_xscale('log')
axes[0].set_yscale('log')

axes[1].set_xlabel(N_col)
axes[1].set_ylabel(memory_col)
axes[1].set_title('Memory Usage vs. Num Points')
axes[1].legend()
axes[1].set_xscale('log')
axes[1].set_yscale('log')

plt.tight_layout()
plt.show()
``` 

</p>
</details>


cleanlab/outlier.py
```diff
@@ -470,7 +455,16 @@ def _get_ood_features_scores(
...
+        distances, indices = knn.kneighbors(features)
+        if correct_knn:  # This should only happen if knn is None at the start of this function
+            _features_for_correction = (
+                knn._fit_X if features is None else features
+            )  # Hacky way to get features (training or test). Storing np.unique results is a hassle.
+            distances, _ = correct_knn_distances_and_indices(
+                features=_features_for_correction,
+                distances=distances,
+                indices=indices,
+            )
```
jwmueller, line 467: Why isn't the correction happening inside of the helper method: `features_to_knn()` or `construct_knn_graph_from_index()` ?  Shouldnt the correction be happening inside of these helpers, so that this correction consistently happens anytime the KNN graph is constructed?

Consider what happens if user runs Datalab sequentially one issue-type at a time. Say they do for one Datalab in order A: outliers, duplicates, non-iid

Then they do for another Datalab in order B: duplicates, non-iid, outliers. 

Isn't it important that the KNN construction and data point distances are consistently represented, so that the end behavior is correct regardless of order A vs B?

jwmueller: or are you saying here it has to be done this way to support test data?
Not sure why that'd be, since for test data we'd never be in situation where `knn` is None, right

cleanlab/internal/neighbor/knn_graph.py
```diff
@@ -0,0 +1,335 @@
...
+
+    # Use np.unique to catch inverse indices of all unique feature sets
+    unique_inverse = np.unique(features, return_inverse=True, axis=0)[1]
+
+    # Map each unique feature set to its indices across the dataset
+    feature_map = {u: np.where(unique_inverse == u)[0] for u in set(unique_inverse)}
+
+    points_missing_exact_duplicates = []
+
+    for i, (dists, inds) in enumerate(zip(distances, indices)):
```
jwmueller, line 274: Why is the loop here over all distances and indices? 
Don't we only need to loop over the exact duplicate sets returned by np.unique ? 

I thought the correction algorithm would be efficient for realistic datasets where np.unique doesnt return many exact duplicates, but this algorithm seems to scale in the order of the dataset itself rather than just how many duplicates there are.

cleanlab/internal/neighbor/knn_graph.py
```diff
@@ -0,0 +1,335 @@
...
+
+    # Prepare the output arrays
+    corrected_distances = np.zeros_like(distances)
+    corrected_indices = np.zeros_like(indices, dtype=int)
+
+    # Use np.unique to catch inverse indices of all unique feature sets
+    unique_inverse = np.unique(features, return_inverse=True, axis=0)[1]
+
+    # Map each unique feature set to its indices across the dataset
+    feature_map = {u: np.where(unique_inverse == u)[0] for u in set(unique_inverse)}
```
jwmueller, line 270: could we name this variable more explicitly, like: `exact_duplicate_map` ?

cleanlab/outlier.py
elisno: Accessing `knn._fit_X` directly is a short-term solution, but moving the correction logic inside `construct_knn_graph_from_index(knn: Nearestneighbors)` may be acceptable for now.  I'd suggest passing `correct_knn` as a parameter to make this correction opt-in, ensuring it only happens when necessary, such as when `features` are missing but need correction via the `knn` object.

jwmueller: yes I think the logic needs to be moved inside the method, otherwise it seems too bug prone as users call Datalab in different issue-type orders and in different ways. Having an opt-in arg for the method seems fine

cleanlab/internal/neighbor/knn_graph.py
elisno: Switched to a loop over the exact duplicate sets returned by `np.unique`.

cleanlab/internal/neighbor/knn_graph.py
elisno: Switched from a map to a list of `exact_duplicate_sets`.

cleanlab/outlier.py
elisno: Logic is now in the helper functions, but the code in `cleanlab/outlier.py` is an exception, as that code is fundamentally different from other places in the code-base that rely on a `knn_graph: csr_matrix`.

The outlier issue check in `Datalab`  handle's this correctly and is the recommended way for handling outliers efficiently, even with corrections for exact duplicates.

elisno:
@huiwengoh I've addressed all of @jwmueller's comments.

Can you give a review and merge this?

elisno:
I've added some graphs that show the runtime of the knn graph construction in the library.

The main conclusion is that:

- As the dataset increases in size, the knn search takes the longest time, and the runtime when correcting for exact duplicates gets amortized.
- The green line shows what the additional runtime is when correcting a knn_graph for exact duplicates.
  - It involves calling `np.unique` on a feature array, but also makes some copies of the output arrays `distances` and `indices`.
- The purple line shows how long it takes to run the core-correction logic.
  - For a single exact duplicate set, we have a fully optimized algorithm that runs many orders of magnitude faster than running the actual knn search.
  - As the number of exact duplicate sets increases, the runtime seems to approach the runtime complexity of the knn search (at least for ~1000 data points.

cleanlab/outlier.py
```diff
@@ -453,7 +455,18 @@ def _get_ood_features_scores(
         # Get distances to k-nearest neighbors Note that the knn object contains the specification of distance metric
         # and n_neighbors (k value) If our query set of features matches the training set used to fit knn, the nearest
         # neighbor of each point is the point itself, at a distance of zero.
-        distances, _ = knn.kneighbors(features)
+        distances, indices = knn.kneighbors(features)
+        if (
+            correct_knn
+        ):  # This should only happen if knn is None at the start of this function. Will NEVER happen for approximate KNN.
```
jwmueller, line 461:
```suggestion
        ):  # This should only happen if knn is None at the start of this function. Will NEVER happen for approximate KNN provided by user.
```

cleanlab/internal/neighbor/knn_graph.py
```diff
@@ -106,6 +120,12 @@ def construct_knn_graph_from_index(knn: NearestNeighbors) -> csr_matrix:
 
     distances, indices = knn.kneighbors(return_distance=True)
 
+    if correct_exact_duplicates:
+        # NOTE: Accessing the features used to fit the knn object is a short-term solution.
+        # WARNING: Accessing an object's private attribute from a third-party library is
```
jwmueller, line 125: seems like incomplete statement?

cleanlab/internal/neighbor/knn_graph.py
```diff
@@ -106,6 +120,12 @@ def construct_knn_graph_from_index(knn: NearestNeighbors) -> csr_matrix:
...
+        _features = knn._fit_X
```
jwmueller, line 126: why do you want to have such a dangerous line of code inside of `construct_knn_graph_from_index()`?  

Where are you using `construct_knn_graph_from_index()` in a place that `features` is not also directly accessible?

Eg. here `features` still seems accessible:

https://github.com/cleanlab/cleanlab/blob/b063279788b636484709302a3b1e79c4bfe30597/cleanlab/internal/neighbor/knn_graph.py#L191-L193

I see you use this method in `issue_manager/outlier`:  https://github.com/cleanlab/cleanlab/blob/b063279788b636484709302a3b1e79c4bfe30597/cleanlab/datalab/internal/issue_manager/outlier.py#L230
where `features` is not accessible. But that doesn't mean we should invoke the dangerous `._fit_X`  everywhere, we should instead only invoke it where absolutely necessary (ie. in outliers only).

My suggestion: make `construct_knn_graph_from_index()` (or a parent method that calls it) able to take in optional arg `features`, and rely on `features` if available instead of `._fit_X`.  Make sure `._fix_X` is only relied on in outlier-code, and explicitly confirm it works across all recent sklearn versions that Cleanlab package supports.

elisno: Updated `construct_knn_graph_from_index()` to take in `correction_features` as an optional argument.
It's a bad idea overall to mix the variable names `knn` and `features` when the `features` are not used by `knn` to do the search.

elisno:
CI failure is unrelated to this PR.

Scikit-learn 1.5.0 was recently released and only affects one test case in
https://github.com/cleanlab/cleanlab/blob/ca3892994e7f40b237b467761f72acf3786e8666/tests/test_classification.py#L785

cleanlab/outlier.py
```diff
@@ -453,7 +455,18 @@ def _get_ood_features_scores(
...
         # and n_neighbors (k value) If our query set of features matches the training set used to fit knn, the nearest
         # neighbor of each point is the point itself, at a distance of zero.
-        distances, _ = knn.kneighbors(features)
+        distances, indices = knn.kneighbors(features)
+        if (
+            correct_knn
+        ):  # This should only happen if knn is None at the start of this function. Will NEVER happen for approximate KNN provided by user.
+            _features_for_correction = (
+                knn._fit_X if features is None else features
+            )  # Hacky way to get features (training or test). Storing np.unique results is a hassle. ONLY WORKS WITH NearestNeighbors
```
jwmueller, line 464:
```suggestion
            )  # Hacky way to get features (training or test). Storing np.unique results is a hassle. ONLY WORKS WITH sklearn NearestNeighbors object
```

cleanlab/internal/neighbor/knn_graph.py
jwmueller: thanks!  Please confirm the following statement:  

Now Outliers would be the only issue manager that would be skipped if sklearn updates to no longer allow `._fit_X` attribute (and in that case the exception would be caught and Datalab would just skip outlier detection).   All of the other issue managers would still work.

elisno: Thanks for the question!

**Short Answer**: 🎉 _**Yes**_ 🎉 , the outlier issue manager would be the only one affected by an update to sklearn that removes the ._fit_X attribute, causing it to be skipped if the exception is caught.

**Detailed Response**:

To provide a more complete picture, there is only one place in our codebase where we use the `._fit_X` attribute, which is in the `OutOfDistribution` class. Based on this, we can confirm the following:

- The outlier issue manager is indeed the only manager that calls `OutOfDistribution` for scoring features.
- This process necessitates the construction of a k-NN graph (or an equivalent data structure) and subsequent correction using the `._fit_X` attribute, which we have tried to avoid.

It's important to note that the outlier issue manager can avoid being skipped if just gets a k-NN graph. In this case, it would not call `OutOfDistribution`, thereby bypassing the need for the `._fit_X` attribute.

Also, our test suite will catch if sklearn removes the ._fit_X attribute. This can be verified in our test file [here](https://github.com/cleanlab/cleanlab/blob/c13f32a2a6c1d6f3166d760b60f805c94d9f54fe/tests/internal/neighbor/test_neighbor.py#L52).

---

In summary, while the statement is correct under the condition that the outlier issue manager calls `OutOfDistribution`, there is a potential workaround by directly providing a k-NN graph to the outlier issue manager.

I hope this clarifies the situation.

Issue #923, opened 2023-12-16
[Datalab new issue type] detect ID columns in features


jwmueller:
[New Datalab issue type](https://docs.cleanlab.ai/master/cleanlab/datalab/guide/custom_issue_manager.html) called something like `identifier_column` that checks whether `features` contains a column that is entirely sequential integers.

That is check whether there exists a column `i` in `features` such that `set(features[:,i]) = set(c, c+1, ..., c+n)` for some integer `c`, where `n` = num-rows of `features`. Note we don't consider the ordering of the column in case the dataset was pre-shuffled.

If the condition is met, then we say this dataset has the `identifier_column` issue (similar to the non-iid issue).

In this case the `info` attribute of Datalab should specify which column(s) seem like the identifier.

`datalab.issues` does not need to reflect this issue-type at all, since it is not something related to the rows of the dataset.

Motivation: ID columns are common in datasets, but generally should be left out of ML modeling or they risk harming the results.

gmottajr:
Hi Jonas ,
I came across your issue on GitHub and would like to clarify a few points to ensure I fully understand your requirements.

Based on your description, it seems you're looking for a method to process a dataset (structured as a list of lists, with each inner list representing a row) and identify columns that consist of sequentially increasing integers. These columns are often identifiers and might not be useful for certain types of data analyses or machine learning models. Is my understanding correct?

To address this, my pre-embrionary proposition would be considering implementing a method within a class that takes a dataset as input, removes any columns with sequentially increasing integers, and then returns the modified dataset with those columns removed. How does that sound to you? Would this approach meet your needs?

Please let me know if this aligns with your expectations or if there are any additional details or specifications you require.

Thank you!

jwmueller:
There's a lot more work involved in than that. The goal is to create a new IssueManager following the instructions from this guide here:

https://docs.cleanlab.ai/master/cleanlab/datalab/guide/custom_issue_manager.html

This new IssueManager should specifically look at each column of the `features` input (a 2D matrix with rows = data points, columns = individual features).

01PrathamS:
hey @jwmueller really sorry for opening direct PR #947 without discussing , i'm writing code for detect id column issue manager by following <https://docs.cleanlab.ai/stable/cleanlab/datalab/guide/custom_issue_manager.html> ,  i also tried it in my local by looking at other issue managers such as null, label, .. but couldn't figure out and getting same error can you please help !! 

```
import numpy as np
from typing import Tuple, Optional

[...]

PR #1120, opened 2024-05-09
added custom issue manager class for detecting identifier columns

MaxJoas:
## Summary

> 🎯 **Purpose**: This PR solves #923 by adding a issue manager class that find identifier columns
>
> 📜 **Example Usage**: 

```python
import numpy as np 
import pandas as pd 
from cleanlab import Datalab
from cleanlab.datalab.internal.issue_manager_factory import register
from cleanlab.datalab.internal.issue_manager.identifier_column import IdentifierColumnIssueManager
register(IdentifierColumnIssueManager)
data = {
    'A': [1, 2, 3, 4, 5],
    'B': [6, 7, 8, 30, 10],
    'C': [2, 12, 13, 14, 15],
    'D': [4, 17, 18, 19, 20],
    'E': [21, 22, 23, 24, 25]
}
df = pd.DataFrame(data)
lab = Datalab(data=df, label_name="D")

lab.find_issues(issue_types={"identifier_column":{"features": df.values}})
lab.info
```

## Impact

No other areas impacted. It is now possible to find a new issue_type.

**Screenshots**
See code snippetabove

## Testing

Ran existing tests, added test_identifier_column.py with last commit.
Tested mainly types sequential and non-sequential arrays (sometimes with and without duplicates and sequences that do not start with 0)
```
@pytest.mark.parametrize(
    "arr, expected_output",
    [
        (np.array([1, 2, 3, 4, 5]), True),
        (np.array([1, 1, 2, 2, 3, 3, 5]), False),
        (np.array([1, 1, 3, 4, 5, 8, 10]), False),
        (np.array([0, 0, 0, 0, 0, 0, 0]), False),
        (np.array([4, 5, 5, 6, 7, 8, 9, 10]), True),
        (np.array([1, 3, 4, 4, 5, 6, 7, -1]), False),
        (np.array([2, 1, 3, 5, 6, 4]), True),
        (np.array([-1, -3, -2, -4, 0]), True),
        (np.array([]), False),
        (np.array([0, 0, 0]), False),
    ],
```

**Unaddressed Cases**
Providing Non-integer and non-numeric columns. I think there will be just an general error fro the issue manager, like with other issues.

https://github.com/cleanlab/cleanlab/issues/923
https://github.com/cleanlab/cleanlab/pull/947

> 🔗 What Git or Slack items (Issues, threads, etc) that are specifically related to
> this work? Please link them here.
None, I know of

## References
None

## Reviewer Notes
None


cleanlab/datalab/internal/issue_manager/identifier_column.py
```diff
@@ -0,0 +1,111 @@
...
+            return False
+        min_val, max_val = arr.min(), arr.max()
+        unique_sorted = set(np.unique(np.sort(arr)).tolist())
+
+        expected_set = set(range(min_val, max_val + 1))
+        # check for special cases (all same )
+        if len(expected_set) == 1:
+            return False
+        else:
+            return expected_set == unique_sorted
```
tataganesh, line 48:
```suggestion
        unique_sorted = np.unique(arr) # Returns a sorted unique list
        min_val, max_val = unique_sorted[0], unique_sorted[-1]
        expected_range = np.arange(min_val, max_val + 1)
        if expected_range.size == 1 or unique_sorted.size != expected_range.size:
            return False
        return bool((expected_range == unique_sorted).all())
```
It might be better to do all the operations in numpy and avoid multiple passes through the array.

cleanlab/datalab/internal/issue_manager/identifier_column.py
```diff
@@ -0,0 +1,111 @@
...
+        Returns:
+            None
+        """
+        if features is None:
+            raise ValueError("features must be provided to check for identifier columns.")
+
+        features = self._prepare_features(features)
+        score = np.array([self._is_sequential(features[:, i]) for i in range(features.shape[1])])
+        indices = [i for i in range(features.shape[1])]
+        issue_indices = [i for i in indices if score[i]]
```
tataganesh, line 98:
```suggestion
        score = np.array([self._is_sequential(features[:, i]) for i in range(features.shape[1])])
        issue_indices = np.where(score)
```
Use numpy operations wherever possible.

cleanlab/datalab/internal/issue_manager/identifier_column.py
```diff
@@ -0,0 +1,111 @@
...
+        self.issues = pd.DataFrame(
+            {
+                f"is_{self.issue_name}_issue": score,
+                self.issue_score_key: score,
+            },
+            index=indices,
```
tataganesh, line 104: This can be removed since the default index should be the range of indices.

cleanlab/datalab/internal/issue_manager/identifier_column.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+
+        self.summary = self.make_summary(score=min(sum(score), 1))
+        self.info = {
+            "identifier_columns": issue_indices,
+            "num_identifier_columns": sum(score),
```
tataganesh, line 110: Replace `sum(score)` with `scores.sum()`.

tataganesh (changes requested):
This is a great start for implementing the issue manager!
I added some minor suggestions.

cleanlab/datalab/internal/issue_manager/identifier_column.py
```diff
@@ -0,0 +1,111 @@
...
+
+        features = self._prepare_features(features)
+        score = np.array([self._is_sequential(features[:, i]) for i in range(features.shape[1])])
+        indices = [i for i in range(features.shape[1])]
+        issue_indices = [i for i in indices if score[i]]
+        self.issues = pd.DataFrame(
+            {
+                f"is_{self.issue_name}_issue": score,
+                self.issue_score_key: score,
+            },
```
tataganesh, line 103: From the description of the issue manager [here](https://github.com/cleanlab/cleanlab/issues/923):

> datalab.issues does not need to reflect this issue-type at all, since it is not something related to the rows of the dataset.

So you do not need to update `self.issues`. Make sure your tests reflect the same.

tataganesh:
Since the input feature columns would almost always be a mixture of integer and float types, we want to ensure that the original type of each column is preserved (especially if features are passed as a list, dict or pandas dataframe).  Here are some pointers to implement this and ensure that we are calling  `_is_sequential` only for integer columns. 


* If `features` is an `np.ndarray`, all elements in the numpy array are likely to be of the same type. You can check if `features`  contains integers, else raise an exception. 

```python
        if isinstance(features, np.ndarray):
            if not np.issubdtype(features.dtype, np.integer):
                raise ValueError("...")  
            return features.T #  Transposed so that we can iterate over columns
```

* If `features` are `list | pd.DataFrame | dict`, convert each column to a numpy array (if needed) and return a list of column arrays. 

E.g. if `features` is a `pd.DataFrame`
```python
      feature_list = list()
      for feature_name, feature_value in features.items():
          feature_list.append(np.array(feature_value))
      return feature_list
```

In `find_issues`, you can then iterate over each column, check if it is of integer type and then call `_is_sequential` on it. The code might look like this: 

```python
processed_features = self._prepare_features(features)

# Check if feature is of integer type and then check for sequential integers. 
score = np.array([np.issubdtype(feature.dtype, np.integer) and self._is_sequential(feature) for feature in features])
```

Let me know if this helps in handling some cases of Non-integer and non-numeric columns!

cleanlab/datalab/internal/issue_manager/identifier_column.py
MaxJoas: Thank you for the constructive review. Just out of interest: Why do you prefer score.sum() ofr sum(score). Is it consistency, or is scores.sum() superior?

cleanlab/datalab/internal/issue_manager/identifier_column.py
MaxJoas: Good catch, makes sense. But when I run an example without updating self.issues, I get the following output:
```
Finding identifier_column issues ...
Error in identifier_column: 'is_identifier_column_issue'
Failed to check for these issue types: [IdentifierColumnIssueManager]

Audit complete. 0 issues found in the dataset.
Please specify some `issue_types` in datalab.find_issues() to see a report.
```
Is this an indented behaviour for this kind of issue?
If I run it with updating self.issues, I get:
```
Finding identifier_column issues ...

Audit complete. 2 issues found in the dataset.
Dataset Information: num_examples: 5, num_classes: 5

Here is a summary of various issues found in your data:

       issue_type  num_issues
identifier_column           2

Learn about each issue: https://docs.cleanlab.ai/stable/cleanlab/datalab/guide/issue_type_description.html
See which examples in your dataset exhibit each issue via: `datalab.get_issues(<ISSUE_NAME>)`

Data indices corresponding to top examples of each issue are shown below.


----------------- identifier_column issues -----------------

About this issue:
        Checks whether there is an identifier_column in the features of a dataset.
   Identifier columns are defined as a column i in features such that
   set(features[:,i]) = set(c, c+1, ..., c+n) for some integer c,
   where n = num-rows of features. If there is such a column, the dataset has
   the identifier_column issue
   

Number of examples with this issue: 2
Overall dataset quality in terms of this issue: 1.0000

   is_identifier_column_issue  identifier_column_score
1                       False                    False
2                       False                    False
3                       False                    False
0                        True                     True
4                        True                     True

Additional Information: 
identifier_columns: (array([0, 4]),)
```

I run the following example:
``` python
import numpy as np 
import pandas as pd 
from cleanlab.datalab.datalab import Datalab
from cleanlab.datalab.internal.issue_manager_factory import register
from cleanlab.datalab.internal.issue_manager.identifier_column import IdentifierColumnIssueManager
register(IdentifierColumnIssueManager)
data = {
    'A': [1, 2, 3, 4, 5],
    'B': [6, 7, 8, 30, 10],
    'C': [2, 12, 13, 14, 15],
    'D': [4, 17, 18, 19, 20],
    'E': [21, 22, 23, 24, 25]
}
df = pd.DataFrame(data)
lab = Datalab(data=df, label_name="D")

lab.find_issues(issue_types={"identifier_column":{"features": df.values}})
lab.info
lab.report()
```
On a side note: I could not import Datalab with `from cleanlab import Datalab`, but I needed to do: `from cleanlab.datalab.datalab import Datalab`.

I also found this problem on [stackoverflow](https://stackoverflow.com/questions/78086800/the-datalab-package-keeps-complaining-about-missing-dependencies). Do you know anything about this problem?

MaxJoas:
> Since the input feature columns would almost always be a mixture of integer and float types, we want to ensure that the original type of each column is preserved (especially if features are passed as a list, dict or pandas dataframe). Here are some pointers to implement this and ensure that we are calling `_is_sequential` only for integer columns.
> 
> * If `features` is an `np.ndarray`, all elements in the numpy array are likely to be of the same type. Return transpose of the features to easily iterate over the columns.
> 
> ```python
>         if isinstance(features, np.ndarray):
>             return features.T #  Transposed so that we can iterate over columns
> ```
> 
> * If `features` are `list | pd.DataFrame | dict`, convert each column to a numpy array (if needed) and return a list of column arrays.
> 
> E.g. if `features` is a `pd.DataFrame`
> 
> ```python
>       feature_list = list()
>       for feature_name, feature_value in features.items():
>           feature_list.append(np.array(feature_value))
>       return feature_list
> ```
> 
> In `find_issues`, you can then iterate over each column, check if it is of integer type and then call `_is_sequential` on it. The code might look like this:
> 
> ```python
> processed_features = self._prepare_features(features)
> 
> # Check if feature is of integer type and then check for sequential integers. 
> score = np.array([np.issubdtype(feature.dtype, np.integer) and self._is_sequential(feature) for feature in processed_features])
> ```
> 
> Let me know if this helps in handling some cases of Non-integer and non-numeric columns!

Thank you, will check this in the next days!

cleanlab/datalab/internal/issue_manager/identifier_column.py
tataganesh: Good point! `score.sum()` will call numpy's sum function, which should be more efficient on numpy arrays than `sum(score)`, where `sum` is a python function for iterables.

cleanlab/datalab/internal/issue_manager/identifier_column.py
tataganesh: Interesting. I will try and find a better way to handle this. In the meantime, a simple workaround would be to  assign `is_identifier_column_issue` as `False` for all rows and `identifier_column_score` as `1.0` for all rows.  Keep in mind that rows here indicate data points and not the feature columns.

cleanlab/datalab/internal/issue_manager/identifier_column.py
tataganesh: I think the variable name is still `score`.

cleanlab/datalab/internal/issue_manager/identifier_column.py
tataganesh: Here is my understanding of the issue columns:
A score of `1.0` is assigned to a sample (for a specific issue, e.g. label issue) if the sample does not exhibit that issue.  Lower scores indicate more severity. Hence, for this issue manager, since none of the rows are affected, the scores for each row (i.e. each sample) should be `1.0`.  And the `is_identifier_column_issue` should be `False` for all rows. The `info` attribute will be used to inspect which columns suffer from the identifier issue. Does that make sense?

MaxJoas: Yes, you're right! Sorry for the confusion. I had the typical boolean values (0=False and 1=True) in mind, but in the context of the issue manager, where low scores are severe issues, it should be the other way around. Thank you!

tataganesh:
@MaxJoas Feel free to request a review from me once sufficient changes (including tests) have been pushed, thanks!

tests/datalab/issue_manager/test_identifier_column.py
```diff
@@ -0,0 +1,120 @@
...
+    manager.find_issues(features)
+    if (manager.issues[f"is_{manager.issue_name}_issue"] == expected_is_identifer_column).all():
+        pass
+    else:
+        raise AssertionError()
+
+    if (manager.issues[manager.issue_score_key] == expected_score).all():
+        pass
+    else:
+        raise AssertionError()
```
tataganesh, line 120: Explicitly check for your condition using `assert condition` instead of using if-else and raising an exception.  
`expected_is_identifer_column` should be of `num_rows` size.

tests/datalab/issue_manager/test_identifier_column.py
```diff
@@ -0,0 +1,120 @@
...
+    prepared_features = manager._prepare_features(features)
+    assert np.array_equal(prepared_features, expected_prepared_features)
+
+    if isinstance(features, str):
+        with pytest.raises(ValueError):
+            manager._prepare_features(features)
+
+
+@pytest.mark.parametrize(
+    "features, expected_num_identifier_columns, expected_indices, expected_total_res",
```
tataganesh, line 54: A few comments on testing `find_issues`: 

* Consider removing `expected_num_identifier_columns` from the test cases and instead use `len(expected_indices)` while testing. 
* Replace expected_total_res` with `expected_score`. 
* Make `expected_indices` a list instead of a numpy array
* You can add a different function named `test_find_issues_mixed_types` with pandas dataframe and nested list inputs containing columns of different types. Just adding 3-4 different cases should suffice.

tests/datalab/issue_manager/test_identifier_column.py
```diff
@@ -0,0 +1,120 @@
...
+            np.array([0]),
+            0.0,
+        ),
+        (
+            np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]]),
+            0,
+            np.array([]),
+            1.0,
+        ),
+        (np.array([[1, 2, 7], [4, 3, 8], [7, 4, 9], [10, 5, 10]]), 2, np.array([1, 2]), 0.0),
```
tataganesh, line 82: These two cases are not very different from the ones above them. Consider removing.

tests/datalab/issue_manager/test_identifier_column.py
```diff
@@ -0,0 +1,120 @@
...
+            np.array([[1, 2, 3], [1, 2, 3], [1, 2, 3]]),
+            0,
+            np.array([]),
+            1.0,
+        ),
+        (
+            np.array([[0, 2, 3], [-1, 5, 6], [-2, 2, 3]]),
+            1,
+            np.array([0]),
+            0.0,
```
tataganesh, line 74: You can change the number of rows/columns of some of these to add more variation.

tataganesh (changes requested):
Thanks for addressing most of the feedback! Here is a brief summary of my latest comments:

**Tests**
* Restructure `find_issues` tests to handle different input scenarios (pandas dataframe, mixed types etc.)
* If possible, try to add variation to your test cases but do not include too many cases.
*  Refer to other issue manager tests and try to restructure your tests in a similar fashion for consistency.

Implementation
* Ensure that dataframe column types are preserved
* Better variable naming and using numpy functions

cleanlab/datalab/internal/issue_manager/identifier_column.py
```diff
@@ -0,0 +1,110 @@
...
+            {
+                f"is_{self.issue_name}_issue": False,
+                self.issue_score_key: [1.0 for _ in range(processed_features.shape[1])],
+            },
+        )
+        # score in summary should be 1.0 if the issue is not presend and 0.0 if at least one column is an identifier column
+        self.summary = self.make_summary(score=1.0 if scores.sum() == 0 else 0.0)
+        self.info = {
+            "identifier_columns": issue_indices,
+            "num_identifier_columns": scores.sum(),
```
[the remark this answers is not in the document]
MaxJoas: yes, good catch, changed it

cleanlab/datalab/internal/issue_manager/identifier_column.py
```diff
@@ -0,0 +1,110 @@
...
+            raise ValueError("features must be provided to check for identifier columns.")
+        processed_features = self._prepare_features(features)
+
+        scores = np.array(
+            [
+                np.issubdtype(feature.dtype, np.integer) and self._is_sequential(feature)
+                for feature in processed_features
+            ]
+        )
+        issue_indices = np.where(scores)
```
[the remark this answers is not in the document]
MaxJoas: done

cleanlab/datalab/internal/issue_manager/identifier_column.py
```diff
@@ -0,0 +1,110 @@
...
+            features (Optional[npt.NDArray | pd.DataFrame | list | dict]):
+            The dataset to check for identifier columns.
+        Returns:
+            None
+        """
+        if features is None:
+            raise ValueError("features must be provided to check for identifier columns.")
+        processed_features = self._prepare_features(features)
+
+        scores = np.array(
```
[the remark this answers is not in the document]
MaxJoas: I think the other issue_manager implementation, name this scores, too.
I would suggest to keep scores and cast it to 1.0,, 0.0 instead of a boolean value?

cleanlab/datalab/internal/issue_manager/identifier_column.py
```diff
@@ -0,0 +1,110 @@
...
+            features: features as npt.NDArray
+        """
+        if isinstance(features, np.ndarray):
+            return features.T
+        elif isinstance(features, pd.DataFrame) or isinstance(features, dict):
+            return np.array([col for _, col in features.items()])
+        elif isinstance(features, list):
+            for col_list in features:
+                if not isinstance(col_list, list):
+                    raise ValueError("features must be a list of lists if it features is a list.")
```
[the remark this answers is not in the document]
MaxJoas: Maybe we check instead if it is either a list of lists or list of numpy arrays?
```python
        elif isinstance(features, list):
            # check if all elements of the list are either lists or numpy arrays
            first_type = type(features[0])
            if first_type not in [list, np.ndarray]:
                raise TypeError("Elements of 'features' must be either all lists or all numpy arrays for type list.")
            for col_list in features:
                if not isinstance(col_list, first_type):
                    raise TypeError("If features is a list, it must be a list of lists, or a list of numpy arrays.")
  ```
  Will add a test for this

File: tests/datalab/issue_manager/test_identifier_column.py [hunk 14]
```diff
@@ -0,0 +1,120 @@
...
+def test_prepare_features(features, expected_prepared_features):
+    lab = Datalab(pd.DataFrame(features))
+    manager = IdentifierColumnIssueManager(datalab=lab)
+
+    prepared_features = manager._prepare_features(features)
+    assert np.array_equal(prepared_features, expected_prepared_features)
+
+    if isinstance(features, str):
+        with pytest.raises(ValueError):
+            manager._prepare_features(features)
```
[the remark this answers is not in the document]
MaxJoas: It tests that the exception gets raised, when the users passes a unsupported input type, I'll pack it into another function and test it with pytest.raises

tests/datalab/issue_manager/test_identifier_column.py
MaxJoas: done, changed them with a more different case

tests/datalab/issue_manager/test_identifier_column.py
MaxJoas: done, added two different ones

MaxJoas:
@tataganesh I've implemented the changes from the last review. Would you review again?

cleanlab/datalab/internal/issue_manager/identifier_column.py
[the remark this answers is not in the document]
tataganesh: In other issue managers, `scores` are used for datapoints/rows, but in this case they are used for columns. Hence, I suggested to use a different variable name.

cleanlab/datalab/internal/issue_manager/identifier_column.py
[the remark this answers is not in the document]
tataganesh: Sorry, what was the change made here?

tests/datalab/issue_manager/test_identifier_column.py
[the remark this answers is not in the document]
tataganesh: Was this test added or removed?

cleanlab/datalab/internal/issue_manager/identifier_column.py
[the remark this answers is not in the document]
tataganesh: Thanks. Is there any reason for still using the `sum` function instead of the `size` function? Also, as suggested before, please rename `issue_indices` to `identifier_column_indices` for more clarity ([reference comment](https://github.com/cleanlab/cleanlab/pull/1120#discussion_r1748878263)).

tests/datalab/issue_manager/test_identifier_column.py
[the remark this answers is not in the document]
MaxJoas: this test was removed, I think we don't need it necessarily

elisno (approved):
LGTM. Great work @MaxJoas! And thank you @tataganesh for the review!

cleanlab/datalab/internal/issue_manager/identifier_column.py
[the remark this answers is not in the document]
MaxJoas: done

tests/datalab/issue_manager/test_identifier_column.py
MaxJoas: done

tests/datalab/issue_manager/test_identifier_column.py
MaxJoas: done

cleanlab/datalab/internal/issue_manager/identifier_column.py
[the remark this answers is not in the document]
MaxJoas: done

PR #1128, opened 2024-05-24
Tubular Tutorial: Dynamically handle displaying exact duplicate sets

nelsonauner:
## Summary

> 🎯 **Purpose**: Describe the objective of your changes in this Pull-Request.

This simple change improves the tutorial by making sure the referenced row (Student 690) actually is displayed in the tutorial! Otherwise can lead to a very confusing result, for example, when I ran the tutorial:

image

A quick change to sort fixes this. Using `mergesort` is important because it's the only stable sort method to keep the duplicate-score-sorting working - see https://stackoverflow.com/questions/33699555/pandas-sorting-by-value-and-then-by-index

Here's the results. Student 690 isn't the first row, so I also updated the copy

image

### DIFF
--- a/docs/source/tutorials/datalab/tabular.ipynb
+++ b/docs/source/tutorials/datalab/tabular.ipynb
@@ -427,15 +427,6 @@
     "indices_to_display = [lowest_scoring_duplicate] + duplicate_results.loc[lowest_scoring_duplicate, \"near_duplicate_sets\"].tolist()\n",
     "\n",
     "# Display the relevant rows from the original dataset\n",
-    "X_raw.iloc[indices_to_display]\n"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
     "X_raw.iloc[indices_to_display]"
    ]
   },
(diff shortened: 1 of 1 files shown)


elisno:
Hi @nelsonauner !

Thank you for pointing out this issue! Good catch on the unstable sorting algorithm.
It turns out there are just too many sets of exact duplicates in this tutorial, leading to varying results.

I've made some changes to ensure a more robust and clear approach.

- In the tutorial, we display the lowest-scoring example (based on the near-duplicate score) and its associated near-duplicate sets. I don't see a reason for us to hard-code which examples to display.

- I also updated the relevant text to be agnostic to specific hardcoded IDs (690, 246, etc.).

Finally, I added tests at the end of the notebook to ensure that all displayed rows are identical (since we state they are exact duplicates).

---

The ideal situation would be if we were able to add more interactivity to the docs, like an interactive widget:


https://github.com/cleanlab/cleanlab/assets/18127060/c1f6a1f8-c247-49cf-bab8-68634e224830

Unfortunately, this isn't so straightforward when we build the docs with these notebooks.

docs/source/tutorials/datalab/tabular.ipynb
```diff
@@ -522,7 +558,75 @@
...
+      "model_module_version": "1.5.0",
+      "model_name": "SliderStyleModel",
+      "state": {
+       "description_width": ""
+      }
+     }
+    },
+    "version_major": 2,
+    "version_minor": 0
+   }
```
elisno, line 629:
```suggestion
   "version": "3.10.12"
```

elisno (approved):

PR #1142, opened 2024-06-12
Update knn shapely score computation

### LINKED ISSUE #1127: update knn shapely score transformation
to be: max(x, 0) , instead of the current transformation: 0.5*(x+1)

Needs property based testing. Test should fail if x is ever negative pre-transformation.

Also update the is_issue threshold to be <= 1e-6

elisno:
## Summary

The original knn shapely score implementation is unlikely to find examples with net-negative contributions to model training.
This PR revisits the implementation of the raw score computation but places the `x -> 0.5(x + 1)` transformation
such that the final score lies in the range [0, 1] in `data_shapley_knn()` (the public function).

Here's the reference from the relevant paper: https://arxiv.org/pdf/1908.08619
image

## Testing

- This PR adds a new test class for a property-based test that asserts that the raw scores lie in the appropriate range.
- It adds several test cases for clear label/feature errors being added are captured in this score.
   - Test the scores and issue flags when there's a label error.
   - Same for one outlier in feature space.
   - How the scores/issue flags behave when there's one exact duplicate (with and without conflicting labels).
   - How sorting the dataset by the data valuation score and removing some percentages of the dataset beats random removals, in terms of model accuracy.
     - One parametrized test for a binary dataset.
     - Another parametrized test for a multi-class dataset.

Running the data valuation check through Datalab gives this strip plot of the different scores across the classes (it's recommended to view the enlarged image on a wider screen, due to the number of classes not fitting well on the strip plot).

image

This image was generated with the following code snippet added to the text tutorial (replace the cell that calls `Datalab.find_issues()` with the following code snippet ([see section 4](https://github.com/cleanlab/cleanlab/blob/master/docs/source/tutorials/datalab/text.ipynb)):

```pyhton
lab = Datalab(data_dict, label_name="labels")
lab.find_issues(features=text_embeddings, issue_types={"data_valuation": {"k": 10}})

# Collect the identified issues
data_valuation_issues = lab.get_issues("data_valuation")

# Display the data valuation issues
display(data_valuation_issues)

import seaborn as sns
import matplotlib.pyplot as plt

# Prepare the data for plotting a histogram
plot_data = data_valuation_issues.join(pd.DataFrame({"given_label": data_dict["labels"]}))

# Normalize the data for proportional count plot
plot_data['count'] = plot_data.groupby('given_label')['data_valuation_score'].transform('count')
plot_data['proportion'] = plot_data.groupby('given_label')['data_valuation_score'].transform(lambda x: x / x.count())


plt.figure(figsize=(21, 6))
sns.stripplot(
    data=plot_data,
    x="given_label",
    y="data_valuation_score",
    jitter=3.5,
    hue="given_label",
    dodge=True,
    alpha=0.5,
)
plt.xticks(rotation=45)
plt.xlim(-1.0, num_classes)
plt.show()
```

Similarily, the visualization in the Misc Datalab workflow tutorial has been turned into a strip plot.

image

- Resolves #1127


elisno:
An unrelated test case fails due to a recent [fastText release](https://pypi.org/project/fasttext/), which has yet to be announced. No information can be found in the repository, as it was archived in Mar 19, 2024.

jwmueller:
screenshot of the new distribution of scores on a real dataset? cant really review this without that information

tests/test_data_valuation.py
```diff
@@ -52,3 +56,55 @@ def test_data_shapley_knn_with_knn_graph(self, labels, knn_graph):
...
+            elements=st.floats(min_value=-1.0, max_value=1.0),
+        )
+    )
+
+    # Generate k such that it is less than the length of labels and features
+    k = draw(st.integers(min_value=1, max_value=length - 1))
+
+    return labels, features, k
+
+
```
jwmueller, line 92: Is there 0 test that introduces errors in a dataset (in features/labels) and verifies the shapley values for those erroneous datapoints is actually lower than for other datapoints?

That seems like a must have test

docs/source/tutorials/datalab/workflows.ipynb
```diff
@@ -345,21 +345,20 @@
     ")\n",
     "\n",
     "# Plot histograms of data valuation scores for each label\n",
-    "sns.histplot(\n",
+    "sns.stripplot(\n",
```
elisno, line 348: Here's how the new plot looks:

image


---

For comparison here's the histplot that's currently on master:
image
The log-scale of the y-axis skews the results, and the scores seem to always be clipped at 0.5.

pyproject.toml
```diff
@@ -9,7 +9,7 @@ name = "cleanlab"
 # requirements files see:
 # https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/
 dependencies = [
-  "numpy>=1.22.0",
+  "numpy~=1.22",
```
elisno, line 12: Numpy was recently updated to 2.0.0, which may fail some unrelated tests.

elisno: We can revert this if we want before merging.

tests/test_data_valuation.py
elisno: Most of these tests are now added in `test_datalab.py`

elisno:
The current failing link-check is unrelated to this PR.

pyproject.toml
jwmueller: Isn't the near-term plan going forward to support 

numpy >= 1.22.0 < 2.0.0

?

Not sure I understand this syntax you're using here, doesn't seem equivalent to me.

docs/source/tutorials/datalab/workflows.ipynb
jwmueller: why is this plot broken down by class? I'm not sure what the reader is supposed to take away from that. I think that might confuse users.

Can you also explain at the end of this Data Valuation section why the scores are all above 0.5?  That also seems confusing to somebody who knows Shapely values.

cleanlab/data_valuation.py
```diff
@@ -113,4 +136,8 @@ def data_shapley_knn(
     # Use provided knn_graph or compute it from features
     if knn_graph is None:
         knn_graph, _ = create_knn_graph_and_index(features, n_neighbors=k, metric=metric)
-    return _knn_shapley_score(knn_graph, labels, k)
+
+    num_examples = labels.shape[0]
+    distances = knn_graph.indices.reshape(num_examples, -1)
+    scores = _knn_shapley_score(neighbor_indices=distances, y=labels, k=k)
+    return 0.5 * (scores + 1)
```
jwmueller, line 143: youre confident this is better than doing:
`max(scores, 0)` ?

jwmueller: what's the smallest raw (pre-transformation) score you've encountered?

pyproject.toml
elisno: I'm also thinking the near-term plan should include supporting version 1.X, specifically 1.22 and above.

---

`~1.X` is equivalent to `>= 1.X, == 1.*`.
https://peps.python.org/pep-0440/#compatible-release

I believe this is a reasonable version specifier as it captures the desired compatibility range.

cleanlab/data_valuation.py
jwmueller: or what about: 

```
(x+1) * 0.001 if x <= 0
0.001 + x*(1 - 0.001) if x > 0
```

docs/source/tutorials/datalab/workflows.ipynb
jwmueller: instead is there anything interesting you can point out about the highest/lowest valuation data points?

elisno: To be clear, the latter image is from before this PR.

For smaller random data points, I've observed values closer to the extreme ranges of [-1, 1].

I also believe that for larger datasets, the raw scores tend to shrink around zero, essentially scaling themselves down. However, I would discourage attempting to rescale these values, as we have limited control over them.

The `0.5(x +1)` transformation is good enough to stick to our canonical [0, 1] range.

cleanlab/data_valuation.py
elisno: For smaller random data points, I've observed values closer to the extreme ranges of [-1, 1].

I also believe that for larger datasets, the raw scores tend to shrink around zero, essentially scaling themselves down. However, I would discourage attempting to rescale these values, as we have limited control over them.

The `0.5(x +1)` transformation is good enough to stick to our canonical [0, 1] range.

elisno: The initial suggestion to use `max(score, 0)` was considered well before the latest revision of the raw score implementation.

Using `0.5 * (scores + 1)` is more effective than max(scores, 0) because it accounts for both positive and negative scores, transforming them consistently into the [0, 1] range.

`0.5 * (scores +1)` is better than `max(scores, 0)` as the positive scores may also be exceedingly small.
Then we risk againt that someone would interpret it as all points have small scores which is an issue.

The `max(score, 0)` approach might lead to misinterpretations, especially when positive scores are very small, potentially conveying that all points have low scores, which is not the case.

We'll have a more accurate and balanced representation of the scores within our desired range if we stick to `0.5 * (scores + 1)`.

jwmueller: sg thanks for the info!

pyproject.toml
jwmueller: how is `>= 1.X, == 1.*.` equivalent to `>= 1.22.0 < 2.0.0` ?

Doesn't your version include 1.18, 1.19, etc? How do we know those work with cleanlab?

docs/source/tutorials/datalab/workflows.ipynb
jwmueller: i'm ok with leaving the image as is, I'm just calling out that the coloring immediately leads me to think I'm supposed to take something away based on the per-class results, but I'm not seeing any clear takeaway based on the per-class results.

If there is supposed to be some takeaway (or even not), it would be good to add one sentence saying it, so others aren't confused like I was.

pyproject.toml
jwmueller: let's just make this `>= 1.22.0 < 2.0.0` unless you have a specific reason not to

elisno: Sorry for any confusion. I thought explaining the pattern with X would make it clearer.

The current change in the PR specifies ~=1.22, which is equivalent to >= 1.22.0 < 2.0.0. This ensures we don't accidentally make a change like >= 1.22.0 < 2.0.1, which would include 2.0.0 in the range. Semantically, this is safer than setting both ranges when we intend to only support versions compatible with 1.22.

---

To clarify, my version does **not** include 1.18, 1.19, etc. It includes 1.22, 1.23, 1.24, 1.25, and 1.26 as of this writing.

docs/source/tutorials/datalab/workflows.ipynb
elisno: Kept the image, and [updated the text](https://github.com/cleanlab/cleanlab/pull/1142/files#diff-1e2bf51a977e3f9deed91acef9f3b6167134a1d1f64e2c6d0e4550a608635e62R328-R332) explaining the 0.5 cutoff and the  takeaway for breaking the plot down by class.

docs/source/tutorials/datalab/workflows.ipynb
```diff
@@ -325,7 +325,11 @@
    "metadata": {},
    "source": [
     "### 4. (Optional) Visualize Data Valuation Scores\n",
-    "Finally, we will visualize the data valuation scores using a histogram to understand the distribution of scores across different labels."
+    "Finally, we will visualize the data valuation scores using a strip plot to understand the distribution across different labels.\n",
+    "\n",
+    "A score below 0.5 indicates a negative contribution to the model's training performance, while a score above 0.5 indicates a positive contribution.\n",
+    "\n",
+    "By examining the scores across different labels, we can identify whether positive or negative contributions are disproportionately concentrated in a single class. This can help us detect potential biases in the training data."
```
jwmueller, line 332:
```suggestion
    "Let's visualize the data valuation scores across our dataset.\n",
    "\n",
    "Cleanlab's Shapely scores are transformed to lie between 0 and 1 such that: a score below 0.5 indicates a negative contribution to the model's training performance, while a score above 0.5 indicates a positive contribution.\n",
    "\n",
    "By examining the scores across different classes, we can identify whether positive or negative contributions are disproportionately concentrated in a single class. This can help detect biases in the training data."
```

jwmueller (approved):
LGTM thanks for clarifications, sorry for confusion on my part! 
Just had one suggested wording change

PR #1144, opened 2024-06-13
Pin fasttext in CI

elisno:


elisno:
Failing tests in CI are unrelated to this change.

https://github.com/huggingface/datasets made a minor release an hour ago. Still need to investigate, but it's limited to `Datalab` for now.

elisno:
This PR only affects one test that runs in CI on linux/macOS.

Previously, the highlighted test always failed on Linux (which run through the test suite faster than other platforms, hence they fail first).

The following screenshot shows the affected test passing:

image

PR #1194, opened 2024-08-23
updating default value of threshold for spurious correlations

allincowell:
Updating default value of threshold from 0.05 to 0.3. The threshold value for one of the test cases is hard-coded back to 0.01 to avoid dependency on the default threshold value should it need to be updated in the future.


elisno:
Thanks for the update, but I recommend keeping the test as-is with the hardcoded threshold of 0.01. This ensures that the test consistently checks behavior against a specific value, regardless of future changes to the default threshold.

elisno (approved):

PR #946, opened 2024-01-06
Add unittest to test_datalab.py 

01PrathamS:
## Summary

> 🎯 **Purpose**: Add unit test to test_datalab.py file
                              1. ```datalab.issues``` solely contains numeric and boolean values.
                              2. ```datalab.issue_summary``` solely contains numeric value

## Links to Relevant Issues or Conversations

Resolves #937

Issue #937, opened 2023-12-28
add a unit test that `datalab.issues` solely contains numeric and boolean values


jwmueller:
Can go in this file:

https://github.com/cleanlab/cleanlab/blob/master/tests/datalab/test_datalab.py

Also add a second unit test that `datalab.issue_summary` solely contains numeric values.

sanjanag:
https://github.com/cleanlab/cleanlab/blob/master/tests/datalab/test_cleanvision_integration.py should also contain a test testing there are solely numeric values in datalab.issues. Currently, it is broken because `score` column does not extend to cleanvision right now.
<img width="959" alt="Screenshot 2024-02-06 at 3 38 40 PM" src="https://github.com/cleanlab/cleanlab/assets/7639432/bb9a85e2-44a3-47cb-8a45-2922d0d7f59e">

PR #1218, opened 2024-12-25
test_datalab.py code improvements

gordon-lim:
## Summary

> 🎯 **Purpose**: Add data type checks for `datalab.find_issues`.

✏️ Added unit tests for: 
* datalab.issues
  * `is_{issue_type}` columns only contain boolean values 
  * `{issue_type}_score` columns only contain numerical values
  * `["given_label", "predicted_label", "distance_to_nearest_neighbor"]` columns only contain numerical values
* datalab.issue_summary: solely contains numeric value
  * `["score", "num_issues"]` columns only contain numerical values 

✏️ Modified `TestDatalabFindNonIIDIssues::test_incremental_search`:

Issue: `TestDatalabFindNonIIDIssues` uses `sorted_embeddings` with 1000 samples but only 3 labels. This causes `test_incremental_search` to fail aforementioned tests because `is_class_imbalance_issue` and `class_imbalance_score` would have NaNs due to incomplete labeling. 

Fix: Test incremental search with only `issue_types={"null": {}}` followed by `issue_types={"non_iid": {}}` as opposed to all default issue types followed by `non_iid`. Change expected `len(summary)` accordingly. As such, we can ignore the class_imbalance columns. This aligns with other tests for incremental search in the rest of the file -- call `find_issues()` with one issue_type followed by the one of interest.

**Unaddressed Cases**

My code adds functions to perform type checks for `datalab.issues` and `datalab.issue_summary`. I call those functions for every 
 test of `datalab.find_issues()`. As such the coverage is limited to the existing scope of inputs.

> Resolves [#937](https://github.com/cleanlab/cleanlab/issues/937)🔗 

Previous effort by @01PrathamS [[#946]](https://github.com/cleanlab/cleanlab/pull/946) used mocked data and did not test for different combinations of inputs to `find_issues()`.

### DIFF
--- a/tests/datalab/datalab/test_datalab.py
+++ b/tests/datalab/datalab/test_datalab.py
@@ -147,6 +147,7 @@ class TestDatalab:
             }
         )
         monkeypatch.setattr(lab, "issue_summary", mock_summary)
+        assert lab.issue_summary is lab.data_issues.issue_summary
 
         label_summary = lab.get_issue_summary(issue_name="label")
         pd.testing.assert_frame_equal(label_summary, mock_summary.iloc[[0]])
@@ -171,6 +172,7 @@ class TestDatalab:
             },
         )
         monkeypatch.setattr(lab, "issues", mock_issues)
+        assert lab.issues is lab.data_issues.issues
 
         mock_predicted_labels = np.array([0, 1, 2, 1, 2])
 
--- a/tests/datalab/datalab/test_datalab.py
+++ b/tests/datalab/datalab/test_datalab.py
@@ -50,30 +50,36 @@ def test_datalab_invalid_datasetdict(dataset, label_name):
         Datalab(datadict, label_name)  # type: ignore
         assert "Please pass a single dataset, not a DatasetDict." in str(e)
 
+
 def check_issues_dtypes(lab):
     """
     Asserts that each column in lab.issues has the expected dtype.
-    
+
     """
 
     for col in lab.issues.filter(like="is_").columns:
-        assert lab.issues[col].dtype.kind == 'b', f"Column '{col}' must be boolean."
+        assert lab.issues[col].dtype.kind == "b", f"Column '{col}' must be boolean."
 
     for col in lab.issues.filter(like="_score").columns:
-        assert lab.issues[col].dtype.kind in {'i', 'f'}, f"Column '{col}' must be numeric."
+        assert lab.issues[col].dtype.kind in {"i", "f"}, f"Column '{col}' must be numeric."
+
 
 def check_issue_summary_dtypes(lab):
     """
     Asserts that each column in lab.issue_summary has the expected dtype.
-    
+
     """
-    
+
     for col in ["score", "num_issues"]:
         series = lab.issue_summary[col]
         dtype = series.dtype
-        assert np.issubdtype(dtype, np.number), (
-            f"Column '{col}' must be numeric"
-        )
+        assert np.issubdtype(dtype, np.number), f"Column '{col}' must be numeric"
+
+
+def check_lab_dtypes(lab):
+    check_issues_dtypes(lab)
+    check_issue_summary_dtypes(lab)
+
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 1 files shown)


tests/datalab/datalab/test_datalab.py
```diff
@@ -50,6 +50,38 @@ def test_datalab_invalid_datasetdict(dataset, label_name):
...
+    for col in lab.issues.filter(like="_score").columns:
+        assert lab.issues[col].dtype.kind in {'i', 'f'}, f"Column '{col}' must be numeric."
+
+    for col in ["given_label", "predicted_label", "distance_to_nearest_neighbor"]:
+        if col in lab.issues.columns:
+            series = lab.issue_summary[col]
+            dtype = series.dtype
+            assert np.issubdtype(dtype, np.number), (
+                f"Column '{col}' must be numeric"
+            )
```
jwmueller, line 71: I don't believe these columns ("given_label", "predicted_label", "distance_to_nearest_neighbor") are supposed to be part of `lab.issue_summary`.  Could you share an example where you found these columns to present in the issue_summary dataframe? 

If you didn't ever observe this, let's get rid of this part of the code.  It's also strange to check `issue_summary` in a method called:  `check_issues_dtypes()`

tests/datalab/datalab/test_datalab.py
```diff
@@ -147,6 +179,7 @@ def test_get_issue_summary(self, lab, monkeypatch):
             }
         )
         monkeypatch.setattr(lab, "issue_summary", mock_summary)
+        assert lab.issue_summary is lab.data_issues.issue_summary
```
jwmueller, line 182: I'd either get rid of this line, or add explanatory print statement whenever this assert fails.  Otherwise it's unclear what this is testing / what developers should do when this assert fails.

tests/datalab/datalab/test_datalab.py
```diff
@@ -171,6 +204,7 @@ def test_get_issues(self, lab, monkeypatch):
             },
         )
         monkeypatch.setattr(lab, "issues", mock_issues)
+        assert lab.issues is lab.data_issues.issues
```
jwmueller, line 207: I'd either get rid of this line, or add explanatory print statement whenever this assert fails. Otherwise it's unclear what this is testing / what developers should do when this assert fails.

tests/datalab/datalab/test_datalab.py
```diff
@@ -50,6 +50,38 @@ def test_datalab_invalid_datasetdict(dataset, label_name):
...
+    Asserts that each column in lab.issue_summary has the expected dtype.
+    
+    """
+    
+    for col in ["score", "num_issues"]:
+        series = lab.issue_summary[col]
+        dtype = series.dtype
+        assert np.issubdtype(dtype, np.number), (
+            f"Column '{col}' must be numeric"
+        )
```
jwmueller, line 84: seems like in the code, you're always calling both 

```
check_issues_dtypes(lab)
check_issue_summary_dtypes(lab)
```
so I'd make another helper method that has both of these in it, and call this new helper throughout to simplify the code.

jwmueller (changes requested):
Thanks for your contribution!  Left some suggestions.

Also note that the black formatter is currently failing on your PR:
https://github.com/cleanlab/cleanlab/actions/runs/12494782645/job/34864813149?pr=1218

You'll need to run `black` formatter on the changed files before you push your final commit to the PR.  You can do this via pre-commit hooks:
https://github.com/cleanlab/cleanlab/blob/master/DEVELOPMENT.md#how-to-style-new-code-contributions

tests/datalab/datalab/test_datalab.py
gordon-lim: My mistake. Those columns appear only with datalab.get_issues() for certain issue_name arguments.

gordon-lim:
Thank you @jwmueller for the prompt and detailed review. I have addressed all of your feedback.

jwmueller (approved):
Thank you for the quick fixes

Issue #1175, opened 2024-07-12
replace Datalab load/save from pickle to more secure format


jwmueller:
such as writing to JSON file instead.

PR #1222, opened 2024-12-31
replace Datalab load/save from pickle to json

gordon-lim:
## Summary

> 🎯 **Purpose**: Replace Datalab load/save from pickle to more secure format (json)

✏️ Main contributions:

1. Update unit tests `test_save()` and `test_load`
  a. `.pkl` ---> `.json`
  b. As per my earlier note, I removed the check for `_data` if dataset is not provided during `Datalab.load()`
  c. I added more checks for other metadata such as `task`, `cleanlab_version`, `verbosity`.
3. Save and load metadata of `Datalab` using json

```python
# Main contribution: replace Datalab load/save from pickle to json

# ---- serialize.py ----
def serialize(cls, path: str, datalab: Datalab, force: bool) -> None:

  # ---- other code ----

  with open(os.path.join(path, OBJECT_FILENAME), "w") as f:
      json.dump(
          {
              "task": str(datalab.task),
              "label_name": datalab.label_name,
              "cleanlab_version": datalab.cleanlab_version,
              "verbosity": datalab.verbosity,
              "info": datalab.info,
              "_labels": datalab._labels,
              "_data_hash": datalab._data_hash,
          },
          f,
          default=custom_serializer,
      )

def deserialize(cls, path: str, data: Optional[Dataset] = None) -> Datalab:
  
  # ---- other code ----

  with open(os.path.join(path, OBJECT_FILENAME), "rb") as f:
    datalab_metadata = json.load(f)
    task = datalab_metadata["task"]
    verbosity = datalab_metadata["verbosity"]
    from cleanlab.datalab.datalab import Datalab

    datalab: Datalab = Datalab(data=[], task=task, verbosity=verbosity)
    datalab.label_name = datalab_metadata["label_name"]
    datalab.cleanlab_version = datalab_metadata["cleanlab_version"]
    datalab.info = datalab_metadata["info"]
    datalab._data_hash = datalab_metadata["_data_hash"]
    datalab._labels.labels = datalab_metadata["_labels"]["labels"]
    datalab._labels.label_map = datalab_metadata["_labels"]["label_map"]
  
```

## Impact

> 🌐 Areas Affected: Documentation for `Datalab.save()` and `Datalab.load` is affected. If my understanding is correct, a further PR is needed to update documentation to explain what is saved/loaded.
>
> 👥 Who’s Affected: Users of `Datalab`, especially those with saved datalab pickle files. 

**Unaddressed Cases**

> ⚠️ I build on existing test cases. Since dataset is not saved, I did not test for objects that are built using the dataset during `Datalab.init()` such as `self._imagelab` and `self._labels`. 

> Resolves [#1175](https://github.com/cleanlab/cleanlab/issues/1175)🔗

### DIFF
--- a/cleanlab/datalab/internal/serialize.py
+++ b/cleanlab/datalab/internal/serialize.py
@@ -27,7 +27,7 @@ from sklearn.neighbors import NearestNeighbors
 from cleanlab.outlier import OutOfDistribution
 
 import cleanlab
-from cleanlab.datalab.internal.data import Data, Label, MultiLabel, MultiClass
+from cleanlab.datalab.internal.data import Data, Label
 
 from datasets.arrow_dataset import Dataset
 
@@ -267,7 +267,9 @@ class _Serializer:
 
             # Load dataset from disk by default (if saved)
             if data is None:
-                loaded_data = Dataset.load_from_disk(os.path.join(path, DATA_DIRNAME))
+                dataset_path = os.path.join(path, DATA_DIRNAME)
+                if os.path.exists(dataset_path):
+                    loaded_data = Dataset.load_from_disk(dataset_path)
 
             datalab: Datalab = Datalab(
                 data=loaded_data if loaded_data else json_data["data"],
--- a/cleanlab/datalab/internal/serialize.py
+++ b/cleanlab/datalab/internal/serialize.py
@@ -27,7 +27,7 @@ from sklearn.neighbors import NearestNeighbors
 from cleanlab.outlier import OutOfDistribution
 
 import cleanlab
-from cleanlab.datalab.internal.data import Data, Label
+from cleanlab.datalab.internal.data import Data
 
 from datasets.arrow_dataset import Dataset
 
--- a/cleanlab/datalab/internal/serialize.py
+++ b/cleanlab/datalab/internal/serialize.py
@@ -35,7 +35,6 @@ if TYPE_CHECKING:  # pragma: no cover
 OBJECT_FILENAME = "datalab.pkl"
 ISSUES_FILENAME = "issues.csv"
 ISSUE_SUMMARY_FILENAME = "summary.csv"
-INFO_FILENAME = "info.pkl"
 DATA_DIRNAME = "data"
 
 
--- a/cleanlab/datalab/internal/serialize.py
+++ b/cleanlab/datalab/internal/serialize.py
@@ -17,7 +17,6 @@ from __future__ import annotations
 
 import os
 import json
-import pickle
 import warnings
 from typing import TYPE_CHECKING, Optional
 
--- a/cleanlab/datalab/internal/serialize.py
+++ b/cleanlab/datalab/internal/serialize.py
@@ -155,6 +155,7 @@ class _Serializer:
                     "cleanlab_version": datalab.cleanlab_version,
                     "verbosity": datalab.verbosity,
                     "info": datalab.info,
+                    "data_hash": datalab._data_hash,
                 },
                 f,
                 default=custom_serializer,
@@ -254,6 +255,7 @@ class _Serializer:
             )
             datalab.cleanlab_version = json_data["cleanlab_version"]
             datalab.info = json_data["info"]
+            datalab._data_hash = json_data["data_hash"]
 
         cls._validate_version(datalab)
 
@@ -267,7 +269,7 @@ class _Serializer:
             datalab.data_issues.issue_summary = pd.read_csv(issue_summary_path)
 
         if data is not None:
-            if hash(data) != hash(datalab._data):
+            if hash(data) != datalab._data_hash:
                 raise ValueError(
                     "Data has been modified since Lab was saved. "
                     "Cannot load Lab with modified data."
--- a/cleanlab/datalab/internal/serialize.py
+++ b/cleanlab/datalab/internal/serialize.py
@@ -22,9 +22,12 @@ from typing import TYPE_CHECKING, Optional
 
 import numpy as np
 import pandas as pd
+from scipy.sparse import csr_matrix
+from sklearn.neighbors import NearestNeighbors
+from cleanlab.outlier import OutOfDistribution
 
 import cleanlab
-from cleanlab.datalab.internal.data import Data, Label
+from cleanlab.datalab.internal.data import Data, Label, MultiLabel, MultiClass
 
 if TYPE_CHECKING:  # pragma: no cover
     from datasets.arrow_dataset import Dataset
@@ -90,17 +93,55 @@ class _Serializer:
             print(f"WARNING: Existing files will be overwritten by newly saved files at: {path}")
 
         def custom_serializer(obj):
+            """Custom serializer for handling specific data types."""
             if isinstance(obj, np.integer):
-                return int(obj)
-            elif isinstance(obj, np.ndarray):
-                return obj.tolist()
-            elif isinstance(obj, Label):
+                return {"__type__": type(obj).__name__, "value": int(obj)}
+
+            if isinstance(obj, np.ndarray):
+                return {"__type__": "ndarray", "data": obj.tolist()}
+
+            if isinstance(obj, csr_matrix):
                 return {
+                    "__type__": "csr_matrix",
+                    "data": obj.data.tolist(),
+                    "indices": obj.indices.tolist(),
+                    "indptr": obj.indptr.tolist(),
+                    "shape": obj.shape,
+                }
+
+            if isinstance(obj, pd.DataFrame):
+                return {
+                    "__type__": "DataFrame",
+                    "data": obj.to_dict(orient="records"),
+                    "columns": obj.columns.tolist(),
+                }
+
+            if isinstance(obj, Label):
+                return {
+                    "__type__": "Label",
+                    "class_name": obj.__class__.__name__,
                     "label_map": obj.label_map,
                     "labels": obj.labels,
                 }
-            else:
-                raise TypeError(f"Type {type(obj)} is not serializable")
+
+            if isinstance(obj, OutOfDistribution):
+                return {
+                    "__type__": "OutOfDistribution",
+                    "params": {k: v for k, v in obj.params.items() if k != "scaling_factor"},
+                }
+
+            if isinstance(obj, NearestNeighbors):
+                return {
+                    "__type__": "NearestNeighbors",
+                    "n_neighbors": obj.n_neighbors,
+                    "metric": obj.metric,
+                    "algorithm": obj.algorithm,
+                    "leaf_size": obj.leaf_size,
+                    "p": obj.p,
+                    "_fit_X": obj._fit_X.tolist(),
+                }
+
+            raise TypeError(f"Type {type(obj)} is not serializable")
 
         with open(os.path.join(path, OBJECT_FILENAME), "w") as f:
             json.dump(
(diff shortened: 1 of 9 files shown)


cleanlab/datalab/internal/serialize.py
```diff
@@ -107,21 +131,32 @@ def deserialize(cls, path: str, data: Optional[Dataset] = None) -> Datalab:
...
+            datalab._data_hash = datalab_metadata["_data_hash"]
+            datalab._labels.labels = datalab_metadata["_labels"]["labels"]
+            datalab._labels.label_map = datalab_metadata["_labels"]["label_map"]
 
         cls._validate_version(datalab)
 
         # Load the issues from disk.
         issues_path = os.path.join(path, ISSUES_FILENAME)
-        if not hasattr(datalab.data_issues, "issues") and os.path.exists(issues_path):
+        if os.path.exists(issues_path):
```
jwmueller, line 151: could you clarify why this was changed?

jwmueller:
Note this PR is failing on windows:
https://github.com/cleanlab/cleanlab/actions/runs/12554401936/job/35137672914?pr=1222
Screenshot 2025-01-03 at 7 05 24 PM

tests/datalab/datalab/test_datalab.py
```diff
@@ -18,7 +18,6 @@
 import contextlib
 import io
 import os
-import pickle
 import timeit
 from pathlib import Path
 from unittest.mock import MagicMock, Mock, patch
```
jwmueller, line 23: CI is complaining about imports that are no longer used:  https://github.com/cleanlab/cleanlab/actions/runs/12554401936/job/35137671038?pr=1222

Screenshot 2025-01-03 at 7 07 01 PM

jwmueller:
Things look good overall, but was still looking for an answer to this comment:
https://github.com/cleanlab/cleanlab/pull/1222#discussion_r1902274426

gordon-lim:
I replied to it. I paste it below for easy reference:

First, is that it is unnecessary. Whenever datalab is initialized, self.data_issues = builder.build() will enforce that datalab.data_issues.issues and datalab.data_issues.issue_summary attributes will always exist.

Second, previously when loading from pickle, the loaded datalab.data_issues.issues and datalab.data_issues.issue_summary attributes will always have the correct values i.e. there's no need to load from the saved CSVs. So we were able to pass test_load without ever entering those if conditionals. However, we are now rebuilding datalab from json so we need to remove the check for hasattr to ensure that we do enter the if conditional and that issues and issue_summary are loaded from csv.

cleanlab/datalab/internal/serialize.py
jwmueller: @gordon-lim Could you explain this briefly?


Also according to [codecov](https://app.codecov.io/gh/cleanlab/cleanlab/pull/1222?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=cleanlab), these if statements are only being partially tested:
```
if os.path.exists(issues_path)
```

and

```
if os.path.exists(issue_summary_path)
```

Not required, but it'd be nice if you can add small unit test that executes the other branch of the if statement which currently isn't ever executed during unit tests.

gordon-lim: I think my comment was pending. Can you see it now?

jwmueller: yes thanks for the explanation, makes sense

jwmueller (changes requested):
When I run our [first Datalab tutorial](https://docs.cleanlab.ai/master/tutorials/datalab/datalab_quickstart.html), and then try to save the Datalab from it, I get this error:

Screenshot 2025-01-04 at 9 52 48 PM

Please run [all of our Datalab tutorials](https://docs.cleanlab.ai/master/tutorials/datalab/index.html) and verify saving/loading all Datalab objects works properly.

gordon-lim (commented):
The original tests for ValueError do not work as expected. Code in question:
```python
with pytest.raises(ValueError) as excinfo:
    Datalab.load(tmp_path, data=dataset.shard(2, 0))
    expected_error_msg = "Length of data (2) does not match length of labels (5)"
    assert expected_error_msg == str(excinfo.value)
```
After a ValueError is raised, the `with` block is exited, so the exact error message is never checked. However, when dataset is shared it inherently implies that the dataset was not modified so even though the error message was "Data has been modified...", it went unchecked. As such, the wrong expected length in the error message was also unchecked (should be 3 instead of 2).

I'm not sure why there was a test for different lengths because of the aforementioned issue but to keep the test, I patched the hash results to bypass the hash check.

tests/datalab/datalab/test_datalab.py
```diff
@@ -489,16 +489,19 @@ def test_load(self, lab, tmp_path, dataset, mock_issues, mock_issue_summary, mon
         assert loaded_lab.data._data == dataset.data
 
         # Misaligned dataset raises a ValueError
-        with pytest.raises(ValueError) as excinfo:
```
gordon-lim, line 533: The original tests for ValueError do not work as expected. Code in question:
```python
with pytest.raises(ValueError) as excinfo:
    Datalab.load(tmp_path, data=dataset.shard(2, 0))
    expected_error_msg = "Length of data (2) does not match length of labels (5)"
    assert expected_error_msg == str(excinfo.value)
```
After a ValueError is raised, the `with` block is exited, so the exact error message is never checked. However, when dataset is shared it inherently implies that the dataset was not modified so even though the error message was "Data has been modified...", it went unchecked. As such, the wrong expected length in the error message was also unchecked (should be 3 instead of 2).

I'm not sure why there was a test for different lengths because of the aforementioned issue but to keep the test, I patched the hash results to bypass the hash check.

tests/datalab/datalab/test_datalab.py
```diff
@@ -489,16 +489,19 @@ def test_load(self, lab, tmp_path, dataset, mock_issues, mock_issue_summary, mon
...
-            Datalab.load(tmp_path, data=dataset.shard(2, 0))
-            expected_error_msg = "Length of data (2) does not match length of labels (5)"
-            assert expected_error_msg == str(excinfo.value)
```
gordon-lim, line 536: The original tests for ValueError do not work as expected. 

After a ValueError is raised, the `with` block is exited, so the exact error message is never checked. However, when dataset is shared it inherently implies that the dataset was not modified so even though the error message was "Data has been modified...", it went unchecked. As such, the wrong expected length in the error message was also unchecked (should be 3 instead of 2).

I'm not sure why there was a test for different lengths because of the aforementioned issue but to keep the test, I patched the hash results to bypass the hash check.

tests/datalab/datalab/test_datalab.py
```diff
@@ -432,9 +432,9 @@ def test_save(self, lab, tmp_path, mock_issues, mock_issue_summary, monkeypatch)
         lab.save(tmp_path, force=True)
         assert tmp_path.exists(), "Save directory was not created"
         assert (tmp_path / "data").is_dir(), "Data directory was not saved"
-        assert (tmp_path / "issues.csv").exists(), "Issues file was not saved"
-        assert (tmp_path / "summary.csv").exists(), "Issue summary file was not saved"
         assert (tmp_path / "datalab.json").exists(), "Datalab file was not saved"
+        assert not (tmp_path / "issues.csv").exists(), "Empty issues dataframe was saved"
+        assert not (tmp_path / "summary.csv").exists(), "Empty issue_summary dataframe was saved"
```
gordon-lim, line 437: Previously, `datalab` object with empty `issues` and `issue_summary` dataframes would be saved. It would not be able to be loaded. `pd.read_csv()` throws an error when fed empty csv. I changed the relevant code to not save to csv when the dataframe is empty and skip accordingly during load. As such, the test case needs to be changed as well to not expect those files to be saved before `issues` and `issue_summary` are mocked.

gordon-lim:
The `Datalab` object relies heavily on the dataset, which must be loaded for full functionality. As a result, it is reasonable for `datalab.save()` to save the dataset by default and for `datalab.load()` to load the dataset from the same path.

If a dataset is not loaded, the `issue` and `issues_summary` DataFrames remain accessible. However, certain features, such as `datalab.report`, will have reduced functionality—for example, reports cannot display images for image datasets without the dataset loaded. Additionally, the `find_issues` method becomes unavailable, effectively putting the `Datalab` object in a "read-only" state.

To clarify this behavior, I’ve updated the documentation and included screenshots for reference. Below is the relevant link for your convenience: [docs](https://docs.cleanlab.ai/stable/cleanlab/datalab/datalab.html#cleanlab.datalab.datalab.Datalab.save)

### `datalab.save()`
datalab.save()

### `datalab.load()`
datalab.load()

Previously, this was not an issue because saving the `Datalab` object using pickle automatically included the dataset. However, when serializing to JSON, special consideration is needed. Upon reconstruction, the dataset must be provided to the constructor to ensure other attributes are initialized correctly. For reference, here is the `Datalab` object constructor:

```python
class Datalab:
    def __init__(
        self,
        data: "DatasetLike",
        task: str = "classification",
        label_name: Optional[str] = None,
        image_key: Optional[str] = None,
        verbosity: int = 1,
    ) -> None:
        # Handle continuous labels for regression tasks
        # Map labels to integers for classification tasks
        self.task = Task.from_str(task)
        self._data = Data(data, self.task, label_name)
        self.data = self._data._data
        self._labels = self._data.labels
        self._label_map = self._labels.label_map
        self.label_name = self._labels.label_name
        self._data_hash = self._data._data_hash
        self.cleanlab_version = cleanlab.version.__version__
        self.verbosity = verbosity

        # Create the builder for DataIssues
        builder = _DataIssuesBuilder(self._data)
        builder.set_imagelab(self._imagelab).set_task(self.task)
        self.data_issues = builder.build()
```

jwmueller:
> The `Datalab` object relies heavily on the dataset, which must be loaded for full functionality. As a result, it is reasonable for `datalab.save()` to save the dataset by default and for `datalab.load()` to load the dataset from the same path.
> 
> If a dataset is not loaded, the `issue` and `issues_summary` DataFrames remain accessible. However, certain features, such as `datalab.report`, will have reduced functionality—for example, reports cannot display images for image datasets without the dataset loaded. Additionally, the `find_issues` method becomes unavailable, effectively putting the `Datalab` object in a "read-only" state.
> 
> To clarify this behavior, I’ve updated the documentation and included screenshots for reference. Below is the relevant link for your convenience: [docs](https://docs.cleanlab.ai/stable/cleanlab/datalab/datalab.html#cleanlab.datalab.datalab.Datalab.save)
> 
> ### `datalab.save()`
> datalab.save()
> 
> ### `datalab.load()`
> datalab.load()
> 
> Previously, this was not an issue because saving the `Datalab` object using pickle automatically included the dataset. However, when serializing to JSON, special consideration is needed. Upon reconstruction, the dataset must be provided to the constructor to ensure other attributes are initialized correctly. For reference, here is the `Datalab` object constructor:
> 
> ```python
> class Datalab:
>     def __init__(
>         self,
>         data: "DatasetLike",
>         task: str = "classification",
>         label_name: Optional[str] = None,
>         image_key: Optional[str] = None,
>         verbosity: int = 1,
>     ) -> None:
>         # Handle continuous labels for regression tasks
>         # Map labels to integers for classification tasks
>         self.task = Task.from_str(task)
>         self._data = Data(data, self.task, label_name)
>         self.data = self._data._data
>         self._labels = self._data.labels
>         self._label_map = self._labels.label_map
>         self.label_name = self._labels.label_name
>         self._data_hash = self._data._data_hash
>         self.cleanlab_version = cleanlab.version.__version__
>         self.verbosity = verbosity
>         self._imagelab = create_imagelab(dataset=self.data, image_key=image_key)
> 
>         # Create the builder for DataIssues
>         builder = _DataIssuesBuilder(self._data)
>         builder.set_imagelab(self._imagelab).set_task(self.task)
>         self.data_issues = builder.build()
> ```

Thanks for pointing this out.  The ideal behavior would be for user to separately save/load their dataset themselves, and then Datalab.save() does not save the dataset at all.  For `Datalab.load()`, it could take in arg which is the loaded dataset (must be exactly the same as the dataset originally was when Datalab was previously created), and this is then properly set as the `self.data` attribute of Datalab.

gordon-lim:
> Thanks for pointing this out. The ideal behavior would be for user to separately save/load their dataset themselves, and then Datalab.save() does not save the dataset at all. For `Datalab.load()`, it could take in arg which is the loaded dataset (must be exactly the same as the dataset originally was when Datalab was previously created), and this is then properly set as the `self.data` attribute of Datalab.

Thank you for the clarification. As a follow up, why are we setting the data attribute instead of reconstructing with the dataset provided? This handles many attributes which are assigned values based on the dataset e.g. `self._labels`, `self._label_map`, `self.data_issues`. How much functionality are we expected to preserve when no dataset is provided vs. when there is one?

jwmueller:
> > Thanks for pointing this out. The ideal behavior would be for user to separately save/load their dataset themselves, and then Datalab.save() does not save the dataset at all. For `Datalab.load()`, it could take in arg which is the loaded dataset (must be exactly the same as the dataset originally was when Datalab was previously created), and this is then properly set as the `self.data` attribute of Datalab.
> 
> Thank you for the clarification. As a follow up, why are we setting the data attribute instead of reconstructing with the dataset provided? This handles many attributes which are assigned values based on the dataset e.g. `self._labels`, `self._label_map`, `self.data_issues`. How much functionality are we expected to preserve when no dataset is provided vs. when there is one?

The ideal workflow users should follow is:

```
datalab1 = Datalab(data)
... find issues, generate report ...
datalab1.save()
save_my_dataset()

# New python session:
data = load_my_dataset()
datalab2 = Datalab.load(path, data=data)
```

At that point `datalab2` should ideally have the same functionality as `datalab1`.  Sounds like you found that's not the case, because `load()` doesn't initialize the Datalab object with `data` properly?

If you're unable to fix this, then can you enumerate the list of problems you found because of this (missing functionalities in `datalab2`).  Then we can merge this PR with a GH issue listing those problems for future work.

gordon-lim:
> > > Thanks for pointing this out. The ideal behavior would be for user to separately save/load their dataset themselves, and then Datalab.save() does not save the dataset at all. For `Datalab.load()`, it could take in arg which is the loaded dataset (must be exactly the same as the dataset originally was when Datalab was previously created), and this is then properly set as the `self.data` attribute of Datalab.
> > 
> > 
> > Thank you for the clarification. As a follow up, why are we setting the data attribute instead of reconstructing with the dataset provided? This handles many attributes which are assigned values based on the dataset e.g. `self._labels`, `self._label_map`, `self.data_issues`. How much functionality are we expected to preserve when no dataset is provided vs. when there is one?
> 
> The ideal workflow users should follow is:
> 
> ```
> datalab1 = Datalab(data)
> ... find issues, generate report ...
> datalab1.save()
> save_my_dataset()
> 
> # New python session:
> data = load_my_dataset()
> datalab2 = Datalab.load(path, data=data)
> ```
> 
> At that point `datalab2` should ideally have the same functionality as `datalab1`. Sounds like you found that's not the case, because `load()` doesn't initialize the Datalab object with `data` properly?
> 
> If you're unable to fix this, then can you enumerate the list of problems you found because of this (missing functionalities in `datalab2`). Then we can merge this PR with a GH issue listing those problems for future work.

In other words, Datalab must be loaded with a dataset and `data` in `Datalab.load` is a required field. This does simplify things a lot. I will make the adjustments.

PR #1235, opened 2025-04-02
Fixed empty label handling in filter.py

GitGautamHub:
### Purpose  
Fix the issue where images with no labels (background-only) cause an `IndexError` in `cleanlab.object_detection.filter.py`.  

### Fix  
Added a check to handle empty label cases properly.  

## Example Code Snippet to Demonstrate Fix

```python
from cleanlab.object_detection.filter import find_label_issues
import numpy as np

# Test case: Image with no labels (only background)
labels = [{"labels": np.array([]), "bboxes": np.array([])}]
predictions = [np.array([])]  # No predictions either

# Function should now handle empty labels without errors
result = find_label_issues(labels, predictions)
print(result)  # Expected Output: []

### DIFF
--- a/cleanlab/object_detection/filter.py
+++ b/cleanlab/object_detection/filter.py
@@ -314,18 +314,22 @@ def _calculate_ap_per_class(
     return ap_per_class_list
 
 
-def _filter_by_class(
-    labels: List[Dict[str, Any]], predictions: List[np.ndarray], class_num: int
-) -> Tuple[List, List]:
-    """
-    Filters predictions and labels based on a specific class number.
-    """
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by jwmueller — CHANGES_REQUESTED** [review]
Thanks for your contribution.  Seems like the unit tests are failing though?

```
cleanlab/object_detection/filter.py:290: IndexError
```

https://github.com/cleanlab/cleanlab/actions/runs/14211326922/job/39818959775?pr=1235

CI is also complaining about the formatting as well, you'll need to run the black formatter before you commit.

Screenshot 2025-04-02 at 10 10 40 PM

GitGautamHub:
Hey @jwmueller ,
I will look into it

PR #1242, opened 2025-07-14
Fix #1175: Replace Pickle with Parquet for Secure Serialization

### LINKED ISSUE #1175: replace Datalab load/save from pickle to more secure format

minhkhoango:
Summary
This PR addresses a critical security vulnerability by replacing the pickle serialization engine.

Resolves: Issue #1175

Problem: The use of pickle exposes the application to a critical RCE (Remote Code Execution) vulnerability.

Solution: Replaced the serialization engine with the memory-safe and efficient pyarrow.parquet format.

Key Benefits:

✅ Security: Eliminates the RCE vulnerability entirely.

✅ Efficiency: Reduces artifact storage size by ~40%.

Solution & Trade-offs
The core change replaces the serialization logic to use Parquet, a secure, column-oriented data format. The _sanitize_for_json function was hardened into a robust anti-corruption layer to handle non-serializable objects like scikit-learn models, ensuring the process is resilient.

As part of this change, we accept a performance trade-off for the sake of security and efficiency:

The benchmark below shows that serialization (save/load) is now slower on the test dataset. This is an expected and acceptable trade-off for the significant gains in security and storage efficiency.

Benchmark Results
The following benchmark was run on a 100,000-row dataset using a cloud-based AMD Threadripper Pro 5995WX instance.

image

Further Investigation: Discovered a Critical Scaling Bottleneck
During the extended 1,000,000-row benchmark, I identified a severe performance scaling limitation in the core find_issues method, unrelated to the serialization engine.

When the dataset grows beyond the CPU's L3 cache, the KNN graph computation becomes memory-bound, not CPU-bound, causing runtime to increase super-quadratically. This is evidence of a classic "CPU Cache Cliff" and fundamentally limits the ability of cleanlab to scale to enterprise-level datasets.

The htop analysis below shows the high-end CPU is mostly idle, waiting for data from RAM, which confirms the bottleneck is memory latency, not compute.

image

The industry-standard solution for this problem is to implement Approximate Nearest Neighbor (ANN) search using a library like Meta's Faiss. While a full implementation is beyond the scope of this security patch, addressing this should be a high-priority architectural improvement for the project.

This PR provides the foundation for future performance work by ensuring the data serialization pipeline is secure and robust.


minhkhoango:
Hi all, just checking if there’s any feedback I can address on this PR? Happy to split it up if helpful

PR #1243, opened 2025-07-21
Updated the numpy dependency version to Numpy 2.0 to avoid nan errors

adibbehjat:
## Summary

🎯 **Purpose**: 

Updating the toml dependency for Numpy to resolve the np.nan error:

```
AttributeError: `np.NaN` was removed in the NumPy 2.0 release. Use `np.nan` instead.
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ name = "cleanlab"
 # requirements files see:
 # https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/
 dependencies = [
-  "numpy~=1.22",
+  "numpy~=2.0",
   "scikit-learn>=1.1",
   "tqdm>=4.53.0",
   "pandas>=1.4.0",


clemwangds:
I'm still having a numpy related install problem which I don't understand.  Has this patch been applied?

```
Downloading cleanlab-2.7.1-py3-none-any.whl (349 kB)
Downloading numpy-1.26.4-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (18.0 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 18.0/18.0 MB 2.9 MB/s eta 0:00:00
Installing collected packages: numpy, cleanlab
  Attempting uninstall: numpy
    Found existing installation: numpy 2.3.1
    Uninstalling numpy-2.3.1:
      Successfully uninstalled numpy-2.3.1
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
thinc 8.3.6 requires numpy<3.0.0,>=2.0.0, but you have numpy 1.26.4 which is incompatible.
Successfully installed cleanlab-2.7.1 numpy-1.26.4

```

Issue #1317, opened 2026-07-07
Dataset-level "is there any signal at all" test (complementing overall_label_health_score)


popixoxipop-collab:
`overall_label_health_score` / `health_summary` report a dataset-level label-quality score, but as far as I can tell from reading `cleanlab/dataset.py` and `cleanlab/count.py`, there's no calibrated test attached to that score for whether it's actually distinguishable from what you'd see on a dataset with **no real label signal** (e.g. shuffled labels). The confident-joint machinery isn't designed to be trustworthy in that regime, so it can still report a health score that looks informative on pure noise. (I originally assumed there was no dataset-level score here at all — that was wrong on my part; the actual gap is narrower: a score exists, just with no significance test attached to it.)

**Proposal**: an exact, calibrated GO/NO-GO test. Conditional on out-of-sample predictions, under H0 = "labels carry no information beyond the class prior," the number of correct predictions is exactly Poisson-binomial distributed. Its survival function gives an exact p-value — no Monte Carlo, no concentration-inequality slack.

**What I'm proposing**:
- `overall_label_signal_pvalue(labels, pred_probs, delta=0.05)` in `cleanlab/dataset.py`, next to `overall_label_health_score`.
- Wired into `health_summary()`'s returned dict as a new `"overall_label_signal_pvalue"` key, only when `labels`+`pred_probs` are given directly (it needs per-example prediction/label alignment, so unlike `overall_label_health_score` it can't be derived from a pre-aggregated `joint`/`confident_joint`).

**Limitations, up front**:
- O(n²) runtime in the current draft (a sequential Poisson-binomial convolution). Fine for small/medium datasets, could be slow at tens of thousands of examples or more. A grouped-by-class or DFT-based algorithm could likely get this to roughly O(n·n_classes) or O(n log n); not implemented yet.
- No multi-label support yet.
- The `<= delta` false-positive-rate bound is exact when the class prior is known; here the prior is estimated from the same sample, so the guarantee is empirically validated (calibration checked under repeated null trials) rather than a formally proven bound in that setting.

**Open question for maintainers**: `cleanlab/datalab/internal/issue_manager/noniid.py` already has a permutation-test-based p-value implemented as a `Datalab` `IssueManager` (for a different question — data-ordering IID-ness). Should this be an `IssueManager` too, instead of a plain function next to `overall_label_health_score`?

[...]

PR #1318, opened 2026-07-07
Add overall_label_signal_pvalue: exact GO/NO-GO test for "is there any label signal at all"

popixoxipop-collab:
## Motivation
`overall_label_health_score` reports *how clean* a dataset's labels are, but there's no calibrated test for whether that score is even meaningfully distinguishable from a dataset with **no real signal** (e.g. shuffled labels) in the first place. This PR adds that test.

Relates to #1317

## What's added
- `cleanlab.dataset.overall_label_signal_pvalue(labels, pred_probs, delta=0.05, multi_label=False, verbose=True) -> float`: an exact finite-sample p-value for H0 = "labels carry no information beyond the class prior," via the Poisson-binomial survival function of held-out accuracy. Reject H0 (declare signal) iff `p_value <= delta`.
- `health_summary()` now also returns `"overall_label_signal_pvalue"` (a new `delta` kwarg controls the printed verdict's threshold). It's `None` when `labels`/`pred_probs` weren't both given directly — see limitation 1 below.

## Why a separate function instead of extending `overall_label_health_score`?
The two questions are logically ordered: "is there a signal at all" needs answering before "how much of the data is clean" is meaningful. I kept them as separate functions (mirroring `overall_label_health_score`'s own shape) rather than overloading one function's return type — open to a different structure if maintainers prefer.

## Limitations / things I'd like feedback on
1. **Requires `labels` + `pred_probs` directly.** Unlike `overall_label_health_score`, this can't be computed from a pre-aggregated `joint`/`confident_joint` alone — the test needs per-example prediction/label alignment, which those aggregate forms don't preserve. Raises `ValueError` otherwise.
2. **O(n²) runtime** (sequential convolution over per-example match probabilities). Fine for small/medium datasets, could be a real bottleneck at tens of thousands of examples or more. A grouped-by-predicted-class convolution, or a DFT-based Poisson-binomial algorithm (Fernandez & Williams, 2010), could likely bring this to roughly O(n·n_classes) or O(n log n) — neither is implemented here; flagging this rather than shipping it quietly.
3. **No multi-label support** yet (raises `ValueError`).
4. **Placement**: this is a plain function next to `overall_label_health_score`. `cleanlab/datalab/internal/issue_manager/noniid.py` has a precedent for a permutation-test p-value implemented as an `IssueManager` instead — genuinely unsure which is the better home for this, would like input.
5. **The `<= delta` false-positive-rate guarantee is exact when the class prior is known**; here it's estimated from the same sample, so in that setting it's empirically validated (see the Type-I calibration test below) rather than a formally proven bound.

## Test plan
- 8 new tests in `tests/test_dataset.py` (`TestOverallLabelSignalPvalue`, `TestHealthSummarySignalPvalue`): pure-noise → not detected, strong-signal → detected, Type-I error rate checked empirically over 200 repeated null trials, length-mismatch / missing-input / `multi_label=True` all raise `ValueError`, and `health_summary` integration both with and without `labels`/`pred_probs` available.
- Full `tests/test_dataset.py` suite (33 tests) passes, no regressions.
- `black --check` (pinned CI version 25.12.0): clean.
- `flake8` (this repo's configured F401/F403 checks): clean.
- `mypy --install-types --non-interactive cleanlab` (same command CI uses): 0 new errors from this diff — checked against a clean-`master` baseline, the same 12 pre-existing errors (in unrelated files: `object_detection`, `segmentation`, `count.py`, `multiannotator.py`, `experimental/`, `datalab/` cleanvision imports) are present with or without this change.

### DIFF
--- a/cleanlab/dataset.py
+++ b/cleanlab/dataset.py
@@ -358,6 +358,117 @@ def overall_label_health_score(
     return health_score
 
 
+def overall_label_signal_pvalue(
+    labels=None,
+    pred_probs=None,
+    *,
+    delta: float = 0.05,
+    multi_label=False,
+    verbose=True,
+) -> float:
+    """Returns an exact p-value testing whether there is *any* extractable signal between the
+    features and labels in your dataset, i.e. whether ``pred_probs`` predicts ``labels`` better
+    than chance.
+
+    This answers a question that is logically prior to `~cleanlab.dataset.overall_label_health_score`:
+    the health score estimates *how much* of a dataset is mislabeled, which presupposes there is a
+    meaningful signal for labels to deviate from. On a dataset with no real signal (e.g. shuffled
+    labels), the confident-joint machinery is not designed to be trustworthy, yet can still return a
+    health score that looks informative. This test gives a calibrated way to catch that case before
+    trusting the health score.
+
+    Conditional on the out-of-sample predictions ``pred_probs.argmax(axis=1)``, under the null
+    hypothesis H0 = "the labels carry no information beyond the class prior", each per-example match
+    indicator ``1[prediction_i == label_i]`` is an independent Bernoulli trial with success
+    probability equal to the empirical prior of the predicted class. The number of correct
+    predictions is therefore exactly Poisson-binomial distributed under H0, and its survival
+    function is an *exact* p-value for that null: no concentration-inequality slack and no Monte
+    Carlo / permutation sampling.
+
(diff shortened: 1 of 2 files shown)


popixoxipop-collab:
Closing this — I filed it without getting full sign-off first, which wasn't appropriate for an outside contribution. Apologies for the noise. The technical content (the significance-test gap analysis, and the corrected complexity claim) may still be useful independently, but I'd rather properly re-evaluate before resubmitting anything.

PR #1323, opened 2026-07-27
Turn near-duplicate score test into a property-based test

### LINKED ISSUE #944: Turn near-duplicate score test into a property-based test
_Originally posted by @elisno in https://github.com/cleanlab/cleanlab/pull/943#discussion_r1442401494_

vishnukannaujia:
Closes #944.

### Description

`TestNearDuplicateIssueManager.test_scores_of_examples_with_issues_are_smaller_than_those_without` (which carried a `# TODO: Turn this into a property-based test`) asserted the *"near-duplicate examples have worse scores than non-near-duplicates"* property against a single hand-crafted toy dataset.

This replaces it with a [Hypothesis](https://hypothesis.readthedocs.io/) property-based test, following the embedding-generation recipe from the issue:

- **`embeddings_with_near_duplicates_strategy`** places base rows far apart (spacing ≈ 1 per feature, well above the default `0.13` near-duplicate threshold), so base rows are never flagged among themselves;
- it then re-appends a **non-empty proper subset** of base rows with tiny noise (≤ `1e-4`), producing genuine near-duplicates that must be flagged while guaranteeing at least one non-duplicated row remains;
- the test asserts the property `max(score[is_issue]) < min(score[non_issue])` across generated inputs.

This checks the invariant on many randomized shapes/dimensions/duplicate-subsets instead of one fixed example, as requested.

### Verification

`--hypothesis-show-statistics`: **100 passing, 0 failing**, ~12% filtered by `assume` (the guard for "both flagged and unflagged rows present", which the strategy already makes the common case). Full `test_duplicate.py` suite passes (7 tests); `black` (line-length 100) clean; no new imports (flake8 `F401/F403` unaffected).

### DIFF
--- a/tests/datalab/issue_manager/test_duplicate.py
+++ b/tests/datalab/issue_manager/test_duplicate.py
@@ -32,6 +32,56 @@ def embeddings_strategy(draw):
     return embeddings
 
 
+@composite
+def embeddings_with_near_duplicates_strategy(draw):
+    """Generate embeddings guaranteed to contain both near-duplicates and non-duplicates.
+
+    Base rows are placed far apart (>= ~1 apart per feature, well above the default
+    ``0.13`` near-duplicate threshold) so they are never flagged among themselves. A
+    non-empty proper subset of base rows is then re-appended with tiny noise, producing
+    genuine near-duplicates that must be flagged while leaving at least one non-duplicated
+    base row. This lets the "issue scores are smaller than non-issue scores" property be
+    checked without relying on a single hand-crafted toy dataset.
+    """
+    n_base = draw(st.integers(min_value=4, max_value=8))
+    dim = draw(st.integers(min_value=2, max_value=3))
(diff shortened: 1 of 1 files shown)


vishnukannaujia:
A quick note on CI to save a reviewer's time: the red checks here (`test`, `Format`, `Type check`, `Maximum Dependency Versions: Python 3.14`, `Bleeding Edge Dependencies`, and the `Core Tests` matrix) are **pre-existing failures on `master`**, not introduced by this PR — the same checks fail on the current master HEAD (`7506257`). This change only adds a Hypothesis property-based test in `tests/datalab/issue_manager/test_duplicate.py` and touches no package code, so it can't affect type-checking, formatting, or the dependency matrices.

Would a maintainer be able to take a look when convenient? Happy to rebase once master's CI is green if you'd prefer to see a clean run first.

